PR #116, opened 2023-09-13
feat: add error boundary [SPA-1507]

loweisz:
* Catch errors happening during rendering of the experience and display them in the Canvas when operating in `editor` mode.
* Send errors to the web app if error originated from the SDK and not the imported components.
* Display informative error screen instead of blank screen when an error happens as shown below:

Screenshot 2023-10-10 at 14 43 13

Screenshot 2023-10-10 at 14 43 19

### DIFF
--- a/src/blocks/ErrorBoundary.css
+++ b/src/blocks/ErrorBoundary.css
@@ -1,0 +1,13 @@
+.error-message {
+    margin: 24px;
+    font-size: var(--font-size-m);
+    font-family: var(--font-stack-primary);
+    color: var(--red-800);
+    padding: 16px;
+    background-color: var(--red-200);
+}
+
+.error-message .title {
+    margin-top: 0;
+    font-size: var(--font-size-l);
+}
--- a/src/blocks/ErrorBoundary.css
+++ b/src/blocks/ErrorBoundary.css
@@ -1,4 +1,4 @@
-.error-message {
+.cf-error-message {
   margin: 24px;
   font-size: var(--font-size-m);
   font-family: var(--font-stack-primary);
@@ -7,12 +7,12 @@
   background-color: var(--red200);
 }
 
-.error-message .title {
+.cf-error-message .title {
   margin-top: 0;
   font-size: var(--font-size-l);
 }
 
-.more-details {
+.cf-error-message .more-details {
   cursor: pointer;
   color: var(--blue700);
 }
--- a/src/blocks/ErrorBoundary.css
+++ b/src/blocks/ErrorBoundary.css
@@ -1,13 +1,13 @@
 .error-message {
-    margin: 24px;
-    font-size: var(--font-size-m);
-    font-family: var(--font-stack-primary);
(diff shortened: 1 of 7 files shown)


src/blocks/ErrorBoundary.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return this.props.children
+  }
+}
+
+export class ImportedComponentErrorBoundary extends React.Component<{ children: ReactElement }> {
+  componentDidCatch(error: Error, _errorInfo: ErrorInfo) {
+    const err = new ImportedComponentError(error.message)
+    err.stack = error.stack
+    throw err
+  }
```
loweisz, line 74: That is pretty smart, good job 💪

src/blocks/ErrorBoundary.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  static getDerivedStateFromError() {
+    return { hasError: true }
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    this.setState({ error, errorInfo })
+    if (!(error instanceof ImportedComponentError)) {
+      sendMessage(OutgoingExperienceBuilderEvent.CANVAS_ERROR, error)
+    }
```
loweisz, line 25: Shall we still throw the error when it is an imported component error? So that we don't swallow it for the user

loweisz (commented):
Very smart solution 💪 just one small suggestion

src/blocks/ErrorBoundary.css
```diff
@@ -0,0 +1,18 @@
+.error-message {
```
Spring3, line 1: Should we prefix this and other class anmes somehow to minimize the risk of affecting other existing elements on the page?

src/blocks/ErrorBoundary.tsx
SofiaMargariti: So I updated the code to re-throw here. One thing though is that the error screen is no longer shown for any ImportedComponentError. If the customer have setup up their error error boundary in their website it will be handled there or a blank screen will be shown. But I guess this is ok because the original intention was to handle errors coming from our own code? What do you think?

src/styles/variables.css
```diff
@@ -27,6 +29,7 @@
 
   /* typography */
   --font-size-m: 0.875rem;
+  --font-size-l: 1rem;
```
Spring3, line 32: oh in this PR I made `--font-size-m` be 1.4rem, which will be larger than this `--font-size-l`. Would you mind renaming it to `--font-size-s` instead to keep it consistent [with this other PR](https://github.com/contentful/experience-builder/pull/137)?

src/blocks/ErrorBoundary.tsx
loweisz: Yes I agree 👍

loweisz:
can't approve as I opened the PR, but LGTM :D

PR #134, opened 2023-10-05
fix: default height for containers

ChidinmaOrajiaku:
Currently, when you nest a container within a parent container that has an existing `non-container` child, the child container defaults to 0 px height.

This PR gives all containers a default fixed height of `200px`.

### DIFF
--- a/src/blocks/VisualEditorBlock.tsx
+++ b/src/blocks/VisualEditorBlock.tsx
@@ -78,7 +78,6 @@ export const VisualEditorBlock = ({
             variableName === 'cfHeight'
               ? calculateNodeDefaultHeight({
                   blockId: node.data.blockId,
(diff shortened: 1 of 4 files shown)


Spring3 (changes requested):
It is still possible to reproduce the issue for empty nested containers, like in the example below, when they are not the only child of another container

Screenshot 2023-10-06 at 11 50 51

ChidinmaOrajiaku:
> It is still possible to reproduce the issue for empty nested containers, like in the example below, when they are not the only child of another container
> 
> Screenshot 2023-10-06 at 11 50 51

Are you using the right SDK branch version? Could you also please confirm from the CSS inspector that the height is 200px to ensure you're testing the right version?

Spring3 (approved):

PR #135, opened 2023-10-09
feat: consistent sizing across options [SPA-1463]

ChidinmaOrajiaku:
- Add height and overflow to visual editor root
- Add capture to scroll event so that nested element scrolls can be bubbled up

### DIFF
--- a/src/styles/VisualEditorRoot.css
+++ b/src/styles/VisualEditorRoot.css
@@ -18,5 +18,5 @@ html::-webkit-scrollbar {
 #VisualEditorRoot.root {
   height: 100vh;
   padding-bottom: 100px;
-  overflow: scroll;
+  overflow: auto;
 }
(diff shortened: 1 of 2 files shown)


src/styles/VisualEditorRoot.css
```diff
@@ -16,6 +16,7 @@ html::-webkit-scrollbar {
 }
 
 #VisualEditorRoot.root {
-  height: 100%;
+  height: 100vh;
   padding-bottom: 100px;
+  overflow: scroll;
```
DanweDE, line 21: Not sure this is a good idea... If you're on Macbook, I know scrollbars are generally hidden (I'd recommend any web dev on a MacBook to [do this](https://apple.stackexchange.com/questions/359794/how-do-i-fix-this-annoying-chrome-for-macos-scroll-bar-issue)) but on other systems you might always see them.

Basically, we'd be forcing scrollbars on that element, even if there's no overflow.

Spring3 (approved):

PR #136, opened 2023-10-11
feat: support ssr [SPA-1435]

Spring3:
## SSR (Server Side Rendering)
```tsx
import { createExperience, fetchers, ExperienceRoot } from '@contentful/experience-builder';
import { GetServerSidePropsContext, InferGetServerSidePropsType } from 'next';

export const getServerSideProps = async ({
  params,
  locale,
}: GetServerSidePropsContext) => {
  const currentLocale = (locale || params?.lang) as string;
  const slug = params?.slug as string; // slug in this example, but devs can choose their own way

  const experienceEntry = await fetchers.fetchExperienceEntry({
    client,
    experienceTypeId,
    locale: currentLocale,
    idenifier: {
      slug,
    },
  });

  const { entries, assets } = await fetchers.fetchReferencedEntities({
    client,
    experienceEntry,
    locale: currentLocale,
  });

  return {
    props: {
      experienceEntry,
      referencedAssets: assets,
      referencedEntries: entries,
      slug,
      locale: currentLocale,
    },
  };
};

function SlugPage({
  experienceEntry,
  referencedAssets,
  referencedEntries,
  slug,
  locale,
}: InferGetServerSidePropsType<typeof getServerSideProps>) {

  const experience = createExperience({
    experienceEntry,
    referencedAssets,
    referencedEntries,
    locale,
    mode: mode as ExternalSDKMode,
  });

  if (!slug || typeof slug !== 'string') {
    return <p>404 Page not found</p>;
  }

  return (
    <main style={{ width: '100%' }}>
      <ExperienceRoot experience={experience} locale={locale} />
    </main>
  );
}

export default SlugPage;
```

Response from server:
Screenshot 2023-10-11 at 14 07 27

## SSG (Server Side Generation)

```tsx
export const getStaticProps = async ({
  params,
  locale,
}: GetServerSidePropsContext) => {
  const currentLocale = (locale || params?.lang) as string;
  const slug = params?.slug as string;

  const experienceEntry = await fetchers.fetchExperienceEntry({
    client,
    experienceTypeId,
    locale: currentLocale,
    idenifier: {
      slug,
    },
  });

  const { entries, assets } = await fetchers.fetchReferencedEntities({
    client,
    experienceEntry,
    locale: currentLocale,
  });

  return {
    props: {
      experienceEntry,
      referencedAssets: assets,
      referencedEntries: entries,
      slug,
      locale: currentLocale,
    },
  };
};

export const getStaticPaths = async () => {
  const experienceEntries = await client.getEntries({
    content_type: experienceTypeId,
  });

  const locales = await client.getLocales()

  const paths = [];

  for (const locale of locales.items) {
    for (const experienceEntry of experienceEntries.items) {
      paths.push({ params: {
        id: experienceEntry.sys.id,
        slug: experienceEntry.fields.slug,
        lang: locale.code,
      }})
    }
  }

  return { paths, fallback: false };
}

function SlugPage({
  experienceEntry,
  referencedAssets,
  referencedEntries,
  slug,
  locale,
}: InferGetServerSidePropsType<typeof getStaticProps>) {

  const experience = createExperience({
    experienceEntry,
    referencedAssets,
    referencedEntries,
    locale,
    mode: mode as ExternalSDKMode,
  });

  if (!slug || typeof slug !== 'string') {
    return <p>404 Page not found</p>;
  }

  return (
    <main style={{ width: '100%' }}>
      <ExperienceRoot experience={experience} locale={locale} />
    </main>
  );
}
```

### Option 1 (current approach - deprecated)

Disadvantages:

* basically gives no option to fetch other than by slug
* no easy way to fetch any other way
* no control over errors

```tsx
import { useExperienceBuilder, ExperienceRoot } from '@contentful/experience-builder'

const MyPage = () => {

  const { experience } = useExpereinceBuilder({ client, mode, experienceTypeId });

  return (
     <ExperienceRoot experience={experience} slug={router.slug} locale={router.locale} />
  )
}
```

### Option 2 - using a new `useFetchExperience` hook

* Kind of the same approach, but more explicit.
* Devs can now add a loading screen, handle errors

Disadvantages:
* still only fetches by slug

```tsx
import { useFetchExperience, ExperienceRoot } from '@contentful/experience-builder'

const MyPage = () => {

  const { fetchBySlug, experience, isFetching } = useFetchExperience({ client, mode });

  useEffect(() => {
     if (!experience && firstAttempt) {
        try {
           const experience = await fetchBySlug({ slug, experienceTypeId, locale });
        } catch (e) {}
     }
  }, [experience, client, slug, locale, experienceTypeId])

  if (isFetching) {
     return null;
  }

  return (
     <ExperienceRoot experience={experience} locale={router.locale} />
  )
}
```

### Option 3 - fetchers or manual. Full flexibility, most agnostic

* More flexibility as fetchers allow fetching by slug and id

```tsx
import { fetchers, createExperience, ExperienceRoot } from '@contentful/experience-builder'

const MyPage = () => {
  const [experience, setExpereince] = useState<Expeirence | undefined>(undefined);

  useEffect(() => {
     if (!experience && firstAttempt) {
        // in this example I am using fetchers, but you can really use any client or any way to fetch the experience entry
        const experienceEntry = await fetchers.fetchExperienceEntry({
           client,
           locale,
           experienceTypeId,
           identifier: { slug }
        });

        // same here
        const { entries, assets } = await fetchers.fetchReferencedEntities({ experienceEntry, client, locale });

        // all they really need to do eventually is get data for this function and feed it 
        const experience = createExperience({
           experienceEntry,
           referencedEntries: entries,
           referencedAssets: assets,
           mode,
           locale
        });

        setExperience(experience);
     }
  }, [experience, client, slug, locale, experienceTypeId])

  if (!experience) {
     return null;
  }

  return (
     <ExperienceRoot experience={experience} locale={router.locale} />
  )
}
```


DanweDE:
> Option 3 - fetchers or manual. Full flexibility, most agnostic

I wonder if we should really expose/recommend both `fetchers.fetchExperienceEntry` and `fetchers.fetchExperienceEntities` in the SDK publicly, or offer a framework-agnostic version of `useFetchExperience` instead.

The reason being, that for now we require separate requests, but as soon as we change the way we fetch, e.g. via a single CDA request, we had to deprecate these two methods again (or at least recommend not using them) and offer a single method as mentioned above instead.

DanweDE:
> `const { fetchBySlug, experience, isFetching } = useFetchExperience({ client, mode });`
> ...
> `const { success, experience, error } = await fetchBySlug({ client, slug, experienceTypeId, locale });`

`fetchBySlug` shouldn't require a `client`, considering the method is coming from `useFetchExperience` which takes the `client` already.

DanweDE:
> Option 1 (current approach - deprecated)

Any thoughts on simply removing this approach immediately and releasing this change as `3.0.0`?

Spring3:
> > Option 3 - fetchers or manual. Full flexibility, most agnostic
> 
> I wonder if we should really expose/recommend both `fetchers.fetchExperienceEntry` and `fetchers.fetchExperienceEntities` in the SDK publicly, or offer a framework-agnostic version of `useFetchExperience` instead.
> 
> The reason being, that for now we require separate requests, but as soon as we change the way we fetch, e.g. via a single CDA request, we had to deprecate these two methods again (or at least recommend not using them) and offer a single method as mentioned above instead.

If I understood your point correctly - you suggest to not export these functions, right?

I think it might be useful for now as it saves time for developers and they wouldn't have to try to understand where references are stored in experience entries and so on. We can provide a separate function for this and save them some time, but I don't think that the point of "in future we might have things differently" is a good point from not providing these helpers now, before that future arrived.

Spring3:
> > Option 1 (current approach - deprecated)
> 
> Any thoughts on simply removing this approach immediately and releasing this change as `3.0.0`?

We can. I only thought that it might be interesting to test it out somewhere before breaking. I don't know if this approach will fit every customer. Maybe we can release it as is and let's say release a 3.0.0 at the end of month, giving it 0,5 months to bake?

Spring3:
> > `const { fetchBySlug, experience, isFetching } = useFetchExperience({ client, mode });`
> > ...
> > `const { success, experience, error } = await fetchBySlug({ client, slug, experienceTypeId, locale });`
> 
> `fetchBySlug` shouldn't require a `client`, considering the method is coming from `useFetchExperience` which takes the `client` already.

Sorry, it was my bad. It doesn't require client.

DanweDE:
> If I understood your point correctly - you suggest to not export these functions, right?

We can export them, but we wouldn't recommend using them in docs, instead

> offer a framework-agnostic version of `useFetchExperience` instead.

So, basically, a function that combines `fetchers.fetchExperienceEntry` and `fetchers.fetchExperienceEntities` (but still uses them under the hood, just like the hook does).

An ES6 generator could also work nicely for this to yield what we have already fetched (just like the hook does).
But the main point here is that it'd be a single function to get everything we need for `createExperience`.

Spring3:
> So, basically, a function that combines `fetchers.fetchExperienceEntry` and `fetchers.fetchExperienceEntities` (but still uses them under the hood, just like the hook does).

I wonder whether that would be useful though. We can let's say provide a set of filters which devs could provide, but it still might not cover all use cases. That's why I exported 2 functions so that, if wanted, they could use one or both of them, or fetch the entire thing the way they wanted

DanweDE:
> That's why I exported 2 functions so that, if wanted, they could use one or both of them, or fetch the entire thing the way they wanted

I see. I mean, the hook is solving that in an elegant way, it exposes `fetchBySlug` but could also expose a `fetchById` and you don't have to worry about fetching experience entry and referenced entities separately (imagine later we also have to fetch templates, etc. but it's all contained within one method).
Another way to avoid multiple functions for that purpose could be by allowing to pass a cda client query object :thinking:

src/typeguards.ts
```diff
@@ -0,0 +1,31 @@
+import type { Entry } from 'contentful'
+import type { DeprecatedExperience, Experience, ExperienceEntry } from './types'
+
+// @ts-expect-error type incompatibility
+export const isExperienceEntry = (entry: Entry): entry is ExperienceEntry => {
```
Chaoste, line 5: One reason why this breaks TypeScript probably is that we're using the `Entry` type from the CDA library which requires the two methods `toPlainObject()` and `update()` which we probably don't have here if I'm not mistaken. So eventually, this can be fixed by using the correct type for the parameter.
For reference, in the experience builder we have a similar checker in `schemaDetector.ts` where the parameter has the type `Entry` (from cma library not cda) and assesses that it's actually an `ExperienceEntry`.

Screenshot 2023-10-23 at 12 12 33
Screenshot 2023-10-23 at 12 15 31

Chaoste: To avoid all these ugly suppression comments in the file, maybe we should just change the type to `any` or (better) `unknown` until we use a proper type for `Entry` in this codebase.

src/blocks/ExperienceRoot.tsx
```diff
@@ -1,22 +1,44 @@
...
+    throw new Error(
+      `Unsupported mode provided: ${experience.mode}. Supported values: ${supportedModes}`
+    )
+  })
+
+  useEffect(() => {
+    if (supportedModes.includes(mode)) {
+      setMode(mode)
+    }
+  }, [mode])
```
Chaoste, line 35: Can you elaborate what this is and why we need this rendering-based validation of the state variable? As far as I see, `setMode` is only called once with a hardcoded string, so this condition will never be true. So maybe, we can remove this whole effect atm?

src/core/fetchers.ts
```diff
@@ -0,0 +1,151 @@
...
+    assetIds.length > 0 ? client.getAssets({ 'sys.id[in]': assetIds, locale }) : { items: [] },
+  ])
+
+  return {
+    entries: (entriesResponse.items ?? []) as Entry[],
+    assets: (assetsResponse.items ?? []) as Asset[],
+  }
+}
+
+type fetchExperienceEntryArgs = {
```
Chaoste, line 59: The types in this file are all somehow very similar. Any chance we can have shared type and make the "arg types" be based on those.

src/core/fetchers.ts
```diff
@@ -0,0 +1,151 @@
...
+  client: ContentfulClientApi<undefined>
+  experienceTypeId: string
+  locale: string
+  idenifier: {
+    slug?: string
+    id?: string
+  }
```
Chaoste, line 66:
```suggestion
  idenifier: {
    slug: string
  } | {
    id: string
  }
```
I guess there is no use case where we want to allow the user to pass an empty object without any identifier.

src/core/fetchers.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export const fetchExperienceEntry = async ({
+  client,
+  experienceTypeId,
+  locale,
+  idenifier,
```
Chaoste, line 73:
```suggestion
  identifier,
```

src/core/fetchers.ts
```diff
@@ -0,0 +1,151 @@
...
+    slug?: string
+    id?: string
+  }
+}
+
+export const fetchExperience = async ({
+  client,
+  experienceTypeId,
+  locale,
+  idenifier,
```
Chaoste, line 132:
```suggestion
  identifier,
```

src/core/fetchers.ts
```diff
@@ -0,0 +1,151 @@
...
+    content_type: experienceTypeId,
+    locale,
+    ...filter,
+  })
+
+  if (entries.items.length > 1) {
+    throw new Error(
+      `More than one experience with identifier: ${JSON.stringify(idenifier)} was found`
+    )
+  }
```
Chaoste, line 113: Did we have this behavior before? I thought that before we decided to take the first one if there were multiple ones. Not saying that I'm in favour of any of both ways but just wondering whether we consider this a change we need to document.

src/typeguards.ts
Chaoste: Mmmh, looking at the logic in `fetchers` I see that the entry is actually coming from the CDA fetch method and thus correct. So it's the returned type here that is not correct.

src/hooks/useExperienceBuilder.ts
```diff
@@ -21,44 +19,41 @@ type UseExperienceBuilderProps = {
   mode?: ExternalSDKMode
 }
 
+/**
+ * @deprecated This hook is deprecated. We re-designed the SDK to enable client or server side rendering
+ */
```
Chaoste, line 24: I know this is colliding with the suggestion made by @DanweDE but I would still keep this in the long run. Do we imagine every customer to have an advanced use case or why don't we keep the API easy to use for easy use cases?

src/hooks/useExperienceBuilder.ts
```diff
@@ -21,44 +19,41 @@ type UseExperienceBuilderProps = {
...
   return {
+    /**
+     * @deprecated please fetch the experience using `useFetchExperience` hook or fetch the data manually using `fetchers` or `client` and create experience with `createExperience` function
+     *
+     * @example
+     *
+     * import { useFetchExperience } from '@contentful/experience-builder'
+     *
+     * const { fetchBySlug, experience, isFetching } = useFetchExperience({ client, mode })
+     */
```
Chaoste, line 48: I like the level of detail for these deprecation notices :)

src/hooks/useFetchExperience.ts
```diff
@@ -0,0 +1,194 @@
...
+      localeCode: string
+    }): Promise<{ success: boolean; error?: Error; experience?: Experience }> => {
+      setIsFetching(true)
+
+      try {
+        const experienceEntry = await fetchExperienceEntry({
+          client,
+          experienceTypeId,
+          locale: localeCode,
+          idenifier: {
```
Chaoste, line 49:
```suggestion
          identifier: {
```

src/hooks/useFetchExperience.ts
```diff
@@ -0,0 +1,194 @@
...
+  const fetchBySlug = useCallback(
+    async ({
+      experienceTypeId,
+      slug,
+      localeCode,
+    }: {
+      experienceTypeId: string
+      slug: string
+      localeCode: string
+    }): Promise<{ success: boolean; error?: Error; experience?: Experience }> => {
```
Chaoste, line 41: Promises are built for having a `resolve` and `reject` case. What is the advantage of having a custom success and error attribute vs using the native approach (which an un-experience developer probably might expect).

Chaoste (changes requested):
Sorry for this late review but vacation & tech summit made me respond that late. This is really a great contribution and I like the new interfaces that we now offer for SSR, SSG, and other client-rendering frameworks than React.
Thanks for adding all these test cases & deprecation notices and bringing more structure to the project.
I field a few questions but my "request changes" is mainly about blocking thinks like typos in our public methods.

ChidinmaOrajiaku:
@Spring3 You said the downside of option 2 for client-side rendering is it only fetches by slug. But from the implementation, `useFetchExperience` also has a `fetchById` method exposed right?

src/hooks/useExperienceBuilder.ts
ChidinmaOrajiaku: +1. For customers who don't wish to go through the process of pre-fetching the entities and creating an experience to get the experience object, how about we leave this easy setup hook?

src/blocks/ExperienceRoot.tsx
Spring3: Before this change, this useEffect was living in `useExperienceBuilder` hook, which handled the switch between the modes, but after the changes in this PR, I moved it into `ExperienceRoot`. 

I think we will be able to remove it in the next bump of the major version

src/core/fetchers.ts
Spring3: good catch 😓

src/core/fetchers.ts
Spring3: yes, we did. Before this PR, it was part of the `useExperienceBuilderStore` hook.

Also we export these functions, so it makes sense in my opinion to make it atomic, strict and avoid assumptions. These are meant to be just helper functions to save a couple of lines of code. If one of them doesn't fit the use case the devs might have, then they can also fetch the entity in another way, which they prefer most

Spring3: You can see [the code here](https://github.com/contentful/experience-builder/blob/main/src/hooks/useExperienceStore.ts#L127-L134)

src/blocks/ExperienceRoot.tsx
Chaoste: Not a blocker as it doesn't do anything but I still don't get why we need to keep it as it doesn't do anything.
Or do you rather mean that this is just a moved code and thus not part of this PR to clean this up? (which is a reasonable argument)

src/hooks/useExperienceBuilder.ts
Spring3: Not sure I understand, but there is [this new useFetchExperience hook](https://github.com/contentful/experience-builder/blob/feat/support-ssr/src/hooks/useFetchExperience.ts#L22) that we also export, which will be the substitute to the `useExperienceBuilder` hook.

Supporting `useExperienceBuilder` further makes it harder to separate functionality. For example, it required `slug` to be provided to the `ExperienceRoot`. Now that we have added the `id`, should now `ExperienceRoot` also require an `id` to be passed?

Deprecating it allows us to streamline the usage in the next major version as regardless whether used with SSR or SSR or CSR, in all 3 cases entities are going to be fetched before using the `ExpereinceRoot` and we will free our sdk from unreliable checks and assumptions to call the right fetch method - by id or by slug.

src/core/fetchers.ts
Spring3: I don't think it works

Screenshot 2023-10-23 at 14 52 53

src/hooks/useFetchExperience.ts
Spring3: Good call. The shape of the response has been inherited from previous implementation. I think we can change it actually

Chaoste (approved):
Thanks for addressing the comments that quickly - the rest is tentative so here is my approval :)

src/hooks/useFetchExperience.ts
Spring3: Applied

dimitrycf:
nit: forgetting to specify `identifier` will produce a pretty cryptic error which doesn't hint to adding `identifier` param

image

src/hooks/useFetchExperience.ts
```diff
@@ -0,0 +1,173 @@
...
+    async ({
+      experienceTypeId,
+      slug,
+      localeCode,
+    }: {
+      experienceTypeId: string
+      slug: string
+      localeCode: string
+    }): Promise<Experience | undefined> => {
+      setIsFetching(true)
```
Chaoste, line 42: One thing that I learned today can save a few lines in this method again:
```
setIsFetching(true);
try {
  // do anything else and even return things
} finally {
  // ensure that the boolean gets reset before the function exits
  setIsFetching(false);
}
```

src/hooks/useFetchExperience.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        setExperience(experience)
+        setIsFetching(false)
+
+        return experience
+      } catch (e) {
+        handleError(errorMessagesWhileFetching.experienceReferences, e)
+        setIsFetching(false)
+        throw e
+      }
```
Chaoste, line 91: You can basically move all into one single try/catch/finally block like this. Personally, I think it's now easier to follow the things being executed in the function as the critical logic is all happening inside one `try` block with no differing indentations.

```suggestion
      setIsFetching(true)
      try {
        const experienceEntry = await fetchExperienceEntry({
          client,
          experienceTypeId,
          locale: localeCode,
          identifier: {
            slug,
          },
        })
         if (!experienceEntry) {
           throw  new Error(`No experience entry with slug: ${slug} exists`)
         }
         const { entries, assets } = await fetchExperienceEntities({
          client,
          experienceEntry,
          locale: localeCode,
        })
        const experience = createExperience({
          experienceEntry,
          referencedAssets: assets,
          referencedEntries: entries,
          locale: localeCode,
          mode,
        })
        setExperience(experience)
        return experience
      } catch (error) {
        handleError(errorMessagesWhileFetching.experience, error)
        throw error
      } finally {
        setIsFetching(false)
      }
```

src/hooks/useFetchExperience.ts
Chaoste: See my comment below how this helps making the function way shorter :)

dimitrycf:
### proposal: to make `fetchExperienceEntry()` to throw on not-found of the entry

Because (as described in the comment above) `fetchExperienceEntities()` expects valid (truthy) `experienceEntry` argument and otherwise throws. It may be semantically consistent to make `fetchExperienceEntry()` to throw too (currently it will return `undefined`) .

We can of course also add special if-statement to the documentation/SSR sample. However taking into accoutn that `fetchExperienceEntities()` will throw anyways, I can't see benefit from this.


image

dimitrycf:
### subjective: fetchExperienceEntries() and fetchExperienceEntities()  are easy to confuse when glancing over with eyes

 
 - maybe we can improve readability (glanceability) of those methods? 


eg.  (or absolutely anything else what would prevent those methods from looking alike)
```

fetchExperienceEntries() -> fetchEntries()
fetchExperienceEntities() ->fetchEntities()
```

image

src/hooks/useFetchExperience.ts
Spring3: It used to work like this, but I wanted to have granularity in terms of what we print to the console. For example, when the function fetches the entry, I wanted to print to the console the error which says "Failed to fetch the experience entry", but for example, when the function fetches the references in the entry, I wanted to print "Failed to fetch the experience entities" or similar.

One try catch resulted in a generic print log, which of course could be resolved if I created different errors types, but this PR already seemed quite big even without it

src/hooks/useFetchExperience.ts
Spring3: But then there is a question - if line 53 works, will `isFetching` be `true` or `false`

Spring3:
> ### When `fetchExperienceEntities()` receives `undefined` as `experienceEntry` we can provide more accurate error
> * [ ]  resolved
> 
> Currently the error is
> 
> ```
> Failed to fetch experience entities. Provided "experienceEntry" does not match experience entry schema
> ```
> 
> However in this case it's not because schema doesn't match, but because `experienceEntry` is undefined
> 
> It would be more user friendly to have early guard which checks for the the `experienceEntry` argument to not be falsy
> 
> image

technically, `undefined` also doesn't match the schema, so the message would fit. I understand your point of view, but I think it is good enough for a developer to go and investigate and realise that their entry is undefined

src/hooks/useFetchExperience.ts
Chaoste: It will be `false`. `Finally` is always called and that's why it's perfect for this kind of "clean-ups" / "resets"

Chaoste: I noticed that my proposal didn't cover the different error messages. I just updated it with these two lines: 
```
const errorMessageKey = experienceEntry ? 'experienceReferences' : 'experience'
handleError(errorMessagesWhileFetching[errorMessageKey], error)
```
So now it does exactly the same as above with differing error messages :)

src/hooks/useFetchExperience.ts
Chaoste: You can still have multiple try/catch blockes within the try/finally block but I agree that this wouldn't be an actual code improvement :)
I addressed this in the comment below, so let me know if that works for you :)

src/hooks/useFetchExperience.ts
Spring3: Validated. It indeed still triggers `finally`. Well good, then this simplifies things. At first I actually had it just like you described, but then I started doubting whether finally would get triggered even after the error is thrown within `catch`

Screenshot 2023-10-24 at 11 54 55

Spring3:
> ### proposal: to make `fetchExperienceEntry()` to throw on not-found of the entry
> * [ ]  resolved
> 
> Because (as described in the comment above) `fetchExperienceEntities()` expects valid (truthy) `experienceEntry` argument and otherwise throws. It may be semantically consistent to make `fetchExperienceEntry()` to throw too (currently it will return `undefined`) .
> 
> We can of course also add special if-statement to the documentation/SSR sample. However taking into accoutn that `fetchExperienceEntities()` will throw anyways, I can't see benefit from this.
> 
> image

This function returns exactly the same output as `client.getEntries()`. I don't think that throwing an error in case `undefined` is fetched is necessary. Even if `undefined` is returned, the next function `fetchExperienceEntities` will throw due to the entry being `undefined`

Spring3:
> ### subjective: fetchExperienceEntries() and fetchExperienceEntities() are easy to confuse when glancing over with eyes
> * [ ]  resolved
> 
> eg. even in github comment they look similar
> 
> image
> 
> * maybe we can improve readability (glanceability) of those methods?
> 
> eg. (or **absolutely anything else** what would prevent those methods from looking alike)
> 
> ```
> 
> fetchExperienceEntries() -> fetchExperience()
> fetchExperienceEntities() ->fetchEntities()
> ```

renamed `fetchExperienceEntities` to `fetchReferencedEntities`

dimitrycf:
### When doing SSR via getServerSideProps we get warning due to `useLayoutEffect()` 

 - [ ] resolved

image


This is due to this `useLayoutEffect()` https://github.com/contentful/experience-builder/blob/main/src/hooks/useStyleTag.ts#L17 

(which was added by Sofia and Lorenz a while ago).


Maybe to differentiate SSR vs CSR we can use the same trick as @Spring3  when he hacked detecting of SSR environment by checking `window !== typeof 'undefined' ? 

image

Spring3:
> by checking `window !== typeof 'undefined' ?

Unfortunately, we can not do that, because in that case react would complain due to "conditional use of hooks" which is prohibited

dimitrycf:
> > by checking `window !== typeof 'undefined' ?
> 
> Unfortunately, we can not do that, because in that case react would complain due to "conditional use of hooks" which is prohibited

Are you sure about that?

Spring3:
I don't know the internals of react that much to be sure how it 

> > > by checking `window !== typeof 'undefined' ?
> > 
> > 
> > Unfortunately, we can not do that, because in that case react would complain due to "conditional use of hooks" which is prohibited
> 
> Are you sure about that?
> 
> ps. because there is no conditional use of hooks as far as I can see 🤷 (hook order will never change when executed in the frontend). The "conditional" use of hooks problem occurs IFAIK because under the hood React assigns numbers to all useHook calls. And later identifies all those hooks by those numbers. If the numbers are consistent within the app, then it should be ok.

Does it complain if you replace `useLayoutEffect` with `useEffect`? If it doesn't complain about `useEffect` then I can use that one instead of `useLayoutEffect` entirely

Spring3:
The question is - would it remove the warning or not. Is there a chance you could check it?

dimitrycf:
### The sample code doesn't handle case when the experience is missing

The 404 thrown upon on slug being empty doesn't seem to run ever 
(as if the slug is empty then `getServerSideProps()` will throw and actual `Home()` component will never run)

image



So we need somehow add handling of not-found composition, and pass it to the compoennt.

Eg. 

image


And the 404 guard should be before `createExperience()`  call 

image

dimitrycf:
> The question is - would it remove the warning or not. Is there a chance you could check it?

The `useCorrectEffect()` removes the warning. Ergo, using `useEffect()` removes the warning too.

src/hooks/useStyleTag.ts
```diff
@@ -14,7 +14,7 @@ import { CSSProperties } from '../types'
 export const useStyleTag = ({ styles, nodeId }: { styles: CSSProperties; nodeId?: string }) => {
   const [className, setClassName] = useState('')
 
-  useLayoutEffect(() => {
+  useEffect(() => {
```
DanweDE, line 17: Wasn't there a reason we use `useLayoutEffect` here?

>> by checking `window !== typeof 'undefined' ?
>
> Unfortunately, we can not do that, because in that case react would complain due to "conditional use of hooks" which is prohibited

I think the correct way of doing that would be outside the component close to the import, basically

```
import { useLayoutEffect, useEffect } from 'react';

const useRightEffect = typeof window === 'undefined' ? useEffect : useLayoutEffect
```

and then within the component, you simply use the `useRightEffect`.

Spring3: I don't think that there was a reason to use `useLayoutEffect`. As Dimitry pointed out above, it was a suggestion from my side on top of Sofia's PR, because I thought it would better fit injecting the styles, but in the current situation - both will work the same way. The difference between useEffect and useLayoutEffect is not significant in order to try to save it this way.

dimitrycf:
## When missing a registered component, and rendering via SSR cryptic errors can appear

 - what do you think of SDK detecting missing components and throwing an early error message?


Here's the situation I encountered:

#### My page should have rendered (in preview mode) like this:
 - purple edged section
 - button in the center of a section
image



#### However, at one point of time I broke something and it looked like this and showed error
 - notice it still renders (has purple edged section)
 - it is missing section with the button  (well actually just the button; section turns out is there but it's white)
 - We have two hydration errors (which whilst point at button are still cryptic)


error 1
image

error 2
image



So first I thought maybe I am mixing up preview/delivery modes and hence the button is missing. But it turned out I simply forgot to register `MyButton` component.

 - taking into account that there's quite a boilerplate code, it may be easy to miss out on a component (also during development it is fair to assume that some components will be added/removed/commented-out/updated/etc)
 - hydration error can be slightly misleading. So maybe we can do better? 
 - is this related to the [ticket that
](https://contentful.atlassian.net/browse/SPA-1509)
image

dimitrycf (approved):
Fantastic work on refactoring the SDK! 

The new API is nicer and cleaner ❤️ 

Approving this, to unblock, however I still believe that `undefined` as experience is not semantically the same as "schema mismatch". And a dedicated message would make up for better DX. https://github.com/contentful/experience-builder/pull/136#issuecomment-1776858808

Spring3:
> what do you think of SDK detecting missing components and throwing an early error message?

I think it might be good to have some way of validation to ensure that all the necessary components are in place, but I am not really a fan of throwing an error and rendering nothing when there is a chance to render at least some portion of the page.

I think during editor mode, such validation functionality might become especially useful as we will be able to highlight such problematic spots with missing components, which might not be easy to discover

Spring3:
> So first I thought maybe I am mixing up preview/delivery modes and hence the button is missing. But it turned out I simply forgot to register MyButton component.

> taking into account that there's quite a lot of boilerplate code, and it may be easy to miss out on a component (also during development it is fair to assume that some components will be added/removed/commented-out/updated/etc)
hydration error can be slightly cryptic and semantically it's not pointing to the root cause. So maybe we can do better and present user with the root cause? (ie. missing component registration?)
is this related to the [SPA-1509](https://contentful.atlassian.net/browse/SPA-1509)

I agree with all this, however, it sounds like the next step on top of this PR. The plan is to give it in hands of the customers and then wait a couple of weeks to hear what they have to say and then release a major version, which can be a good base for the further improvements related to server side rendering features

PR #138, opened 2023-10-12
feat: change default max width size to none 

ChidinmaOrajiaku:


src/core/definitions/variables.ts
```diff
@@ -90,7 +90,7 @@ export const builtInStyles: Record<
     type: 'Text',
     group: 'style',
     description: 'The max-width of the section',
-    defaultValue: '100%',
+    defaultValue: 'none',
```
Spring3, line 93: should we maybe set the default value to be `initial`? I thought that it might work better if someone used max width via `builtInStyles`

ChidinmaOrajiaku: Good idea. Done

Spring3 (approved):

PR #140, opened 2023-10-20
fix: adapt container to margins and paddings [SPA-1523]

ChidinmaOrajiaku:
- Use border-box sizing for container components
- Adapt container width to margins


SofiaMargariti (approved):
👍

PR #143, opened 2023-10-25
feat: introduce a prefix to our global variables to avoid overlaps [SPA-1587]

Chaoste:
Based on https://github.com/contentful/experience-builder/pull/142

Use prefixes for our global variables coming from F36.


src/styles/variables.css
```diff
@@ -2,41 +2,44 @@
...
-  --gray500: #67728a;
-  --gray600: #5a657c;
-  --gray700: #414d63;
-  --gray800: #1b273a;
-  --gray900: #111b2b;
-  --red200: #ffe0e0;
-  --red800: #7f0010;
-  --color-white: #ffffff;
-  --glow-primary: 0px 0px 0px 3px #e8f5ff;
+  --f36-blue100: #e8f5ff;
```
SofiaMargariti, line 12: It might be inevitable at some point these variables to diverge from forma. Also if forma later on decides to adopt the `--f36` prefix as well we'll have the same issue (Probably unlikely but you never know).
Maybe we should use a prefix specific to Exp Builder and avoid mentions of forma all together?

Chaoste: Mmmh, I'm in favor of sticking to our built-in design library and forcing us to use these colors, spacings, font sizes, etc. - designers and UX engineers will probably agree on this. So, that's why I would like to keep this prefix specific for the F36 tokens we should reuse. I can change it to `--experience-builder-f36` but still I don't want us to ever change those values to avoid such issues again (see the current text size of the SDKs error screen).

Chaoste: I would not try to think here too general/ broad but keep it specific as we can quickly change this prefix later on and so far, we don't anticipate any case where we actually want to use those variables outside of the SDK.

SofiaMargariti: So if we need to introduce a new variable that is not used in forma later on, would we use `--f36` again or introduce a new prefix then?

Spring3: If we make a decision to change the prefix, let's also make sure that the prefix is short to save a bit on the size of the css file. I know that at this point it is micro optimisation, but it might help us save in the long run

Chaoste: Mmmh, I don't anticipate such a scenario but if so, then we should have a non f36 prefix :)

Chaoste: Following the conclusion in Slack, I changed the prefix `--exp-builder` and added a note to the section where they are coming from.

src/blocks/ErrorBoundary.css
```diff
@@ -1,18 +1,18 @@
...
   margin-top: 0;
-  font-size: var(--font-size-l);
+  font-size: var(--exp-builder-font-size-l);
 }
 
 .cf-error-message .more-details {
   cursor: pointer;
-  color: var(--blue700);
+  color: var(--exp-builder-blue700);
 }
```
Spring3, line 18: would be good to move this file to `styles` folder as well for consistency

Spring3 (approved):

ChidinmaOrajiaku (approved):

PR #145, opened 2023-10-26
feat: initial setup for experience-builder-types [SPA-1395]

anwaar931:
- Will export Experience builder types

### DIFF
--- a/packages/experience-builder-types/package.json
+++ b/packages/experience-builder-types/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "experience-builder-types",
+  "version": "1.0.0",
+  "description": "Experience builder types",
+  "main": "src/index.ts",
+  "author": "Contentful GmbH",
+  "license": "MIT"
+}
--- a/packages/experience-builder-types/README.md
+++ b/packages/experience-builder-types/README.md
@@ -1,6 +1,6 @@
-# `experience-builder-types`
+# `@contentful/experience-builder-types`
 
-> TODO: description
+Type definitions and constants for the Contentful Experience Builder.
 
 ## Usage
 
--- a/packages/experience-builder-types/__tests__/experience-builder-types.test.js
+++ b/packages/experience-builder-types/__tests__/experience-builder-types.test.js
@@ -1,7 +1,0 @@
-'use strict';
-
-const experienceBuilderTypes = require('..');
-const assert = require('assert').strict;
-
-assert.strictEqual(experienceBuilderTypes(), 'Hello from experienceBuilderTypes');
-console.info('experienceBuilderTypes tests passed');
--- a/packages/experience-builder-types/src/experience-builder-types.js
+++ b/packages/experience-builder-types/src/experience-builder-types.js
@@ -1,7 +1,0 @@
-'use strict';
-
-module.exports = experienceBuilderTypes;
-
-function experienceBuilderTypes() {
-  return 'Hello from experienceBuilderTypes';
-}


.prettierrc
```diff
@@ -4,5 +4,5 @@
   "useTabs": false,
   "singleQuote": true,
   "bracketSameLine": true,
-  "semi": false
+  "semi": true
```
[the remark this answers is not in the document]
Chaoste: Also it makes this PR impossible to review. If you really want to add this I would used stacked PRs to allow reviewing any other changes in this PR that are not caused by the auto-formatter.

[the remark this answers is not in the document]
anwaar931: @Chaoste I didn't added this change, but I guess after merge I am getting existed changes
https://github.com/contentful/experience-builder/blob/main/.prettierrc#L7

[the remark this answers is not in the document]
anwaar931: Actually no, this change you pointed out is not in my PR. :/  Merge is fine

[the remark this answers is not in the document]
Chaoste: Ah, I see now that @Spring3 already added this. Maybe it was due to a missing merge commit that it was shown as a change in your PR 🤷‍♀️

packages/experience-builder-types/src/types.ts
```diff
@@ -0,0 +1,330 @@
+/**
+ * danv:
+ * NOTE!! The code commented here will be used in future. We commented it out to remove not yet fully unsupported parts
+ */
+
+import type { ContentfulClientApi, Entry } from 'contentful';
```
Chaoste, line 6: So we still need this file?

packages/experience-builder-types/src/constants.ts
```diff
@@ -0,0 +1,72 @@
+import { InternalSDKMode, SchemaVersions } from './types'
+
+export const SCROLL_STATES = {
+  Start: 'scrollStart',
+  IsScrolling: 'isScrolling',
+  End: 'scrollEnd',
+}
```
Chaoste, line 7: As written per DM, it would be great to expose a type based on this constant like this:
```
export type ScrollState = typeof SCROLL_STATES[keyof typeof SCROLL_STATES];
```

Chaoste: Similar for the other object literals below. This way, we can replace the old & deprecated types with those new ones.

Chaoste: (or is this supposed to be a separate PR afterwards to reduce code changes?)

Chaoste (changes requested):
Nice changes! I only have a few questions but except for that it looks good already.

Reminder: We should also add a `catalog-info.yml` that exposes both packages to backstage.

packages/experience-builder-types/src/constants.ts
anwaar931: I added them in next commit, lost them previously because of merge conflicts in types.ts

I added them in types.ts of experience-builder-types package

Chaoste (approved):

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     executor: docker-with-node
     steps:
       - with-workspace
-      - run: yarn test:coverage
+      - run: yarn test
```
Spring3, line 66: I thought that the next step (on line 67) sends the coverage somewhere. Is it not the case?

packages/experience-builder-sdk/jest.config.js
```diff
@@ -10,4 +10,7 @@ export default {
     '^.+\\.svg$': '<rootDir>/test/fileMock.ts',
     '^.+\\.svg\\?react$': '<rootDir>/test/fileMock.ts',
   },
+  transform: {
+    '\\.[jt]sx?$': 'ts-jest',
```
Spring3, line 14: What problem is it solving?

packages/experience-builder-sdk/src/constants.ts
```diff
@@ -1,34 +1,6 @@
...
-  'cfBackgroundImageScaling',
-  'cfBackgroundImageAlignment',
-  // For backwards compatibility
-  // we need to keep those in this constant array
-  // so that omit() in <VisualEditorBlock> and <CompositionBlock>
-  // can filter them out and not pass as props
-  'cfBackgroundImageAlignmentVertical',
-  'cfBackgroundImageAlignmentHorizontal',
-];
+export * from '@contentful/experience-builder-types/dist/src/constants';
```
Spring3, line 3: Are we sure we want constants to be in "types" package?

packages/experience-builder-sdk/src/types.ts
```diff
@@ -1,330 +1 @@
...
-  /**
-   * @deprecated please don't use
-   */
-  experienceTypeId: string;
-  /**
-   * @deprecated please don't use
-   */
-  mode: InternalSDKMode;
-}
+export * from '@contentful/experience-builder-types/dist/src/types';
```
Spring3, line 1: should we just delete this `types.ts` file in general and import directly from types everywehre?

packages/experience-builder-sdk/src/typeguards.ts
```diff
@@ -1,29 +1 @@
...
-  experience: Experience | DeprecatedExperience
-): experience is DeprecatedExperience => {
-  return (
-    experience &&
-    {}.hasOwnProperty.call(experience, 'client') &&
-    {}.hasOwnProperty.call(experience, 'experienceTypeId') &&
-    {}.hasOwnProperty.call(experience, 'mode')
-  );
-};
+export { isDeprecatedExperience, isExperienceEntry } from '@contentful/experience-builder-types';
```
Spring3, line 1: should we delete this `typeguards` file entirely and import things from types directly?

packages/experience-builder-types/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "test": "NODE_ENV=test jest"
+  },
+  "dependencies": {
+    "@contentful/visual-sdk": "^1.0.0-alpha.26",
+    "contentful": "^10.3.0"
```
Spring3, line 22: Q: Should both of these be peer dependencies?

packages/experience-builder-types/src/constants.ts
```diff
@@ -0,0 +1,72 @@
...
+  // we need to keep those in this constant array
+  // so that omit() in <VisualEditorBlock> and <CompositionBlock>
+  // can filter them out and not pass as props
+  'cfBackgroundImageAlignmentVertical',
+  'cfBackgroundImageAlignmentHorizontal',
+]
+
+export const supportedModes: InternalSDKMode[] = ['delivery', 'preview', 'editor']
+// this is the array of version which currently LATEST_SCHEMA_VERSION is compatible with
+export const compatibleVersions: SchemaVersions[] = ['2023-08-23', '2023-09-28']
```
Spring3, line 72: I think we should keep constants, especially things like `compatibleVersion` in the sdk, not in types package

packages/experience-builder-types/test/__fixtures__/composition.ts
```diff
@@ -0,0 +1,173 @@
+import { LATEST_SCHEMA_VERSION } from '../../src';
+import { Composition, ExperienceEntry, SchemaVersions } from '../../src/types';
```
Spring3, line 2: should we delete fixtures from this package? I don't think they belong here

Spring3 (changes requested):

.circleci/config.yml
anwaar931: I removed it as it will run `test:coverage` for each package, but now added it back.

packages/experience-builder-sdk/src/types.ts
anwaar931: I did this so we don't have to change it everywhere and then we can still have local types just for `ExperienceBuilder` which are not meant to be exported. So I think we should keep it.

packages/experience-builder-sdk/jest.config.js
anwaar931: So as we are importing types from `experience-builder-types` repo. It exposes types in dist folder with js file, We need to tell jest to compile them as ES6, as they are not commonJs js files.

packages/experience-builder-types/test/__fixtures__/composition.ts
anwaar931: I moved our `EntityStore` which we are extending from VisualSdkEntityStore to `types` package. So is it's unit test file. That's why we need these fixtures.

packages/experience-builder-sdk/jest.config.js
Spring3: Is there a way to build both - es6 and cjs files? Maybe vite provides such ability

anwaar931: Not sure about vite, but swc builds for both and its lightweight than vite

packages/experience-builder-types/test/__fixtures__/composition.ts
Spring3: Is there a reason why we need EntityStore to be moved into the types package? It is only used by the sdk, no?

anwaar931: We are using EntityStore in `Experience` type. So moved it along. Otherwise I think there will be circular dependency issue.

Spring3: But I think `Experience` uses the type from `visual-sdk`, isn't it? The EntityStore is just a build up on top of `EntityStore` class that is imported from `visual-sdk`

anwaar931: `Experience` uses our `EntityStore` which we have created by extending `EntityStore` from `visual-sdk`

https://github.com/contentful/experience-builder/blob/main/packages/experience-builder-sdk/src/types.ts#L310

Spring3: In that case, can we keep the experience type on the sdk side?

in that case we might need to make Experience a generit type. I wouldn't pull the entityStore implementation into the types package

```
type Experience<T extends EntityStore> = {
   entityStore: T,
   ...
}
```

anwaar931: Added these changes.

Spring3 (approved):

PR #149, opened 2023-11-03
feat: [] send also the margins with the coordinates

loweisz:
To calcuate the dnd indicator also respecting the margins of the element, we need to also send the marings with every element. 

There will be a matching PR on the other side to support this.

### DIFF
--- a/packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
+++ b/packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
@@ -5,17 +5,28 @@ import { sendMessage } from './sendMessage';
 export class MouseOverHandler {
   private currentHoveredElementId: string | null = null;
 
+  private getMargins = (element: HTMLElement) => {
+    const styles = window.getComputedStyle(element);
+    const top = parseInt(styles.marginTop);
+    const bottom = parseInt(styles.marginBottom);
+    const left = parseInt(styles.marginLeft);
+    const right = parseInt(styles.marginRight);
+
+    return { top, bottom, left, right };
+  };
+
   private getFullCoordinates = (element: HTMLElement) => {
     const validChildren = Array.from(element.children).filter(
       (child) => child instanceof HTMLElement && child.dataset.cfNodeBlockType === 'block'
     );
 
     const { left, top, width, height } = element.getBoundingClientRect();
+    const margins = this.getMargins(element);
 
     const childrenCoordinates = validChildren.map((child) => {
       const { left, top, width, height } = child.getBoundingClientRect();
 
-      return { left, top, width, height };
+      return { left, top, width, height, margins };
     });
 
     return {
@@ -23,6 +34,7 @@ export class MouseOverHandler {
       top,
       width,
       height,
+      margins,
       childrenCoordinates,
     };
   };


packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
```diff
@@ -5,24 +5,36 @@ import { sendMessage } from './sendMessage';
 export class MouseOverHandler {
   private currentHoveredElementId: string | null = null;
 
+  private getMargins = (element: HTMLElement) => {
+    const styles = window.getComputedStyle(element);
+    const top = parseInt(styles.marginTop);
```
Chaoste, line 10: We can expect here not only pixel values but also a variety of others. I assume this breaks this logic

Chaoste: `/^\d{1,}(px|cm|mm|in|pt|pc|em|ex|ch|rem|vw|vh|vmin|vmax|%)?$/`

Chaoste (changes requested):

packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
Chaoste: As discussed via DM, we [get always pixel values here](https://stackoverflow.com/questions/48016082/window-getcomputedstyle-always-returns-measurement-in-pixels)

Chaoste (approved):

packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
```diff
@@ -5,24 +5,36 @@ import { sendMessage } from './sendMessage';
 export class MouseOverHandler {
   private currentHoveredElementId: string | null = null;
 
+  private getMargins = (element: HTMLElement) => {
+    const styles = window.getComputedStyle(element);
```
Spring3, line 9: Can this code be executed on server? If so, please add the undefined check. Otherwise, it can break on SSR/SSG

loweisz:
> > There will be a matching PR on the other side to support this.
> 
> Please make sure it functions in retro-compatible way and still works well if margins haven't been sent. To avoid a case when it would break for customers with sdk < next

I marked them as potentially undefined in the user_interface, we handle the case that they are not provided 👍

packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
loweisz: That's a valid point 👍 I will add this

packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
```diff
@@ -5,24 +5,37 @@ import { sendMessage } from './sendMessage';
 export class MouseOverHandler {
   private currentHoveredElementId: string | null = null;
 
+  private getMargins = (element: HTMLElement) => {
+    if (!window) return undefined;
```
Spring3, line 9:
```suggestion
    if (typeof window !== 'undefined') return undefined;
```

I may be wrong, but I think the current check would still cause an error during SSR.

Spring3 (approved):
No other requests aside from the one I left above :)

packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
loweisz: Oh, I did not know this, thanks 👍

PR #150, opened 2023-11-03
Fix: selected component jumps around on scroll [ALT-62]

SofiaMargariti:
The canvas interaction tooltip was jumping around to previous selected components when scrolling the page. It turns out this was due to the onScroll listener being registered multiple times - one for each time the selected node changed.
In order to avoid this I am switching the selected node to be Ref which won't trigger the useEffect to rerun on change, but it will still have the latest value when the listener runs.
See also this SO question: https://stackoverflow.com/questions/55565444/how-to-register-event-with-useeffect-hooks

#### Before

https://github.com/contentful/experience-builder/assets/2773012/2ba384d5-ed1a-4dc8-b6c3-adf023dba865

#### After

https://github.com/contentful/experience-builder/assets/2773012/a1673f8f-25de-4715-911a-3df6ab916872

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
@@ -63,7 +63,7 @@ export function VisualEditorContextProvider({
   const [dataSource, setDataSource] = useState<CompositionDataSource>({});
   const [unboundValues, setUnboundValues] = useState<CompositionUnboundValues>({});
   const [isDragging, setIsDragging] = useState(false);
-  const [selectedNodeId, setSelectedNodeId] = useState<string>('');
+  const selectedNodeId = useRef<string>('');
   const [locale, setLocale] = useState<string>(initialLocale);
 
   const entityStore = useRef<EditorModeEntityStore>(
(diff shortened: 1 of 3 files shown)


packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
```diff
@@ -187,16 +187,11 @@ export function VisualEditorContextProvider({
           }
           break;
         }
-        case INCOMING_EVENTS.SelectedComponentChanged: {
```
Spring3, line 190: Are you sure we don't need it?

It looks like we send this event when [bound node is updated with changes via slide in editor](https://github.com/contentful/user_interface/blob/master/src/javascripts/features/experience-builder/contexts/useCompositionCanvasSubscriber.ts#L301)

SofiaMargariti: I am removing it here: https://github.com/contentful/user_interface/pull/18193/files we have two events for the same purpose. The user_interface PR will need to go first and I've tested it with v2.6.2 sdk version and I didn't see any issues

Spring3: Ah, I see it now. Thanks for pointing it out

Spring3 (approved):

packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
```diff
@@ -272,7 +267,11 @@ export function VisualEditorContextProvider({
       window.removeEventListener('scroll', onScroll);
       clearTimeout(timeoutId);
     };
-  }, [mode, selectedNodeId]);
+  }, [mode]);
+
+  const setSelectedNodeId = (nodeId: string) => {
```
Chaoste, line 272: Just want to highlight that this means that changing the selected node doesn't enforce a rerender. So if there is a case where we solely change the selected node but nothing else, we might fail updating subsequent components or UI.

Chaoste (approved):
Just added a callout as this could case a bug later on, so we are aware of this new introduced behaviour.

PR #151, opened 2023-11-03
feat: render design component within an experience [SPA-1583]

ChidinmaOrajiaku:
Replace & render design components within an experience in the editor mode

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
@@ -2,10 +2,13 @@ import React, { useMemo } from 'react';
 
 import type { UnresolvedLink } from 'contentful';
 import omit from 'lodash.omit';
-import { EntityStore } from '@contentful/visual-sdk';
+import { EntityStore } from '../core/EntityStore';
 
 import { CF_STYLE_ATTRIBUTES, CONTENTFUL_CONTAINER_ID, CONTENTFUL_SECTION_ID } from '../constants';
-import { getComponentRegistration } from '../core/componentRegistry';
+import {
+  createDesignComponentRegistration,
+  getComponentRegistration,
+} from '../core/componentRegistry';
 import { buildCfStyles } from '../core/stylesUtils';
 import { ResolveDesignValueType } from '../hooks/useBreakpoints';
 import { useStyleTag } from '../hooks/useStyleTag';
@@ -19,6 +22,8 @@ import type {
 } from '../types';
 import { ContentfulContainer } from './ContentfulContainer';
 import { transformContentValue } from './transformers';
+import { resolveDesignComponent } from '../core/preview/designComponentUtils';
+import { DesignComponent } from './DesignComponent';
 
 type CompositionBlockProps = {
   node: CompositionNode;
@@ -31,7 +36,7 @@ type CompositionBlockProps = {
 };
 
 export const CompositionBlock = ({
-  node,
+  node: rawNode,
   locale,
   entityStore,
   dataSource,
@@ -39,10 +44,29 @@ export const CompositionBlock = ({
   breakpoints,
   resolveDesignValue,
 }: CompositionBlockProps) => {
-  const componentRegistration = useMemo(
-    () => getComponentRegistration(node.definitionId as string),
-    [node]
-  );
+  const node = useMemo(() => {
+    const isDesignComponent = rawNode.definitionId.startsWith('DesignComponent');
+
+    return isDesignComponent
+      ? resolveDesignComponent({
+          node: rawNode,
+          entityStore,
+        })
+      : rawNode;
+  }, [entityStore, rawNode]);
+
+  const componentRegistration = useMemo(() => {
+    const registeration = getComponentRegistration(node.definitionId as string);
+    const isDesignComponent = node.definitionId.startsWith('DesignComponent');
+
+    if (isDesignComponent && !registeration) {
+      return createDesignComponentRegistration({
+        definitionId: node.definitionId as string,
+        component: DesignComponent,
+      });
+    }
+    return registeration;
+  }, [node]);
 
   const nodeProps = useMemo(() => {
     if (!componentRegistration) {
@@ -59,14 +83,14 @@ export const CompositionBlock = ({
         case 'BoundValue': {
           const [, uuid, ...path] = variable.path.split('/');
           const binding = dataSource[uuid] as UnresolvedLink<'Entry' | 'Asset'>;
-          const value = entityStore?.getValue(binding, path.slice(0, -1));
+          const value = entityStore?.getFieldValue(binding, path.slice(0, -1));
           const variableDefinition = componentRegistration.definition.variables[variableName];
           acc[variableName] = transformContentValue(value, variableDefinition);
           break;
         }
         case 'UnboundValue': {
           const uuid = variable.key;
-          acc[variableName] = unboundValues[uuid]?.value;
+          acc[variableName] = (entityStore?.unboundValues || unboundValues)[uuid]?.value;
           break;
         }
         default:
--- a/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
@@ -83,7 +83,7 @@ export const CompositionBlock = ({
         case 'BoundValue': {
           const [, uuid, ...path] = variable.path.split('/');
           const binding = dataSource[uuid] as UnresolvedLink<'Entry' | 'Asset'>;
-          const value = entityStore?.getFieldValue(binding, path.slice(0, -1));
+          const value = entityStore?.getValue(binding, path.slice(0, -1));
           const variableDefinition = componentRegistration.definition.variables[variableName];
           acc[variableName] = transformContentValue(value, variableDefinition);
           break;
--- a/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
@@ -63,15 +63,15 @@ export const CompositionBlock = ({
   }, [entityStore, isDesignComponent, rawNode]);
 
   const componentRegistration = useMemo(() => {
-    const registeration = getComponentRegistration(node.definitionId as string);
+    const registration = getComponentRegistration(node.definitionId as string);
 
-    if (isDesignComponent && !registeration) {
+    if (isDesignComponent && !registration) {
       return createDesignComponentRegistration({
         definitionId: node.definitionId as string,
         component: DesignComponent,
       });
     }
-    return registeration;
+    return registration;
   }, [isDesignComponent, node.definitionId]);
 
   const nodeProps = useMemo(() => {
--- a/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
@@ -18,12 +18,14 @@ import type {
   CompositionNode,
   CompositionUnboundValues,
   CompositionVariableValueType,
+  ExperienceEntry,
   StyleProps,
 } from '../types';
 import { ContentfulContainer } from './ContentfulContainer';
 import { transformContentValue } from './transformers';
 import { resolveDesignComponent } from '../core/preview/designComponentUtils';
 import { DesignComponent } from './DesignComponent';
+import { checkIfDesignComponent } from '../utils';
 
 type CompositionBlockProps = {
   node: CompositionNode;
@@ -33,6 +35,7 @@ type CompositionBlockProps = {
   entityStore?: EntityStore;
   breakpoints: Breakpoint[];
   resolveDesignValue: ResolveDesignValueType;
+  usedComponents: ExperienceEntry['fields']['usedComponents'];
 };
 
 export const CompositionBlock = ({
@@ -43,21 +46,24 @@ export const CompositionBlock = ({
   unboundValues,
   breakpoints,
   resolveDesignValue,
+  usedComponents,
 }: CompositionBlockProps) => {
-  const node = useMemo(() => {
-    const isDesignComponent = rawNode.definitionId.startsWith('DesignComponent');
+  const isDesignComponent = useMemo(
+    () => checkIfDesignComponent({ componentId: rawNode.definitionId, usedComponents }),
+    [rawNode.definitionId, usedComponents]
+  );
 
+  const node = useMemo(() => {
     return isDesignComponent
       ? resolveDesignComponent({
           node: rawNode,
           entityStore,
         })
       : rawNode;
-  }, [entityStore, rawNode]);
+  }, [entityStore, isDesignComponent, rawNode]);
 
   const componentRegistration = useMemo(() => {
     const registeration = getComponentRegistration(node.definitionId as string);
-    const isDesignComponent = node.definitionId.startsWith('DesignComponent');
 
     if (isDesignComponent && !registeration) {
       return createDesignComponentRegistration({
@@ -66,7 +72,7 @@ export const CompositionBlock = ({
       });
     }
     return registeration;
-  }, [node]);
+  }, [isDesignComponent, node.definitionId]);
 
   const nodeProps = useMemo(() => {
     if (!componentRegistration) {
@@ -129,6 +135,7 @@ export const CompositionBlock = ({
               entityStore={entityStore}
               breakpoints={breakpoints}
               resolveDesignValue={resolveDesignValue}
+              usedComponents={usedComponents}
             />
           );
         })
--- a/packages/experience-builder-sdk/src/blocks/DesignComponent.tsx
+++ b/packages/experience-builder-sdk/src/blocks/DesignComponent.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+// Feel free to do any magic as regards variable definitions for design components
+// Or if this isn't necessary by the time we figure that part out, we can bid this part farewell
+export const DesignComponent = ({ ...props }) => {
+  return <div {...props} />;
+};
--- a/packages/experience-builder-sdk/src/blocks/DesignComponent.tsx
+++ b/packages/experience-builder-sdk/src/blocks/DesignComponent.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 
+const designComponentStyle = { display: 'contents' };
+
 // Feel free to do any magic as regards variable definitions for design components
 // Or if this isn't necessary by the time we figure that part out, we can bid this part farewell
 export const DesignComponent = ({ ...props }) => {
   // Using a display contents so design component content/children
   // can appear as if they are direct children of the div wrapper's parent
-  return <div {...props} style={{ display: 'contents' }} />;
+  return <div {...props} style={designComponentStyle} />;
 };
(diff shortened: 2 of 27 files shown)


SofiaMargariti:
Still reviewing the code, but there is an issue that I found you can see below:
* I create a component
* before I save or reload the composition I drag and drop another normal component into the canvas
* The new component I just created disappears and shows as missing in the layers tab

https://github.com/contentful/experience-builder/assets/2773012/1e540b52-febc-422c-bfa1-4f5e6e8e5dea

If I do the same thing but before dropping the component I save and reload the page it works as expected.

ChidinmaOrajiaku:
> Still reviewing the code, but there is an issue that I found you can see below:
> 
> * I create a component
> * before I save or reload the composition I drag and drop another normal component into the canvas
> * The new component I just created disappears and shows as missing in the layers tab
> 
>  Screen.Recording.2023-11-07.at.08.50.51.mov 
> If I do the same thing but before dropping the component I save and reload the page it works as expected.

Hmm weird. Can't reproduce this

https://github.com/contentful/experience-builder/assets/30434146/debb9d46-d8da-46cd-a0b2-136cb80f9cdb

packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
```diff
@@ -187,6 +201,33 @@ export function VisualEditorContextProvider({
           }
           break;
         }
+        case INCOMING_EVENTS.DesignComponentsUpdated: {
```
SofiaMargariti, line 204: This is more of `DesignComponentAdded` event, right? It handles only a single component and it seems it's only applicable when a new component is added to the composition. The sys object and component definition won't change after a component has been created/added so what does 'updated' mean here?
Also is it possible to have this event only update the registry so there is some separation of responsibilities each event has a single responsibility. The tree will be updated by the `INCOMING_EVENTS.CompositionUpdated` anyway right?

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      dataSource[uuid] = { ...compositionDataSource[uuid] };
+    } else if (variable.type === 'UnboundValue') {
+      const uuid = variable.key;
+      unboundValues[uuid] = compositionUnboundValues[uuid];
+    }
+  }
+
+  const isDesignComponent = node.definitionId?.startsWith('DesignComponent');
+  const nodeId = isDesignComponent ? node.definitionId?.split('-')[1] : generateRandomId(16);
```
SofiaMargariti, line 42: Why do we need this random generated ID? Don't all nodes have a nodeId already?

packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
ChidinmaOrajiaku: I don't call the CompositionUpdated event on design component created, only this event. I did this because `CompositionUpdated` also sets datasource, design, values and the rest of them, so I wanted there to be a split. So whenever the tree is updated as a result of design component created and added to the tree, I send the tree and other other payload in this event. 

Also, good point about the naming. Will change

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
ChidinmaOrajiaku: Because as at the time we fetch design component entry's children in the SDK, they don't have a node ID and need to be deserialized. Hence we generate a node id for them during deserialization (just like we do during node deserialization in the web app)

packages/experience-builder-sdk/src/utils.ts
```diff
@@ -112,3 +115,23 @@ export const getInsertionData = ({
...
+
+export const generateRandomId = (count: number): string => {
+  const LETTERS = 'abcdefghijklmnopqvwxyzABCDEFGHIJKLMNOPQVWXYZ';
+  const NUMS = '0123456789';
+  const ALNUM = NUMS + LETTERS;
+
+  return times(count, () => ALNUM[random(0, ALNUM.length - 1)]).join('');
+};
+
+export const checkIfDesignComponent = ({
```
Chaoste, line 127: Reads to me like a typo when thinking of other helper methods like `checkIsExperience` or was this intentional?
```suggestion
export const checkIsDesignComponent = ({
```

packages/experience-builder-sdk/src/utils.ts
```diff
@@ -112,3 +115,23 @@ export const getInsertionData = ({
     };
   }
 };
+
+export const generateRandomId = (count: number): string => {
```
Chaoste, line 119: [nitpick] a more explicit variable name could be `letterCount` or `length` IMHO

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  const deserializedNode = deserializeDesignComponentNode({
+    node: {
+      definitionId: node.data.blockId || '',
+      variables: {},
+      children: componentFields.componentTree.children,
+    },
+    nodeId: node.data.id,
+    parentId: node.parentId,
+    experienceDataSource: {},
+    experiencenUnboundValues: componentFields.unboundValues,
```
Chaoste, line 106:
```suggestion
    experienceUnboundValues: componentFields.unboundValues,
```

packages/experience-builder-sdk/src/blocks/VisualEditorRoot.tsx
```diff
@@ -55,8 +55,11 @@ const VisualEditorRootComponents = () => {
   useEffect(() => {
     const resolveEntities = async () => {
       setEntitiesFetched(false);
-      const entityLinks = Object.values(dataSource || {});
-      await entityStore.current.fetchEntities(entityLinks);
+      const datasourceEntityLinks = Object.values(dataSource || {});
```
Chaoste, line 58:
```suggestion
      const dataSourceEntityLinks = Object.values(dataSource || {});
```

packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
```diff
@@ -187,6 +201,33 @@ export function VisualEditorContextProvider({
...
+          const {
+            tree,
+            designComponent,
+            designComponentDefinition,
+          }: {
+            tree: CompositionTree;
+            designComponent: Entry;
+            designComponentDefinition: ComponentRegistration['definition'];
+          } = payload;
+          if (designComponent) {
```
Chaoste, line 214: Is there a case where `designComponent` can be undefined? If yes, I would adjust the type of the variables to have type safety. If no, we can remove this guard clause :)

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
```diff
@@ -40,18 +45,37 @@ type VisualEditorBlockProps = {
   areEntitiesFetched: boolean;
 };
 
+const designComponentTypes = ['DesignComponent', 'DesignComponentBlock'];
```
Chaoste: - Constants should be in SCREAMING_SNAKE_CASE
- I would like to add `node` as `type` can mean a lot (e.g. content type, experience type)

```suggestion
const DESIGN_COMPONENT_NODE_TYPES = ['DesignComponent', 'DesignComponentBlock'];
```

packages/experience-builder-sdk/src/blocks/DesignComponent.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+// Feel free to do any magic as regards variable definitions for design components
+// Or if this isn't necessary by the time we figure that part out, we can bid this part farewell
+export const DesignComponent = ({ ...props }) => {
+  // Using a display contents so design component content/children
+  // can appear as if they are direct children of the div wrapper's parent
+  return <div {...props} style={{ display: 'contents' }} />;
```
Chaoste, line 8: According to [caniuse.com](https://caniuse.com/?search=display%20contents), this display value is not yet stable and might have some issues, e.g. with buttons.

Is the original container that was removed to a design component, a children of the design component in here?

In that case, I would suggest to directly render this container as before and not render a div-wrapper around it.

Ofc, this means we need to assume that the componentTree of a design component can only have one direct child on the first layer and this child must be a container. Every other child on top level won't be rendered (and we would have to make sure that there is only one node on root-level when updating a design component)

packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
```diff
@@ -28,22 +35,45 @@ type CompositionBlockProps = {
...
+    return isDesignComponent
+      ? resolveDesignComponent({
+          node: rawNode,
+          entityStore,
+        })
+      : rawNode;
+  }, [entityStore, isDesignComponent, rawNode]);
+
+  const componentRegistration = useMemo(() => {
+    const registeration = getComponentRegistration(node.definitionId as string);
```
Chaoste, line 66:
```suggestion
    const registration = getComponentRegistration(node.definitionId as string);
```

Chaoste (changes requested):

packages/experience-builder-sdk/src/utils.ts
ChidinmaOrajiaku: It's intentional actual. It reads to me "check IF this component is a design component" hence `checkIfDesignComponent`

packages/experience-builder-sdk/src/blocks/VisualEditorContext.tsx
ChidinmaOrajiaku: I added it just in case someone calls the event and doesn't send the designComponent (for some reason). If that case occurs, then there's no need updating an undefined entity and all other operations there

packages/experience-builder-sdk/src/core/componentRegistry.ts
```diff
@@ -129,3 +129,35 @@ export const getComponentRegistration = (id: string) => {
...
+  component,
+}: {
+  definitionId: string;
+  defintionName?: string;
+  component: ComponentRegistration['component'];
+}) => {
+  const componentRegistration = componentRegistry.get(definitionId);
+
+  if (componentRegistration) {
+    return componentRegistration;
```
Spring3, line 149: would it make sense to align the response shape to what is on line 162?

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+      unboundValues[uuid] = experienceUnboundValues[uuid];
+    }
+  }
+
+  const isDesignComponent = designComponentsRegistry.has(node.definitionId);
+
+  const children: CompositionComponentNode[] = node.children.map((child) =>
+    deserializeDesignComponentNode({
+      node: child,
+      nodeId: generateRandomId(16),
```
Spring3, line 48: will it be different each time? Should we try to use the existing id if it is already there?

Spring3 (changes requested):
I have tested it locally and it worked quite well. I haven't yet finished reviewing this PR (and am yet to check the web app side), but I already spotted a couple of things that might be worth addressing.

I will continue to review this PR tomorrow. Well done so far!

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
```diff
@@ -40,18 +45,37 @@ type VisualEditorBlockProps = {
   areEntitiesFetched: boolean;
 };
 
+const DESIGN_COMPONENT_NODE_TYPES = ['designComponent', 'designComponentBlock'];
```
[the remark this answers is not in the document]
Chaoste: I think the types package would be good so we can use it in the web app as well

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
```diff
@@ -40,18 +45,37 @@ type VisualEditorBlockProps = {
...
+  const componentRegistration = useMemo(() => {
+    const registration = getComponentRegistration(node.data.blockId as string);
+    if (node.type === 'designComponent' && !registration) {
+      return createDesignComponentRegistration({
+        definitionId: node.data.blockId as string,
+        component: DesignComponent,
+      });
+    }
+    return registration;
+  }, [node]);
```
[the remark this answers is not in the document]
Chaoste: Theoretically, this would have no effect on the performance of this component if this is your concern. tbh I kinda like that one memo returns one variable and thus split up as we need both variable below.

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
```diff
@@ -140,6 +194,12 @@ export const VisualEditorBlock = ({
   const children =
     definition.children === true
       ? node.children.map((childNode) => {
+          // if parent is a design component, only render children belonging to the design component
+          if (
+            DESIGN_COMPONENT_NODE_TYPES.includes(node.type) &&
+            !DESIGN_COMPONENT_NODE_TYPES.includes(childNode.type)
+          )
+            return null;
```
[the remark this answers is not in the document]
Chaoste: Daniel W referred to them today as "slots", i.e. places where one could put children inside a design component instance. In the near future, we won't tackle this. Maybe, we have time left before GA but hard to tell right now.

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
ChidinmaOrajiaku: There's no existing node Id when we fetch the design component's children, which is why I generate a node id on deserialisation (similar to what we do in the web app)

packages/experience-builder-sdk/src/core/componentRegistry.ts
ChidinmaOrajiaku: Align the response shape in what way? This check is for incase we have the definition of that component saved already, then it returns it instead of re-setting it. The shape is already aligned

packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
```diff
@@ -28,22 +35,45 @@ type CompositionBlockProps = {
...
+    const registration = getComponentRegistration(node.definitionId as string);
+
+    if (isDesignComponent && !registration) {
+      return createDesignComponentRegistration({
+        definitionId: node.definitionId as string,
+        component: DesignComponent,
+      });
+    }
+    return registration;
+  }, [isDesignComponent, node.definitionId]);
```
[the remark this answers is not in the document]
ChidinmaOrajiaku: I would say we shouldn't. I kinda like how the useMemo's have separate concerns and different dependencies.

packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
```diff
@@ -2,10 +2,13 @@ import React, { useMemo } from 'react';
 
 import type { UnresolvedLink } from 'contentful';
 import omit from 'lodash.omit';
-import { EntityStore } from '@contentful/visual-sdk';
+import { EntityStore } from '../core/EntityStore';
```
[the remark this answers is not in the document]
ChidinmaOrajiaku: The Composition block is for preview/delivery mode and uses the core EntityStore

The Visual editor block uses EditorModeEntityStore.

So it should be fine to use that core Entity store type with its methods for the preview/delivery mode block.

packages/experience-builder-sdk/src/blocks/CompositionBlock.tsx
```diff
@@ -66,7 +96,7 @@ export const CompositionBlock = ({
         }
         case 'UnboundValue': {
           const uuid = variable.key;
-          acc[variableName] = unboundValues[uuid]?.value;
+          acc[variableName] = (entityStore?.unboundValues || unboundValues)[uuid]?.value;
```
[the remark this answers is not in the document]
ChidinmaOrajiaku: But Daniel, you wrote the code that stores the unbound values in the preview/delivery mode entity store [here](https://github.com/contentful/experience-builder/blob/feat/render-design-component-within-an-experience/packages/experience-builder-sdk/src/core/EntityStore.ts#L46) 😂.

The unbound values props passed into this component comes from the store [here](https://github.com/contentful/experience-builder/blob/79031a2ddf46adefad40055852f7e21d5e58fd81/packages/experience-builder-sdk/src/blocks/PreviewDeliveryRoot.tsx#L51)

I also don't see any problems with storing them in the entity store. I'm using it directly here because I added an update unbound values method [here](https://github.com/contentful/experience-builder/pull/151/files/cc74cd52fc4452978b94faa6f64e6264f3249381) that I use to update the unbound values object when deserializing design component nodes [here](https://github.com/contentful/experience-builder/pull/151/files/cc74cd52fc4452978b94faa6f64e6264f3249381#diff-966851c3ad5b7a4330fcc1f676a4c2a9583fab1a6d57266c0ddd50b9a6acb3bbR38)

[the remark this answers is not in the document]
ChidinmaOrajiaku: Also `CompositionBlock` is not used in editor mode, it's used only for preview/delivery mode. We use the `VisualEditorBlock` in editor mode, so the stores are correctly different and supply different things.

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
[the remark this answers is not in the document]
ChidinmaOrajiaku: +1. As I said above, I like how each useMemo returns one variable and have separate concerns.

packages/experience-builder-sdk/src/blocks/VisualEditorBlock.tsx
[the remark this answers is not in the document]
ChidinmaOrajiaku: Oki 👍

Chaoste (approved):
Just functionally tested everything carefully and it all works very well!

packages/experience-builder-sdk/src/core/componentRegistry.ts
Spring3: On line 62 we return both - a component and a component Definition.

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
Spring3: I thought that each node had a `data.id` in it (which is already a randomly generated string), so I meant that id

packages/experience-builder-sdk/src/core/componentRegistry.ts
ChidinmaOrajiaku: If you check what's stored in the component registry, it's the component and its definition. So 162 and this line have the same response shape

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
ChidinmaOrajiaku: The design component block nodes deserialized here in the SDK. The reason we have a node id normally is because the nodes are deserialized and sent from the web app. It's different for design components.

For design components, we get a design component id from the tree, fetch that entry, and deserialize the nodes here in the SDK to render. During the deserialization process here is when I'm generating a node id.

Spring3 (approved):
Sorry for blocking this PR for so long unintentionally. I couldn't find time to check on it due to other things that required attention

PR #155, opened 2023-11-07
fix(experience-builder-sdk): update rollup config to not include react in build []

elylucas:
Currently, React is included in the production build, which could cause issues and increases the size of the built bundle. This update adds the rollup-plugin-peer-deps-external package and updates the Vite config to use it. This will exclude any external libraries marked as external peer deps. This drops the bundle size from 167KB to 89KB.


primeinteger (approved):

Spring3 (approved):
Thanks! good to learn about this

chasepoirier (approved):

PR #156, opened 2023-11-07
fix(experience-builder-sdk): export ExternalSDKMode type []

elylucas:
Exports the ExternalSDKMode from the main types entry point. This appeared to be accidentally missing.

### DIFF
--- a/packages/experience-builder-sdk/src/index.ts
+++ b/packages/experience-builder-sdk/src/index.ts
@@ -5,6 +5,7 @@ export { calculateNodeDefaultHeight } from './core/stylesUtils';
 export type {
   ComponentDefinition,
   ComponentRegistration,
+  EntityStore,
   Experience,
   ExternalSDKMode,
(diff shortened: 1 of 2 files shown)


Chaoste (approved):

packages/experience-builder-sdk/src/index.ts
```diff
@@ -6,6 +6,7 @@ export type {
   ComponentDefinition,
   ComponentRegistration,
   Experience,
+  ExternalSDKMode,
```
Spring3, line 9: we might need to also export `EntityStore` type that is taken from the `@contentful/visual-sdk`

Spring3 (approved):

chasepoirier (approved):

packages/experience-builder-sdk/src/index.ts
elylucas: added this one as well

PR #161, opened 2023-11-15
feat(experience-builder-sdk)!: make returned experience reactive to props []

elylucas:
## Updates to useFetchExperience reactivity

This PR updates the useFetchExperience hook to make the experience returned from the hook reactive based on the props that are passed into the hook. This will eliminate the need for the user to have to manually call the fetchBySlug method manually, which will simplify the dev setup when configuring EB.

The call to fetch the experience can now be simplified to this inside of a component:

```tsx
const { experience, error } = useFetchExperience({
  client,
  mode,
  experienceTypeId,
  localeCode,
  slug,
});

if (error) {
  return <div>{error.message}</div>;
}

if (!experience) {
  return null;
}

return <ExperienceRoot experience={experience as any} locale={localeCode} />;
```

If the `slug`, `experienceTypeId`, or `localeCode` change, the new experience will automatically be fetched.

## Updates to error handling in useFetchExperience

This PR also removes the rethrowing of error of exceptions from within the hook. React can't catch errors raised in hooks, and the recommended practice is to return a reactive error object from the hook that when present, will signify an error occurred in the hook.

The tests for `useFetchExperience` were refactored to allow for this error handling change.

## Breaking Changes

There is a breaking change in the API for the hook, as it now needs all the additional params that were passed into `fetchBySlug` passed into the call to the hook itself so it can call `fetchBySlug`/`fetchById` internally. The new params to pass in are:

```
slug?: string;
id?: string;
experienceTypeId: string;
localeCode: string;
```

Here, either slug or id is required. An error will be returned if neither are passed in.

If we don't want this to be a breaking change, we can probably get it to work making all these params optional, but the user will still need to call the fetch methods manually in that case.

## Discussion topics

- [ ] Do `fetchBySlug` and `fetchById` still need to be exposed by the hook? This PR does not remove them, but if they are not needed we should not returned them anymore as it adds to the API.


packages/experience-builder-sdk/src/hooks/useFetchExperience.test.ts
```diff
@@ -26,51 +27,44 @@ describe('useFetchExperience', () => {
...
+    const { result } = await act(async () => {
+      return renderHook(useFetchExperience, {
+        initialProps: {
+          client: clientMock,
+          mode: 'preview' as ExternalSDKMode,
+          slug,
+          experienceTypeId,
+          localeCode,
+        },
+      });
```
Chaoste, line 40: Wrapping `render` statements inside `act` is considered a bad practice afaik. I can recommend reading [this](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily) and the linked blog post.
TLDR; In the first render cycle, it fetches the experience and thus has an async behaviour. React notices this which could harm the integrity of our unit tests and thus probably printed a warning in your case. To address this, the test setup should make sure that all async this settled before running any further action or assertion. For that you can usually use something like `waitFor` that will wait for something to change in the returned object that signals you that it's not fetching anymore.

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
```diff
@@ -15,14 +15,26 @@ const handleError = (generalMessage: string, error: unknown) => {
...
+  mode,
+  client,
+  slug,
+  id,
+  experienceTypeId,
+  localeCode,
+}: useClientSideExperienceFetchersProps) => {
   const [experience, setExperience] = useState<Experience<EntityStore> | undefined>(undefined);
   const [isFetching, setIsFetching] = useState(false);
+  const [error, setError] = useState<Error>();
```
Chaoste, line 37: That's a great improvement! Thanks for adding this :)

Chaoste (commented):
Thanks for this improvement!
As you were wondering about dropping `fetchBySlug` etc. in the PR description: We intentionally wrote those imperative methods as the SDK should behave as framework-agnostic as possible. At some point, we will have to revisit all the logic to make it work without React for our customers. To prepare for that, we expose plain functions like these fetch methods. So I wouldn't drop them and probably, we should keep a few tests that ensure calling them directly works as well.

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
```diff
@@ -144,21 +159,32 @@ export const useFetchExperience = ({ mode, client }: useClientsideExperienceFetc
...
   );
 
+  useEffect(() => {
+    if (slug) {
+      fetchBySlug({ slug, localeCode, experienceTypeId });
+    } else if (id) {
+      fetchById({ id, localeCode, experienceTypeId });
+    } else {
+      setError(Error('Either slug or id must be provided to useFetchExperience'));
+    }
```
chasepoirier, line 183: Small code pattern opinion; I find early returns to be a more readable pattern compared to else ifs. It doesn't matter too much here since the benefits are mainly for nested ifs but I do like to try using it for all conditionals

Something like:
```ts
useEffect(() => {
  if (slug) {
    fetchBySlug({ slug, localeCode, experienceTypeId });
    return;
  }

  if(id) {
    fetchById({ id, localeCode, experienceTypeId });
    return;
  }

  setError(Error('Either slug or id must be provided to useFetchExperience'));
[]);
```

packages/experience-builder-sdk/src/hooks/useFetchExperience.test.ts
elylucas: thanks! TIL. Updating to wrap the  expects in a waitFor

Chaoste (approved):
Thanks for updating the tests!

If you have a few minutes, I can recommend reading the [complete blog post about "common mistakes with RTL"](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) :)

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
```diff
@@ -144,21 +159,34 @@ export const useFetchExperience = ({ mode, client }: useClientsideExperienceFetc
...
     },
     [client, mode]
   );
 
+  useEffect(() => {
+    if (slug) {
+      fetchBySlug({ slug, localeCode, experienceTypeId });
+      return;
+    }
+    if (id) {
```
DanweDE, line 181: Nit: Wonder if the order matters, what if both slug and ID are provided for some reason, wouldn't we prefer fetching by ID?

DanweDE (approved):

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
```diff
@@ -144,21 +159,34 @@ export const useFetchExperience = ({ mode, client }: useClientsideExperienceFetc
...
+      fetchById({ id, localeCode, experienceTypeId });
+      return;
+    }
+    setError(Error('Either slug or id must be provided to useFetchExperience'));
+  }, [experienceTypeId, fetchById, fetchBySlug, id, localeCode, slug]);
+
   return {
+    error,
     fetchBySlug,
     fetchById,
```
Spring3, line 191: Aren't we now mixing 2 approaches together? If we commit to returning the `error` from the hook, then I would suggest instead returning `refetch` as a method, similar to `react-query` instead of the methods we currently return from the hook.

Separate functions are available via `import { fetchers } from '@contentful/experience-builder'` if needed.

Spring3 (commented):
Shared my thoughts below.

One point to take into account is also that we had some more plans [before releasing a major version](https://contentful.atlassian.net/browse/SPA-1634)

If this PR gets merged, please don't forget to update the docs [on wiki](https://github.com/contentful/experience-builder/wiki) and our [doc-app](https://github.com/contentful/doc-app/blob/master/content/pages/developers/docs/tutorials/general/experience-builder.md)

PR #164, opened 2023-11-16
fix: allow ExperienceRoot to render in editor mode when a new Experience entry is first created [ALT-106]

primeinteger:
This changes the `ExperienceRoot` component to allow the `experience` prop to be optional.  
* This helps to display the Visual Editor in editor mode when an Experience entry is first created in Contentful and before it has been saved to the database

Background context
* When I updated the EB Toolkit to the latest EB SDK in contentful/experience-builder-toolkit#40 - I noticed when creating a new Experience entry that the `findBySlug()` method fails to fetch the new page (it hasn't been saved to the database yet) and ends up rendering a blank canvas and no components in the sidebar from [this early return](https://github.com/contentful/experience-builder-toolkit/blob/50b0a6f938b4f00263f7f86698204158428f5ceb/packages/test-app/src/Page.tsx#L44-L46)

Updated `contentful` dep
* I updated the `contentful` client to the latest version in this PR because the `ContentfulClientApi` type was mismatching the client in the toolkit when using different versions.  By updating the SDK and toolkit both to a newer version of this package, the type mismatch is resolved.


packages/experience-builder-sdk/src/ExperienceRoot.tsx
```diff
@@ -21,6 +21,13 @@ type ExperienceRootProps = {
 
 export const ExperienceRoot = ({ locale, experience, slug }: ExperienceRootProps) => {
   const [mode, setMode] = useState<InternalSDKMode>(() => {
+    if (!experience) {
+      if (window !== window.parent) {
```
Chaoste, line 25: If I'm not mistaken, this check could also render in an SSR/ SSG context where `window` is not defined. We use this safeguard check (almost) everytime in the SDK when we need to access `window`.
```suggestion
      if (typeof window !== 'undefined' && window !== window.parent) {
```

package-lock.json
Chaoste, line 1: We're using `yarn` in this repository. To avoid having two lock files, I suggest dropping this one and sticking to `yarn.lock` :)

packages/experience-builder-sdk/src/ExperienceRoot.tsx
primeinteger: Good insight, thank you for the suggestion

Chaoste (approved):

PR #167, opened 2023-11-21
feat: handle bindings for design components [SPA-1606]

anwaar931:
- Handling unbound values bindings for design components


packages/experience-builder-sdk/src/core/preview/designComponentUtils.ts
```diff
@@ -28,6 +28,9 @@ export const resolveDesignComponent = ({
   }
 
   const componentFields = (designComponent as ExperienceEntry).fields as unknown as Composition;
+  if (!componentFields) {
+    return node;
+  }
```
Chaoste, line 33: We talked about this in the pairing session. IIRC `fields` would only be empty if it was a `Link` which will happen on CMA but not on CPA/ CDA.
The type of `usedComponents` is already correct in that regard: `Array<Link<'Entry'> | ExperienceEntry>`

Instead of casting back and forth, I would rewrite it this way so TS can infer which of both scenario we're in:
```suggestion
  if (!('fields' in designComponent)) {
    return node;
  }
  const componentFields = designComponent.fields;
```

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -33,10 +37,18 @@ export const deserializeDesignComponentNode = ({
     if (variable.type === 'BoundValue') {
       const [, uuid, ,] = variable.path.split('/');
 
-      dataSource[uuid] = { ...experienceDataSource[uuid] };
+      dataSource[uuid] = { ...designComponentDataSource[uuid] };
     } else if (variable.type === 'UnboundValue') {
```
ChidinmaOrajiaku, line 41: Is there a scenario where the variable type would be Bound value or unbound value? I thought we'd only have `ComponentValue` and `DesignValue` as variables types, and on looking up the `variableMapping`, we know if it's a bound or unbound value

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -102,8 +116,10 @@ export const resolveDesignComponent = ({
     },
     nodeId: node.data.id,
     parentId: node.parentId,
-    experienceDataSource: {},
-    experienceUnboundValues: componentFields.unboundValues,
+    designComponentDataSource: {},
+    designComponentUnboundValues: componentFields.unboundValues,
+    componentInstanceProps: node.data.props,
```
ChidinmaOrajiaku, line 121: Here we mean the top-level container props yeah?

Chaoste (approved):
Nice improvements! Just a small nitpick :)

ChidinmaOrajiaku (changes requested):
We need to omit these props from being passed when creating the component [here](https://github.com/contentful/experience-builder/blob/5cbb0bae2878f434ea9b6ec43f2d18fccc620d56/packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx#L267)

Screenshot 2023-11-22 at 11 22 00

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
anwaar931: Yep root level design component props

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
anwaar931: There shouldn't be a scenario, I can remove this old code

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
```diff
@@ -15,14 +15,18 @@ export const deserializeDesignComponentNode = ({
...
+  componentInstanceProps,
+  componentInstanceUnboundValues,
 }: {
   node: CompositionNode;
   nodeId: string;
   parentId?: string;
-  experienceDataSource: CompositionDataSource;
-  experienceUnboundValues: CompositionUnboundValues;
+  designComponentDataSource: CompositionDataSource;
+  designComponentUnboundValues: CompositionUnboundValues;
```
ChidinmaOrajiaku, line 27: Will we ever have a need for the `designComponentDataSource` and `designComponentUnboundValues`? 

Because looking at line 41 -51, we make use of the componentInstanceUnboundValues and we'd also probably do the same when we start dealing with the data source, so we never really take a look at the unbound values object in the design component.

Except we use it as a fallback value for when there `componentInstanceUnboundValues` is empty or does not have the value

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
Chaoste: Maybe later on, when we allow to fix content variables within the design component, they might be direct unbound values / data source instead of being configurable through a ComponentValue but that's more an idea that we might implement or not.

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
Chaoste: They contain the default value so we could use it as a fallback but theoretically, we copy over everything and thus don't need to look them up here again, right?

packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx
```diff
@@ -130,6 +130,18 @@ export const VisualEditorBlock = ({
             ...acc,
             [variableName]: transformContentValue(value, variableDefinition),
           };
+        } else if (variableMapping.type === 'ComponentValue') {
+          // For design component, we are only handling binding for UnboundValues for now
+          const value = getUnboundValues({
+            key: variableMapping.key,
+            fallback: variableDefinition.defaultValue,
```
ChidinmaOrajiaku, line 137: Aha! This is probably where the [object Object] I talked about [here](https://github.com/contentful/user_interface/pull/18355#pullrequestreview-1743970513) is coming from, coz the default value is an object of type and key/path.

Would be good if we grab the default value from the design component unbound values (coz that's where the actual default values are stored yeah?).

packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx
```diff
@@ -130,6 +130,18 @@ export const VisualEditorBlock = ({
             ...acc,
             [variableName]: transformContentValue(value, variableDefinition),
           };
+        } else if (variableMapping.type === 'ComponentValue') {
```
ChidinmaOrajiaku, line 133: We might not need this check `if (variableMapping.type === 'ComponentValue')` if we return the default value if there's nothing in the `componentInstanceUnboundValues` object [here](https://github.com/contentful/experience-builder/blob/76713e64584d10f7fe6ef7eef7c3cbeb951dee53/packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts#L45).

Because what's happening in this check is almost the same as what's happening on line 146. If when we deserialize the nodes, we insert a value or default value in the node unbound value object, we probably won't need the extra check here.

packages/experience-builder-sdk/src/core/editor/designComponentUtils.ts
ChidinmaOrajiaku: I think that should be the case - using it as a fallback. But there's a miscommunication somewhere between the `designComponentUnboundValues` and the variable mapping, coz the uuids are totally different (which shouldn't be). 

Look at the `unboundValues` in the instance and the design component unboundValues in this screenshot. The uuids don't match at all, so we can't do a look up there. 

Screenshot 2023-11-22 at 11 31 00

anwaar931: We cannot copy same `uuid`s of `UnboundValues` from design component to its instance. In a scenario where customer is using an instance of design component multiple times in Experience.  `uuid`s should be different then.

Chaoste: I think it matches right now the developer components, doesn't it? If you would change there a default values it would overwrite existing instances of those as the default values are copied over, correct?

SofiaMargariti (approved):

PR #168, opened 2023-11-21
feat: [SPA-1574] add media type and handle it

loweisz:
This will add the `Media` type and also handle the asset being tagged in the `path` of the bound prop. 

more info in this PR: https://github.com/contentful/user_interface/pull/18433


packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx
```diff
@@ -121,9 +121,26 @@ export const VisualEditorBlock = ({
           const [, uuid, ...path] = variableMapping.path.split('/');
           const binding = dataSource[uuid] as Link<'Entry' | 'Asset'>;
 
-          const boundValue = areEntitiesFetched
+          let boundValue: string | Link<'Asset'> | undefined = areEntitiesFetched
             ? entityStore.current?.getValue(binding, path.slice(0, -1))
             : undefined;
+
+          // If there is no boundValue found, we check with the second last item removed as well
+          // this might be the case for the additional linked asset
```
Chaoste, line 129: "we check with the second last item removed as well"
I don't really get what this means, sorry ^^

Chaoste (changes requested):

Chaoste (approved):
Thanks!

packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx
```diff
@@ -121,9 +121,27 @@ export const VisualEditorBlock = ({
...
+          // In some cases, there may be an asset linked in the path, so we need to consider this scenario:
+          // If no 'boundValue' is found, we also attempt to extract the value associated with the second-to-last item in the path.
+          // If successful, it means we have identified the linked asset.
+
+          if (!boundValue) {
+            boundValue = areEntitiesFetched
+              ? (entityStore.current?.getValue(
+                  binding,
+                  path.slice(0, -2)
+                ) as unknown as Link<'Asset'>)
```
Spring3, line 137: to make it a bit more robust, would it make sense to add a typeguard that would perform this check instead of confidently casting the returned object to `Link<Asset>`?

SofiaMargariti (approved):

packages/experience-builder-sdk/src/blocks/editor/VisualEditorBlock.tsx
loweisz: The Problem is that the type returned from the `getValue` function of the entityStore comes from  `@contentful/visual-sdk`, which is `string` or `undefined`. 

How would do that without making a change in that package?

PR #171, opened 2023-11-29
[ALT-148] Allows user defined styling layout through Component Definition injection at the root level

ryunsong-contentful:
This PR allows the user to define their layoutDefinition for the layout sidebar as follows below.
Here's the user_interface associated PR that shows the UI and the mechanics of how it would work https://github.com/contentful/user_interface/pull/18533

```
export const buttonDefinition: ComponentDefinition = {
  id: 'ryuns-button',
  name: 'ryuns-button',
  category: 'buttons',
  **designTokensDefinition: {
    spacing: { XS: '0px', S: '16px', M: '32px', L: '64px' },
  },**
  // thumbnailUrl: string | undefined;
  variables: {
    buttonTitle: {
      type: 'Text',
      displayName: 'Button Title',
      defaultValue: 'Click Me'
    },
    marginValues: {
      type: 'Text',
      displayName: 'Margin Values',
      defaultValue: ['0px', '16px', '32px', '64px']
    }
  },
  builtInStyles: ["cfHorizontalAlignment", "cfVerticalAlignment", "cfMargin", "cfPadding", "cfBackgroundImageAlignment",
    "cfBackgroundColor", "cfWidth", "cfHeight", 'cfMaxWidth', 'cfFlexDirection', 'cfFlexWrap', 'cfBorder',
    'cfGap', 'cfBackgroundImageUrl', 'cfBackgroundImageScaling', 'cfBackgroundImageAlignment'
  ]
  // children?: boolean | undefined,
}
```

### DIFF
--- a/packages/experience-builder-sdk/src/core/designTokenRegistry.ts
+++ b/packages/experience-builder-sdk/src/core/designTokenRegistry.ts
@@ -1,0 +1,14 @@
+import { OUTGOING_EVENTS } from '../constants';
+import { sendMessage } from '../communication/sendMessage';
+import { DesignTokensDefinition } from '../types';
+
+/**
+ * Register design tokens styling
+ * @param designTokenDefinition - {[key:string]: Record<string, string>}
+ * @returns void
+ */
+export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition) => {
+  sendMessage(OUTGOING_EVENTS.DesignTokens, {
+    designTokens: designTokenDefinition,
+  });
+};
--- a/packages/experience-builder-types/src/constants.ts
+++ b/packages/experience-builder-types/src/constants.ts
@@ -12,6 +12,7 @@ export const OUTGOING_EVENTS = {
   NewHoveredElement: 'newHoveredElement',
   ComponentSelected: 'componentSelected',
   RegisteredComponents: 'registeredComponents',
+  DesignTokens: 'registerDesignTokens',
   RequestComponentTreeUpdate: 'requestComponentTreeUpdate',
   ComponentDropped: 'componentDropped',
   CanvasReload: 'canvasReload',
--- a/packages/experience-builder-types/src/types.ts
+++ b/packages/experience-builder-types/src/types.ts
@@ -254,6 +254,8 @@ export type Composition = {
   componentSettings?: ExperienceComponentSettings;
 };
 
+export type DesignTokensDefinition = { [key: string]: Record<string, string> };
+
 export type ExperienceEntry = {
   sys: Entry['sys'];
   fields: Composition;
--- a/packages/experience-builder-sdk/src/index.ts
+++ b/packages/experience-builder-sdk/src/index.ts
@@ -1,6 +1,7 @@
 export { ExperienceRoot } from './ExperienceRoot';
 export { useExperienceBuilder, getValueForBreakpoint, useFetchExperience } from './hooks';
 export { defineComponents } from './core/componentRegistry';
+export { defineDesignTokens } from './core/designTokenRegistry';
 export { calculateNodeDefaultHeight } from './utils/stylesUtils';
 export { checkIfDesignComponent } from './utils/utils';
 export type {


packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -158,11 +158,6 @@ export function VisualEditorContextProvider({
         return;
       }
 
-      console.debug(
```
ryunsong-contentful, line 161: Are these debugs still useful to have?

packages/experience-builder-types/src/types.ts
```diff
@@ -243,6 +243,7 @@ export type ExperienceComponentSettings = {
 export type Composition = {
   title: string;
   slug: string;
+  tokenDefinition?: DesignTokensDefinition;
```
SofiaMargariti, line 246: This is type for the fields of the Experience Entry (previously called composition). `tokenDefinitions` won't be a separate field, i.e. defined in the Content Type, right? So I don't think we need it here

packages/experience-builder-sdk/src/core/designTokenRegistry.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Register design tokens styling
+ * @param designTokenDefinition - {[key:string]: Record<string, string>}
+ * @returns void
+ */
+export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition) => {
+  sendMessage(OUTGOING_EVENTS.DesignTokens, {
+    designTokens: designTokenDefinition,
+  });
+};
```
SofiaMargariti, line 14: You will need to save the tokens into an object so that you can use them during rendering to resolve the actual values, similar to what we do for registered components: https://github.com/contentful/experience-builder/blob/f67f2f9b6ded47f860ee3404749fff88f96d8415/packages/experience-builder-sdk/src/core/componentRegistry.ts#L30.
I am not sure if this is in the scope of this PR though, feel free to disregard if it's gonna be done in a followup.

packages/experience-builder-types/src/constants.ts
```diff
@@ -12,6 +12,7 @@ export const OUTGOING_EVENTS = {
   NewHoveredElement: 'newHoveredElement',
   ComponentSelected: 'componentSelected',
   RegisteredComponents: 'registeredComponents',
+  DesignTokens: 'registeredDesignTokens',
```
SofiaMargariti, line 15: This should probably indicate the action so `RegisterDesignTokens` or `DesignTokensRegistered` might be better. (Unfortunately we have not agreed on a specific pattern for the rest of the events so they are inconsistent with each other.)

packages/experience-builder-sdk/src/core/designTokenRegistry.ts
ryunsong-contentful: Can you explain this part? So my understanding is that we need to save them on the user_interface side (which I did by using the immer building block store). This way, whenever we render the layout sidebar that uses the token key, we can pull the object from the store and then resolve the value at runtime.

SofiaMargariti: We'd need to resolve values both in the sidebar in user_interface and in the SDK side when rendering a component that uses a token. User interface could pass back to the SDK the actual value but that would work only for the editor mode. In the preview/delivery mode the SDK needs to also be able to resolve the values of the tokens directly since it cannot rely on user_interface.
For example in `preview`/`delivery` mode (and I think `editor` mode should work similarly) you would have an experience where somewhere in the `componentTree` would hold:
```
variableName: {
  type: "DesignToken",
  key: "padding.SM"
}
```
Since the actual value is not saved in the experience you'd need to look up that key into the DesignTokensRegistry.

That being said, this is more related to resolving the tokens than registering them so we could leave it out for now and add it when we implement that part.

packages/experience-builder-types/src/constants.ts
```diff
@@ -12,6 +12,7 @@ export const OUTGOING_EVENTS = {
   NewHoveredElement: 'newHoveredElement',
   ComponentSelected: 'componentSelected',
   RegisteredComponents: 'registeredComponents',
+  DesignTokens: 'registerDesignTokens',
```
SofiaMargariti, line 15:
```suggestion
  RegisterDesignTokens: 'registerDesignTokens',
```
nitpick: could you make the key the same as the value to be more clear?

SofiaMargariti (approved):
Just a nitpick and the comment about the registry is not a blocker for now

PR #174, opened 2023-12-04
Use rollup build []

elylucas:
A breaking change to the Visual SDK api exposed an issue in the SDK where the Visual SDK was being bundled with the package, so the SDK would always use whatever version it had installed instead of the version specified in the package dependencies. 

I tried to fix the issue by specifying additional rollup options to the Vite config, but I was running into weird (and not reproducible locally) build errors in Circle. I tried adding a rollup config and was able to get the build to work with it, so I'm putting this forth as an option to build the package. Vite was not intended to be a package bundler, and as a result, had some side effects that came along with it. Since Vite uses Rollup under the hood, this PR removes the unneeded Vite build and goes to Rollup directly. 

As a result, the package is now slimmer as no other unwanted dependencies get packaged into the SDK. They will be instead added by the clients bundler dependent on what is specified as dependencies and peer dependencies.

I also fixed the breaking changes that came along the new Visual SDK package and fixed some tests that were erroring as a result of it.


packages/experience-builder-sdk/src/core/preview/EntityStore.ts
```diff
@@ -59,7 +59,10 @@ export class EntityStore extends VisualSdkEntityStore {
     entityLink: UnresolvedLink<'Entry' | 'Asset'>,
     path: string[]
   ): string | undefined {
-    const entity = this.entitiesMap.get(entityLink.sys.id);
+    const entity =
+      entityLink.sys.linkType === 'Entry'
```
elylucas, line 63: Updated to reflect that `entitiesMap` was removed from the Visual SDK. I used `entryMap` and `assetMap` to accomplish fetching either an Entry or an Asset and used the `linkType` to determine which one to call.

packages/experience-builder-sdk/package.json
```diff
@@ -57,9 +62,13 @@
     "jest": "^29.5.0",
     "jest-environment-jsdom": "^29.5.0",
     "jsdom": "^21.1.0",
+    "postcss-import": "^15.1.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "relative-deps": "^1.0.7",
```
Spring3, line 68: was this dependency left on accident? (is also appears in the `prepare` command)

packages/experience-builder-sdk/package.json
```diff
@@ -9,9 +9,9 @@
   "license": "MIT",
   "scripts": {
     "dev": "vite",
-    "prepare": "node ./bin/injectSDKVersion.cjs",
-    "prebuild": "node ./bin/injectSDKVersion.cjs",
-    "build": "vite build",
+    "prepare": "node ./bin/injectSDKVersion.cjs && relative-deps",
+    "prebuild": "rimraf dist && node ./bin/injectSDKVersion.cjs",
+    "build": "rollup -c rollup.config.mjs",
     "watch": "vite build --watch",
```
Spring3, line 15: is there a chance we could entirely cut `vite` out?

It see that there is still this script, the plugins and the config file in the project

PR #175, opened 2023-12-05
feat!: register toolkit components and include wrap container by default [ALT-115]

primeinteger:
### Purpose

This change registers five basic components from `@contentful/experience-builder-components`
* Image
* Button
* Heading
* Text
* RichText

These five components and any components that the customer app registers with the SDK will now be wrapped in a container by default using the `withContainerWrapper` util.

The customer can use the ComponentRegistration options to change the behavior of the container wrapper.
* wrapComponent - `true` (default) wraps component in a container element and `false` can be used to opt-out and skip using a container wrapper
* wrapContainerTag - `'div'` (default) uses a \<div\> for the wrapper and other element types can be used like `'span'` if needed

This is potentially a breaking change (rendering the container wrapper by default for registered components) and should be a major version bump in the SDK.

### Dependency

Depends on change to toolkit components: https://github.com/contentful/experience-builder-toolkit/pull/48
- [x] TODO: bump package version for `@contentful/experience-builder-components` once this PR is merged

### Ticket

Jira: https://contentful.atlassian.net/browse/ALT-115

### DIFF
--- a/packages/experience-builder-sdk/package.json
+++ b/packages/experience-builder-sdk/package.json
@@ -9,7 +9,7 @@
   "license": "MIT",
   "scripts": {
     "dev": "vite",
-    "prepare": "node ./bin/injectSDKVersion.cjs",
+    "prepare": "node ./bin/injectSDKVersion.cjs && relative-deps",
     "prebuild": "node ./bin/injectSDKVersion.cjs",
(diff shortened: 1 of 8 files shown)


elylucas (approved):
Looks good after the package ref is updated!

PR #178, opened 2023-12-06
fix(experience-builder-types): extend variables type on component definition [ALT-57]

primeinteger:
Suggested by @chasepoirier, this changes the type for `variables` in the Component Definition object to include hints for the keys of built in style variables.

Users can set default values for the built in styles we provide so this extended type can help inform the variable names for styles we provide.

```diff
   variables: {
+    cfPadding: {
+      displayName: 'Padding',
+      type: 'Text',
+      group: 'style',
+      defaultValue: '0px 10px 0px 10px',
+    },
```


chasepoirier (approved):
lgtm 🚀

PR #193, opened 2023-12-13
fix: inject script for Visual Editor [ALT-169]

primeinteger:
Changes:
* Adds a new `setVisualEditorMode(mode)` method that can be used to configure the visual editor import using lazy loading (default) or script injection.
* When using script injection, the visual editor package is loaded from https://unpkg.com
* Changes how the visual editor is initialized, using window events to communicate the components registry and locale

Ticket: https://contentful.atlassian.net/browse/ALT-169


packages/experience-builder-sdk/src/blocks/VisualEditor.tsx
```diff
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { InternalSDKMode } from '../types';
-import { componentRegistry } from '../core/componentRegistry';
-import VisualEditor from '@contentful/experience-builder-visual-editor';
 import { VisualEditorContextProvider } from './editor/VisualEditorContext';
 
+const version = '0.0.1-pre-20231212T210325.0';
```
primeinteger, line 5: This variable can be any semver tag or version, for example:
```ts
const version = '^0.0.1-pre';
```

chasepoirier: Yeah this is a great start. In the future I think we can potentially look at having the user_interface communicate the build version via post message so there becomes no reliance on users updating their SDK to get the most recent version of the visual editor.

Definitely not a requirement right now

packages/experience-builder-sdk/src/blocks/VisualEditor.tsx
```diff
@@ -1,9 +1,27 @@
...
+    script.async = true;
+
+    document.body.appendChild(script);
+
+    return () => {
+      document.body.removeChild(script);
+    };
+  }, []);
+
+  return <div id="cf-visual-editor" />;
```
chasepoirier, line 22: Can we move this id value `cf-visual-editor` to a constant in the `core` package so that both the visual-editor and sdk can import that constant instead?

chasepoirier (changes requested):

packages/experience-builder-sdk/src/blocks/VisualEditor.tsx
primeinteger: That's a great idea. I'll work on this change

chasepoirier (approved):

PR #194, opened 2023-12-14
fix: render component values in preview and resolve links properly []

Chaoste:
Apply the same changes as in https://github.com/contentful/experience-builder/pull/192 but for preview mode where we have a different structure (the raw data format).

Also, the function `EntityStore.getValue` renders warnings when it receives full entities instead of links. The links are already resolved by default through our APIs, thus the logic needs to be aware of both cases.

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
@@ -111,12 +111,8 @@ export const CompositionBlock = ({
         }
         case 'UnboundValue': {
           const uuid = variable.key;
+          console.log(variableName, uuid, acc[variableName]);
           acc[variableName] = (entityStore?.unboundValues || unboundValues)[uuid]?.value;
-          break;
-        }
-        case 'ComponentValue': {
-          const uuid = variable.key;
-          acc[variableName] = unboundValues[uuid]?.value;
           break;
         }
         default:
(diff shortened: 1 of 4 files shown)


packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -111,14 +111,10 @@ export const CompositionBlock = ({
         }
         case 'UnboundValue': {
           const uuid = variable.key;
+          console.log(variableName, uuid, acc[variableName]);
```
SofiaMargariti, line 114: Is this still needed?

packages/experience-builder-sdk/src/core/preview/designComponentUtils.ts
```diff
@@ -1,7 +1,54 @@
...
+  const children: CompositionNode[] = node.children.map((child) =>
+    deserializeDesignComponentNode({
+      node: child,
+      componentInstanceVariables,
+    })
+  );
+
+  return {
+    // separate node type identifiers for design components and their blocks, so we can treat them differently in as much as we want
+    // type: isDesignComponent ? DESIGN_COMPONENT_NODE_TYPE : DESIGN_COMPONENT_BLOCK_NODE_TYPE,
```
SofiaMargariti, line 45: Do we still need to set the type here? I tested it and seems to work fine without it. Should this comment be removed?

SofiaMargariti (approved):
Looks good and works as expected! Just some nitpick comments.

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
Chaoste: Hupps, good catch :)

packages/experience-builder-sdk/src/core/preview/designComponentUtils.ts
Chaoste: Yes, you're right. Thanks for pointing out :)

PR #195, opened 2023-12-14
feat: resolves the design token values [ALT-171]

ryunsong-contentful:
## Approach
Only do this for margin and padding so we can first agree on the implementation structure and then once agreed we can apply it for all other tokens like direction, background color, border.

Additionally, the strategy is as follows
Use a set to determine which potential variables (cfMargin, cfPadding) could have template strings
If the variable is one of those types which could have a template string, run through a specific parser that will know how to handle the given type in the `designTokenRegistry` file. This way we can map specific parsers to specific variable types which may share or differ in their structure. Therefore as we add a new variable type (ie. border in a separate PR) we can refactor this code to generalize as we go along

## Purpose
Stores the design token key instead of the hard coded value, and then is able to resolve the design token key and display the value instead.

Screenshot 2023-12-14 at 12 34 31 PM

## TODOs
- The other design tokens such as direction, size, border, background color
- Pills to show the user that what they've selected is a design token

### DIFF
--- a/packages/experience-builder-sdk/src/core/designTokenRegistry.ts
+++ b/packages/experience-builder-sdk/src/core/designTokenRegistry.ts
@@ -2,13 +2,43 @@ import { OUTGOING_EVENTS } from '../constants';
 import { sendMessage } from '../communication/sendMessage';
 import { DesignTokensDefinition } from '../types';
 
+const designTokensRegistry = {} as DesignTokensDefinition;
+const templateStringRegex = /\$\{([^}]+)\}/g;
+
 /**
  * Register design tokens styling
  * @param designTokenDefinition - {[key:string]: Record<string, string>}
  * @returns void
  */
 export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition) => {
+  Object.assign(designTokensRegistry, designTokenDefinition);
   sendMessage(OUTGOING_EVENTS.DesignTokens, {
     designTokens: designTokenDefinition,
   });
+};
+
+export const getDesignTokenRegistrationForSpacing = (breakpointValue: string) => {
+  let resolvedValue = '';
+  const values = breakpointValue.split(' ');
+  values.forEach((value) => {
+    let tokenValue = value;
+    if (isTemplateStringFormat(value)) tokenValue = resolveSpacingDesignToken(value);
+    resolvedValue += `${tokenValue} `;
+  });
+
+  return resolvedValue;
+};
+
+// Using this because export const StringTemplateRegex = /\${(.*?)\}/g doesn't work
+const isTemplateStringFormat = (str: string) => {
+  return templateStringRegex.test(str);
+};
+
+const resolveSpacingDesignToken = (templateString: string) => {
+  if (!templateStringRegex.test(templateString)) return templateString;
+  const nonTemplateValue = templateString.replace(templateStringRegex, '$1');
+  const designKeys = nonTemplateValue.split('.');
+  const spacingValues = designTokensRegistry[designKeys[0]] as DesignTokensDefinition;
+  const resolvedValue = spacingValues[designKeys[1]] as string;
+  return resolvedValue || '0px';
 };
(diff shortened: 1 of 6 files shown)


package-lock.json
primeinteger, line 1: I've accidently generated this file before too, but since this repo only uses `yarn`, I think you can delete this

yarn.lock
primeinteger, line 1: The changes to this file were possibly generated by `npm`.  When this happens to me, I delete node_modules directory and regenerate the yarn.lock file using `yarn`

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
```diff
@@ -26,6 +28,8 @@ const toCSSMediaQuery = ({ query }: Breakpoint): string | undefined => {
   return undefined;
 };
 
+const AvailableDesignTokenVariables = new Set(['cfPadding', 'cfMargin']);
```
primeinteger, line 31: The uppercase first letter is standard for component names and types/interfaces but uncommon for variable names.
```suggestion
const availableDesignTokenVariables = new Set(['cfPadding', 'cfMargin']);
```

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
```diff
@@ -46,14 +50,19 @@ const getFallbackBreakpointIndex = (breakpoints: Breakpoint[]) => {
...
+  variableName: string
 ) => {
   const fallbackBreakpointIndex = getFallbackBreakpointIndex(breakpoints);
   const fallbackBreakpointId = breakpoints[fallbackBreakpointIndex].id;
   if (valuesByBreakpoint instanceof Object) {
     // Assume that the values are sorted by media query to apply the cascading CSS logic
     for (let index = activeBreakpointIndex; index >= 0; index--) {
       const breakpointId = breakpoints[index].id;
+      if (AvailableDesignTokenVariables.has(variableName)) {
+        if (variableName === 'cfMargin' || variableName === 'cfPadding')
```
primeinteger, line 63: The if statements could be combined
```suggestion
      if (AvailableDesignTokenVariables.has(variableName) && ['cfMargin',  'cfPadding'].includes(variableName)) {
```

packages/experience-builder-sdk/src/core/designTokenRegistry.ts
```diff
@@ -2,13 +2,44 @@ import { OUTGOING_EVENTS } from '../constants';
...
+  });
+
+  return resolvedValue;
+};
+
+// Using this because export const StringTemplateRegex = /\${(.*?)\}/g doesn't work
+const isTemplateStringFormat = (str: string) => {
+  if (!str || str.length < 3) return false;
+  return str.charAt(0) === '$' && str.charAt(1) === '{' && str.charAt(str.length - 1) === '}';
+};
```
primeinteger, line 35: The regex you added to the `resolveSpacingDesignToken` method could be used here if you wanted
```suggestion
const isTemplateStringFormat = (str: string) => {
  const regex = /\$\{([^}]+)\}/g;
  return regex.test(str);
};
```

primeinteger (approved):
Nice work on this!

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
ryunsong-contentful: I purposely did this so that when there's other attributes like border or whatever, I will add a separate clause for that so that I can direct that to the `getDesignTokenRegistrationForBorder` function. So this is the structure I want for the future even though today, in this PR you can shorten it like this. Hopefully what I wrote makes sense

ryunsong-contentful:
```
      if (AvailableDesignTokenVariables.has(variableName)) {
        if (variableName === 'cfMargin' || variableName === 'cfPadding')
          return getDesignTokenRegistrationForSpacing(valuesByBreakpoint[breakpointId]);
        if(variableName === 'cfBorder')return 
          getDesignTokenRegistrationForBorder(valuesByBreakpoint[breakpointId]);
      }
```

primeinteger: Got it, makes sense 👍 

If enough values are checked on the inner part, a `switch` might be preferable too

PR #196, opened 2023-12-15
fix: calculate component client rect via children [SPA-1652]

Chaoste:
`display:contents` leads to having no "box" in the actual DOM as somehow mentioned [here](https://drafts.csswg.org/css-display/#transformations) and [here](https://dev.to/ben/how-would-you-go-about-detecting-the-height-of-content-within-an-iframe-14d8#comment-f17j).

Thus, it fails to calculate any bounding client rect for this node. As a work around we calculate the size of each child and add them up.

### DIFF
--- a/packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
+++ b/packages/experience-builder-sdk/src/communication/MouseOverHandler.ts
@@ -1,5 +1,5 @@
 import { HoveredElement } from '../types';
-import { OUTGOING_EVENTS } from '../constants';
+import { DESIGN_COMPONENT_NODE_TYPE, OUTGOING_EVENTS } from '../constants';
 import { sendMessage } from './sendMessage';
 
 export class MouseOverHandler {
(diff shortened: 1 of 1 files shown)


dimitrycf:
I tried tracking the moment when this newly added logic gets executed. And it doesn't seem to be executed when hovering over {design components} 🤷 

https://www.loom.com/share/e4507b870e7c45bdb17ecd05ace057b9

Chaoste:
> I tried tracking the moment when this newly added logic gets executed. And it doesn't seem to be executed when hovering over {design components} 🤷
> 
> https://www.loom.com/share/e4507b870e7c45bdb17ecd05ace057b9

As written in Slack, your video is showing how it should be have. So all good here :) Note that this only about hovering instances via the layers tab.

SofiaMargariti:
> I tried tracking the moment when this newly added logic gets executed. And it doesn't seem to be executed when hovering over {design components} 🤷
> 
> https://www.loom.com/share/e4507b870e7c45bdb17ecd05ace057b9

When hovering with the mouse you're not actually hovering on the surrounding div but on the child container, so the `getBoundingClientRect` will work. You have to hover on the component from the Layers tab to replicate the issue.

dimitrycf (approved):
LGTM

PR #200, opened 2023-12-18
feat: initialise entity store sustainably in editor mode [SPA-1711]

Chaoste:
During the first initialization phase, the entity store is switching often between a "filled" and an "empty" state. When we're unlucky, somewhere in between it tries to render design component blocks but they are not yet in the entity store as it was emptied again. As a result, we see this warning everytime in the console and in some scenarios, the component is not rendered at all (e,g, colorful coin demo):
> `Entry for design component with ID '${componentId}' not found`

### 1. Recycle fetched entities from preview mode
The SDK will always start in non-editor mode and already load the entities in the preview entity store. After it switches to editor mode, it won't set the list of entities to `[]` but to `previousEntityStore?.entities` to eventually reuse already loaded entities.

### 2. Only reset entity store when locale actually changed
We have an effect that should re-initialize the entity store when the locale has changed. This would usually also get called in the first render cycle and therefore drop the preload entities. By adding a check `locale === entityStore.current.locale`, we are unsure to only reset it when necessary.

### 3. One place to manage the entity store
I moved logic that would change the entity store all in one place. Before, the "root" component would affect the context. This is now all happening in one single place (context).
_As a future improvement, we could extract this again into a separate hook, so the context becomes lighter again._

## Positive Side Effect
When changing code in the SDK, our demo project would hot reload but since a few versions ago, the canvas didn't catch up properly and just stayed empty. With these improvements, the canvas is now again able to hot reload, saving us from reloading the only web app every time we test changes in the SDK or demo.

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -18,6 +18,7 @@ import {
   CompositionTree,
   CompositionUnboundValues,
   InternalSDKMode,
+  EntityStore,
   Link,
 } from '../../types';
 import { INCOMING_EVENTS, OUTGOING_EVENTS, SCROLL_STATES, INTERNAL_EVENTS } from '../../constants';
@@ -37,6 +38,7 @@ type VisualEditorContextType = {
   unboundValues: CompositionUnboundValues;
   breakpoints: Breakpoint[];
   entityStore: React.MutableRefObject<EditorModeEntityStore>;
+  areEntitiesFetched: boolean;
 };
 
 export const VisualEditorContext = React.createContext<VisualEditorContextType>({
@@ -51,12 +53,14 @@ export const VisualEditorContext = React.createContext<VisualEditorContextType>(
   locale: null,
   breakpoints: [],
   entityStore: {} as React.MutableRefObject<EditorModeEntityStore>,
+  areEntitiesFetched: false,
 });
 
 type VisualEditorContextProviderProps = {
   initialLocale: string;
   mode: InternalSDKMode;
   children: ReactElement;
+  previousEntityStore?: EntityStore;
 };
 
 export const designComponentsRegistry = new Map<string, Link<'Entry'>>([]);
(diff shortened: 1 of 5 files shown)


packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -78,14 +83,43 @@ export function VisualEditorContextProvider({
...
+    if (!locale || locale === entityStore.current.locale) return;
+    entityStore.current = new EditorModeEntityStore({
+      entities: [],
+      locale: locale,
+    });
+  }, [locale]);
+
+  // When the tree was updated, we store the dataSource and
+  // afterward, this effect fetches the respective entities.
+  useEffect(() => {
```
SofiaMargariti, line 109: Was there a reason for moving this useEffect in the context? Thinking about the issues we were seeing with contexts triggering unnecessary network requests, isn't it better for this useEffect to stay in the component so it doesn't accidentally trigger every time this context is used?

Chaoste: Technically, this doesn't change anything. The `<VisualEditorContextProvider>` is only used once in the wrapper of `<VisualEditorRoot>`. So on every render cycle, the logic will once run through the context component **and** once through the root component. It's triggered exactly the same amount of times as before.

Note that `useContext` doesn't render any effects but just returns the generated object that is returned by the provider. So, consuming the context values via `useContext` is not the same as "using"/ "rendering" the context multiple times.

SofiaMargariti: You right, the problem we've seen in user_interface was not related to `useContext` but regular hooks that included `useEffect`

SofiaMargariti (approved):
Looks good!

I noticed one strange thing where after the reload the bindings are lost and it falls back to the default values, but it seems to be an issue on main as well so I am approving the PR.
You can see in the recording below - the initial image comes from an asset and all changes are saved.

https://github.com/contentful/experience-builder/assets/2773012/398953b0-82aa-426b-a938-f257f9326777

PR #201, opened 2023-12-18
chore: toolkit repo migration

chasepoirier:
This PR is a successful merge of the Toolkit repo into the SDK repo which was performed by following these steps: https://contentful.atlassian.net/wiki/spaces/PROD/pages/4413883469/2023-12-11+Migration+to+Single+Repo


.circleci/config.yml
```diff
@@ -12,12 +12,12 @@ commands:
   install-dependencies:
     steps:
```
elylucas, line 13: looks like we have circle and gha in here. Did we mean to keep one over the other?

packages/experience-builder-sdk/package.json
```diff
@@ -18,16 +18,16 @@
...
     "prettier:check": "prettier --check '**/*.{jsx,js,ts,tsx,css}'",
     "prettier": "prettier --write '**/*.{jsx,js,ts,tsx,css}'",
     "preview": "vite preview"
   },
   "dependencies": {
-    "@contentful/experience-builder-components": "^0.0.1-pre-20231217T133016.0",
-    "@contentful/experience-builder-core": "^0.0.1-pre-20231217T133016.0",
-    "@contentful/experience-builder-visual-editor": "^0.0.1-pre-20231217T133016.0",
+    "@contentful/experience-builder-components": "file:../components",
+    "@contentful/experience-builder-types": "file:../experience-builder-types",
```
elylucas, line 29: the reference to types should be removed from the sdk

elylucas (changes requested):
A question on CI providers and one minor change

primeinteger (approved):
🚀

.circleci/config.yml
chasepoirier: @elylucas  Good call out. I think we should just use GHA so should I comment out the Circle job or actually delete the .circle directory?

elylucas: We can remove it. If we need to get back to it later we can

chasepoirier:
> A question on CI providers and one minor change

Seems like our PRs will fail due to a requirement of having Circle. I've instead modified the job so that it doesn't really do anything and all real actions are handled by GHA.



We can revisit removing or modifying Circle once the changes hit main

elylucas (approved):

PR #202, opened 2023-12-19
fix: enable typography styles on the default Text component [ALT-144]

primeinteger:
Adds the config for new Typography built-in styles to the SDK/core.
* The new Typography styles are optional and opt-in per component.
 
Enables the new styles on the default Text component.
* I'll add new input controls to the UI for these new styles next.


packages/components/src/components/Text/index.ts
```diff
@@ -7,7 +7,20 @@ export const TextComponentDefinition: ComponentDefinition = {
   id: 'text',
   name: 'Text',
   category: 'Contentful',
-  builtInStyles: ['cfMargin', 'cfPadding'],
+  builtInStyles: [
+    'cfMargin',
+    'cfPadding',
+    'cfFontSize',
+    'cfFontWeight',
```
ryunsong-contentful, line 14: This is more of a product comment. I think to start, maybe it's useful to have `FontWeight` and `TextBold` and see how it plays with the usability of our text component, but I could also see it being redundant.

packages/core/src/definitions/styles.ts
```diff
@@ -172,3 +171,114 @@ export const builtInStyles: Record<
...
+    group: 'style',
+    description: 'The letter spacing of the section',
+    defaultValue: '0px',
+  },
+  cfTextColor: {
+    displayName: 'Text Color',
+    type: 'Text',
+    group: 'style',
+    description: 'The text color of the section',
+    defaultValue: 'rgba(0, 0, 0, 1)',
```
ryunsong-contentful, line 211: Maybe consider f36 black token? Could be a travis question

ryunsong-contentful (approved):
Love it, super clean

packages/core/src/definitions/styles.ts
primeinteger: The default background color style defined in this file uses `rgba` so I wanted to keep it consistent with that

```ts
  cfBackgroundColor: {
    displayName: 'Background',
    type: 'Text',
    group: 'style',
    description: 'The background color of the section',
    defaultValue: 'rgba(255, 255, 255, 0)',
  },
```

I think the SDK does not include F36 tokens because it would increase the bundle size and the SDK is made to be as lightweight as possible.

packages/components/src/components/Text/index.ts
primeinteger: Yep, the `cfTextBold` boolean will map to a "quick toggle" bold button while the `cfFontWeight` string will map to a select or input that has more options for various weights.

Both fields will be hooked up to the css styles generator for `font-weight` style, so the UI will need to account for the logic when both options are enabled in a component (but I don't think it will be too bad)

PR #203, opened 2023-12-19
chore(core): fix typings issue

elylucas:
This PR fixes a typings bug. Before, the typings would export code that would include the internal paths imports (IE: `import { blah } from '@types'. These should be resolved to their relative paths, which this PR fixes.


primeinteger (approved):
Looks good!  I see the `VisualEditorMode` enum was already moved to a different file.

PR #206, opened 2023-12-20
feat: add useFetchBySlug, useFetchById hooks and plain js fetchBySlug and fetchById methods

elylucas:
This PR backs out the changes made in #161, as when I merged it I thought of a better way to handle the situation. Instead of overloading the `useFetchExperience` method to handle fetching by slug and id, we now expose two new methods, `useFetchById` and `useFetchBySlug`. Both these methods are reactive like the changes in 161. 

I left the `userFetchExperience` as is before 161, but marked it as deprecated in favor of the new methods. This makes the previous changes in 161 that were breaking no longer breaking.

Also, this was a good opportunity to expose some vanilla JS `fetchBySlug` and `fetchById` methods that are framework agnostic, so I refactored those out of the hook and moved them into the core library (and re-exported from sdk). These will be useful for those that want plain async methods to fetch an experience, like from the server side or when using libraries like React Query to manage the state.


chasepoirier (approved):
Looks good to me!

Two key benefits I see here; 1) an easier client set up 2) another step closer to framework agnostic by making the `core` package framework agnositc.

Love it!

packages/experience-builder-sdk/src/hooks/useFetchBySlug.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+        slug,
+        experienceTypeId,
+        localeCode,
+      },
+    });
+
+    await waitFor(() => {
+      expect(result.current).toBeDefined();
+      expect(result.current.isLoading).toBe(true);
+    });
```
Chaoste, line 43: [nitpick]
It's not breaking the test but more in general, I want to call out that `waitFor` should only wait for one assertion at a time. You can read more about the reasoning behind this [here](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback).

```suggestion
    // Just wait until we reached a specific state after async things happened
    await waitFor(() => expect(result.current).toBeDefined());
    // Then run all the sync assertions as it is not expected that something else async will change the state again
    expect(result.current.isLoading).toBe(true);
```

packages/experience-builder-sdk/src/hooks/useFetchByBase.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react';
+import { EntityStore } from '@contentful/experience-builder-core';
+import type { Experience } from '@contentful/experience-builder-core/types';
+
+export const useFetchByBase = (fetchMethod: () => Promise<Experience<EntityStore> | undefined>) => {
```
Chaoste, line 5: I really like that this new helper hook enriches async methods with a loading & error state.

packages/experience-builder-sdk/src/hooks/useFetchById.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from 'react';
+import type { ContentfulClientApi } from 'contentful';
+import type { ExternalSDKMode } from '@contentful/experience-builder-core/types';
+import { useFetchByBase } from './useFetchByBase';
+import { fetchById } from '@contentful/experience-builder-core';
+
+export type useFetchByIdArgs = {
```
Chaoste, line 7: [nitpick]
Following our CF-internal naming styles, this would usually be written like this:
```suggestion
export type UseFetchByIdProps = {
```
- A type is always written in UpperCamelCase
- The type describing the props for a hook or component uses always the suffix "Props"

It's really just a nitpick but I think it's a nicer DX if we use similar casing and prefixes/ suffixes for such things :)

Chaoste (approved):
Nice work, thank you!  🚀

PR #209, opened 2023-12-20
fix: switching locale makes design components disappear [SPA-1711]

Chaoste:
I discovered a few flaws that might end up in the design components and content not loading at all initially. I implements some small tweaks that are likely to fix this :)

In some cases, we already have a pre-filled entity store. This happens sometimes as we start in preview mode and then switch over to editor mode taking over the already-fetched entities. It also happens when using hot reloading.
In those cases, we should try to render the tree which we currently block via the boolean `areEntitiesFetched`.

To not silently just render nothing, we now try to render from the beginning on by setting `areEntitiesFetched` to true initially.

Also, we only update this variable only and only if we are really fetching something. The editor entity store is not returning either a boolean or a promise depending on whether it needs to fetch or not.

Last but not least, the design component registry is sometimes empty even though we update it when the new tree update arrives. I suspect a wrong order of the logic and thus switch the order of code lines when receiving the new tree.

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -63,6 +63,9 @@ type VisualEditorContextProviderProps = {
   previousEntityStore?: EntityStore;
 };
 
+// Note: During development, the hot reloading might empty this and it
+// stays empty leading to not rendering design components. Ideally, this is
+// integrated into the state machine to keep track of its state.
 export const designComponentsRegistry = new Map<string, Link<'Entry'>>([]);
 export const setDesignComponents = (designComponents: Link<'Entry'>[]) => {
   for (const designComponent of designComponents) {
@@ -83,7 +86,9 @@ export function VisualEditorContextProvider({
   const [isDragging, setIsDragging] = useState(false);
   const selectedNodeId = useRef<string>('');
   const [locale, setLocale] = useState<string>(initialLocale);
-  const [areEntitiesFetched, setEntitiesFetched] = useState(false);
+  // Optimistic loading. Try to render everything as early as possible.
+  // The entities might already be defined in the entity store.
+  const [areEntitiesFetched, setEntitiesFetched] = useState(true);
 
   const [entityStore, setEntityStore] = useState<EditorModeEntityStore>(
     () =>
@@ -115,12 +120,13 @@ export function VisualEditorContextProvider({
   // afterward, this effect fetches the respective entities.
   useEffect(() => {
     const resolveEntities = async () => {
-      setEntitiesFetched(false);
       const dataSourceEntityLinks = Object.values(dataSource || {});
-      await entityStore.fetchEntities([
-        ...dataSourceEntityLinks,
-        ...(designComponentsRegistry.values() || []),
-      ]);
+      const entityLinks = [...dataSourceEntityLinks, ...(designComponentsRegistry.values() || [])];
+      const deferredEntities = entityStore.fetchEntities(entityLinks);
+      // Only update the state and rerender when we're actually fetching something
+      if (deferredEntities === false) return;
+      setEntitiesFetched(false);
+      await deferredEntities;
       setEntitiesFetched(true);
     };
 
@@ -221,9 +227,10 @@ export function VisualEditorContextProvider({
             changedNode?: CompositionComponentNode;
             changedValueType?: CompositionComponentPropValue['type'];
           } = payload;
+          // Make sure to first store the design components before setting the tree and thus triggering a rerender
+          designComponents && setDesignComponents(designComponents);
           setTree(tree);
           setLocale(locale);
-          designComponents && setDesignComponents(designComponents);
 
           if (changedNode) {
             /**
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -228,7 +228,9 @@ export function VisualEditorContextProvider({
             changedValueType?: CompositionComponentPropValue['type'];
           } = payload;
           // Make sure to first store the design components before setting the tree and thus triggering a rerender
-          designComponents && setDesignComponents(designComponents);
+          if (designComponents) {
+            setDesignComponents(designComponents);
+          }
           setTree(tree);
           setLocale(locale);
 
@@ -296,7 +298,9 @@ export function VisualEditorContextProvider({
         }
         case INCOMING_EVENTS.UpdatedEntity: {
           const { entity } = payload;
-          entity && entityStore.updateEntity(entity);
+          if (entity) {
+            entityStore.updateEntity(entity);
+          }
           break;
         }
         case INCOMING_EVENTS.RequestEditorMode: {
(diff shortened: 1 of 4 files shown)


packages/experience-builder-sdk/src/core/editor/EditorModeEntityStore.ts
```diff
@@ -40,18 +40,23 @@ export class EditorModeEntityStore extends EditorEntityStore {
...
 
     // Entries and assets will be stored in entryMap and assetMap
-    await Promise.allSettled([
-      this.fetchEntries([...uniqueEntryLinks]),
-      this.fetchAssets([...uniqueAssetLinks]),
-    ]);
+    return Promise.all([this.fetchEntries(uniqueEntryIds), this.fetchAssets(uniqueAssetIds)]).then(
+      ([entries, assets]) => [...entries, ...assets]
+    );
   }
```
dimitrycf, line 60: The fact that `fetchEntities()` returns strict `false` is a bit surprising (not here, but from the where fetchEntiteis() is called). 

Also the `fetchEntities()` is de-facto an async function (as it returns a promise). What do you think of making it consistent via this snippet:

```ts
  async fetchEntities(entityLinks: UnresolvedLink<'Entry' | 'Asset'>[]) : Promise<false | Array<Entry|Asset>> {
    const entryLinks = entityLinks.filter((link) => link.sys?.linkType === 'Entry');
    const assetLinks = entityLinks.filter((link) => link.sys?.linkType === 'Asset');

    const uniqueEntryIds = [...new Set(entryLinks.map((link) => link.sys.id))];
    const uniqueAssetIds = [...new Set(assetLinks.map((link) => link.sys.id))];

    const { missing: missingEntryIds } = this.getEntitiesFromMap('Entry', uniqueEntryIds);
    const { missing: missingAssetIds } = this.getEntitiesFromMap('Asset', uniqueAssetIds);

    // Return false to indicate that no async fetching is happening
    if (!missingAssetIds.length && !missingEntryIds.length) return false;

    // Entries and assets will be stored in entryMap and assetMap
    return await Promise.all([this.fetchEntries(uniqueEntryIds), this.fetchAssets(uniqueAssetIds)]).then(
      ([entries, assets]) => [...entries, ...assets]
    );
  }
```

Do you think there's a benefit in adding type to the function? 

```
```

packages/experience-builder-sdk/src/core/designTokenRegistry.ts
```diff
@@ -17,7 +17,9 @@ export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition
   });
 };
 
-export const getDesignTokenRegistration = (breakpointValue: string) => {
+export const getDesignTokenRegistration = (breakpointValue?: string) => {
+  if (!breakpointValue) return breakpointValue;
+
   let resolvedValue = '';
```
dimitrycf, line 23: Looks like you had some specific edge case in mind. 
But to me the semantics of this looks awkward: 

> take any falsy value and return back _that same_ falsy value 🤷 ...

With such a semantics, it's hard for me to understand the intent of this. (to handle empty strings? to handle undefines? there's also no hint of what is the range of allowed values ....

If it would have been always returning `null` or `false` as result, then I would be able to read it as "when falsy crap comes in, just return false (as marker for no value)" but now if `""`  (empty string) comes, it also returns back the same empty string? 🤷

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -221,9 +227,10 @@ export function VisualEditorContextProvider({
             changedNode?: CompositionComponentNode;
             changedValueType?: CompositionComponentPropValue['type'];
           } = payload;
+          // Make sure to first store the design components before setting the tree and thus triggering a rerender
+          designComponents && setDesignComponents(designComponents);
```
dimitrycf, line 231: [subjective] 🙄 I also love the  hacky `&&`  pattern :) (and it works great for JSX), but my understanding was that for non-JSX code this should be used only on special occasions ;)

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -115,12 +120,13 @@ export function VisualEditorContextProvider({
...
-      setEntitiesFetched(false);
       const dataSourceEntityLinks = Object.values(dataSource || {});
-      await entityStore.fetchEntities([
-        ...dataSourceEntityLinks,
-        ...(designComponentsRegistry.values() || []),
-      ]);
+      const entityLinks = [...dataSourceEntityLinks, ...(designComponentsRegistry.values() || [])];
+      const deferredEntities = entityStore.fetchEntities(entityLinks);
+      // Only update the state and rerender when we're actually fetching something
+      if (deferredEntities === false) return;
```
dimitrycf, line 127: fthis ` === false` is unexpected because return values of the `fetchEntities()` is not obvious. I made comment on this on the `fetchEntities()` code below.

And IMO function which returns sync `false` and `Promise<Entry[]>` is very confusing. And even the call pattern is unusual because of that :)

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -115,12 +120,13 @@ export function VisualEditorContextProvider({
...
+      setEntitiesFetched(false);
+      await deferredEntities;
```
dimitrycf, line 129: Whilst this gymnastics with moving the  `setEntitiesFetched()` looks cool, I couldn't see through to why this would make any difference. 

The `resolveEntities()` runs top to bottom (aka "runs to completion JS mechanism") hence whether it's on top or bottom it will be executed in the same frame of the event loop.

dimitrycf (approved):
Approving to unblock Thomas.
Feel free to address my review comments as to your liking.

packages/experience-builder-sdk/src/core/editor/EditorModeEntityStore.ts
Chaoste: There is a problem: When I make the `false` response part of the promise. Then I need to "await" it completely. The current logic relies on a synced "false" response. If it's not "false" but a promise, I update the state variable `areEntitiesFetched` to reflect that it just started rerendering. That would not be possible if I transform it like you proposed.
Though, I agree that it's not the most elegant way. I will remember to get back to this if we figure out that the loading bugs are gone. If so, I will try reverting parts of this PR so we have a more beautiful code :)

packages/experience-builder-sdk/src/core/designTokenRegistry.ts
Chaoste: The reason for that is that I don't want to change the behavior. Unfortunately, the new design token logic by @ryunsong-contentful broke some of the existing design variable logic (e.g. the broken rendering of default height and width). This function was added recently as some kind of middleware.
With my change, I want to "turn off" the middleware logic if it's not appliable. So if there is an empty string, undefined, or another falsy value, don't try to transform it anyway. This way, this function is agnostic of any variable-specific default values which might be one of the previously mentioned falsy values.

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: You're right. It's not good - I just moved lines here without thinking. I will refactor this part :)

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: You're right.
I changed the response of the function and added an explanatory JSDoc why this is helpful for us to avoid re-renders.

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: Calling the function, triggers eventually async behavior which can be joined later via `await`. We make use of this trick, to first check whether the function is actually doing something async. If not, we know that it didn't fetch anything and thus we should not rerender. I added more comments to make this visible.

PR #210, opened 2023-12-20
fix: design token logic breaks auto height for empty containers []

Chaoste:
image


dimitrycf (approved):
Thanks for fixing this problem so fast! ⚡

PR #214, opened 2023-12-20
fix: update text transform label and pull builtInStyles from core package [ALT-227]

primeinteger:
* Changed text transform's `None` label to `Normal` to match the design verbiage
* Removed the duplicate `builtInStyles` definition
  * Importing this variable from `@contentful/experience-builder-core` instead of  the core directory within the sdk


packages/core/src/definitions/styles.ts
```diff
@@ -96,7 +96,7 @@ export const builtInStyles: Partial<
     type: 'Text',
     group: 'style',
     description: 'The orientation of the section',
-    defaultValue: 'row',
+    defaultValue: 'column',
```
primeinteger, line 99: This value changed to `'column'` in the sdk
- Same line in the file being deleted: [variables.ts#L102](https://github.com/contentful/experience-builder/pull/214/files#diff-5391ff296f3c429f92a2d0db7ace344511b03a37567e20c9687ce439abfa4b1fL102)

ryunsong-contentful (approved):

PR #217, opened 2023-12-22
fix: improve initial fetching mechanism [SPA-1711]

Chaoste:
The current fetching logic for initial entities (data source and design components) was sometimes not fetching entities correctly and failed thereby to render. While we identified the root cause of the user interface being too slow for the initial fetching, this PR is adding more transparency to this process by adding debug, warning, and error logs in unexpected cases.

The biggest logical change in this PR is the function `fetchMissingEntities` which is basically the previous effect but in a new callback. This function is now also imperatively called when we receive the design components. I added this as design components are not part of the React state and thus won't cause a re-rendering solely through this change. To avoid possible race conditions related to this, we're explicitly trying to fetch again when we're in this situation. 

~Also, I'm checking again the entities cache just before the JSX block as again we can't memoize data that is not in the React state (yes, I'm looking at you, `deisgnComponentsRegistry`!). Doing so, we can rely more on the true state of `areEntitiesFetched`.~

Update: To avoid infinite loops when entities are missing completely, I removed the manual map lookup and went back to fetch only once initially. I renamed the respective state variable, so it's clearer that this is about the initial fetch and is not intended to fetch automatically when something changed in the dataSource. As @SofiaMargariti pointed out, the new boolean `areEntitesResolvedInParent` (currently in the `development` branch) will make sure that our check runs once but also when the user interface is ready to handle it.

**Notes for the future:**
- I think that this boolean `areEntitiesFetched` is not the best approach. Every time we fetch something, we don't render any bindings on the canvas for a short moment which leads to a blinking in the UI. However, right now we need to do this as only through this boolean variable we force the child components to re-render and load the entities from the entitiy store. If we had some *more sophisticated state machine with selectors*, the components would be able to re-render as soon as the needed entities are in the store. By that, we could drop this boolean variable and don't have this blinking UI.
- The `designComponentsRegistry` should be moved to the React state. The current tweaks are not nice maintainable code. Alternatively, we could use `zustand` and use a selector which would ensure a re-rendering in React. Probably not a nice solution as we're trying to keep the SDK as lightweight as possible (though zustand is very lightweight).

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -38,7 +38,7 @@ type VisualEditorContextType = {
   unboundValues: CompositionUnboundValues;
   breakpoints: Breakpoint[];
   entityStore: EditorModeEntityStore;
-  areEntitiesFetched: boolean;
+  areInitialEntitiesFetched: boolean;
 };
 
 export const VisualEditorContext = React.createContext<VisualEditorContextType>({
@@ -53,7 +53,7 @@ export const VisualEditorContext = React.createContext<VisualEditorContextType>(
   locale: null,
   breakpoints: [],
   entityStore: {} as EditorModeEntityStore,
-  areEntitiesFetched: false,
+  areInitialEntitiesFetched: false,
 });
 
 type VisualEditorContextProviderProps = {
@@ -86,9 +86,7 @@ export function VisualEditorContextProvider({
   const [isDragging, setIsDragging] = useState(false);
   const selectedNodeId = useRef<string>('');
   const [locale, setLocale] = useState<string>(initialLocale);
-  // Optimistic loading. Try to render everything as early as possible.
-  // The entities might already be defined in the entity store.
-  const [isFetchingEntities, setFetchingEntities] = useState(false);
+  const [areInitialEntitiesFetched, setInitialEntitiesFetched] = useState(false);
 
   const [entityStore, setEntityStore] = useState<EditorModeEntityStore>(
     () =>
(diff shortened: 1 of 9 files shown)


packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -116,28 +116,42 @@ export function VisualEditorContextProvider({
...
+  // Either gets called when dataSource changed or designComponetsRegistry changed (manually)
+  const fetchMissingEntities = useCallback(async () => {
+    if (isFetchingEntities) return;
+    const requiredEntityLinks = [
+      ...Object.values(dataSource),
+      ...designComponentsRegistry.values(),
+    ];
+    const { missingAssetIds, missingEntryIds } =
+      entityStore.getMissingEntityIds(requiredEntityLinks);
+    const areEntitiesFetched = !missingAssetIds.length && !missingEntryIds.length;
```
SofiaMargariti, line 128: Will this cause any issues if an entity is legitimately missing, e.g. if it has been deleted from the space?

SofiaMargariti:
Do we still need the fetchMissingEntities logic if you're going to implement the fix proposed by Chase [here](https://contentful.slack.com/archives/C01R9P3CBH7/p1703233722005279?thread_ts=1703233662.635659&cid=C01R9P3CBH7)?

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: Mmmh, you're right that this probably ends up in an infinite loop of fetching.
I will revert my fix here and reduce the PR to the quality of life improvements. His approach might be more stable in that sense.

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -116,28 +115,35 @@ export function VisualEditorContextProvider({
     );
   }, [entityStore, locale]);
 
+  // Either gets called when dataSource changed or designComponetsRegistry changed (manually)
+  const fetchMissingEntities = useCallback(async () => {
+    const entityLinks = [...Object.values(dataSource), ...designComponentsRegistry.values()];
+    const { missingAssetIds, missingEntryIds } = entityStore.getMissingEntityIds(entityLinks);
+    // Only continue and trigger rerendering when we need to fetch something and we're not fetching yet
+    if (!missingAssetIds.length && !missingEntryIds.length) return;
```
SofiaMargariti, line 123:
```suggestion
    if (!missingAssetIds.length && !missingEntryIds.length) {
      setInitialEntitiesFetched(true);
      return;
    }
```
I got a blank canvas when trying this out and the reason seemed to be that there were no missing entities but `initialEntitiesFetched` was still false. This seems to have fixed it. In my case there were no missing entities because they had been fetched in preview mode already.

Chaoste: Aaaaah, sorry I was too quick again on this one. Will test this carefully again as well. Thanks for showing me this fix!

SofiaMargariti (approved):

PR #218, opened 2023-12-22
fix: refetch when dataSource or design components changed [SPA-1711]

Chaoste:
With my previous PR, the SDK did only fetch once which caused rendering issues as nto all entities might have been requested yet.

Now, we're checking the entityStore and update it everytime the data source or design components change (via componentTreeUpdated)

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -121,7 +121,7 @@ export function VisualEditorContextProvider({
 
   // Either gets called when dataSource changed or designComponetsRegistry changed (manually)
   const fetchMissingEntities = useCallback(
-    async (newDataSource?: typeof dataSource) => {
+    async (newDataSource?: CompositionDataSource) => {
       const entityLinks = [
         ...Object.values(newDataSource ?? dataSource),
         ...designComponentsRegistry.values(),
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
@@ -38,7 +38,7 @@ type VisualEditorContextType = {
   unboundValues: CompositionUnboundValues;
   breakpoints: Breakpoint[];
   entityStore: EditorModeEntityStore;
-  areInitialEntitiesFetched: boolean;
+  areEntitiesFetched: boolean;
 };
 
 export const VisualEditorContext = React.createContext<VisualEditorContextType>({
@@ -53,7 +53,7 @@ export const VisualEditorContext = React.createContext<VisualEditorContextType>(
   locale: null,
   breakpoints: [],
   entityStore: {} as EditorModeEntityStore,
-  areInitialEntitiesFetched: false,
+  areEntitiesFetched: false,
 });
 
 type VisualEditorContextProviderProps = {
@@ -86,7 +86,10 @@ export function VisualEditorContextProvider({
   const [isDragging, setIsDragging] = useState(false);
   const selectedNodeId = useRef<string>('');
   const [locale, setLocale] = useState<string>(initialLocale);
-  const [areInitialEntitiesFetched, setInitialEntitiesFetched] = useState(false);
+  // Set to true when entities were fetched from the parent app.
+  // Reset to false when we receive a tree update and need to validate
+  // again whether all necessary entities are fetched.
+  const [areEntitiesFetched, setEntitiesFetched] = useState(false);
   const [isFetchingEntities, setFetchingEntities] = useState(false);
 
   const [entityStore, setEntityStore] = useState<EditorModeEntityStore>(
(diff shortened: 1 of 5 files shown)


packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -337,7 +352,7 @@ export function VisualEditorContextProvider({
     return () => {
       window.removeEventListener('message', onMessage);
     };
-  }, [fetchMissingEntities, entityStore, mode]);
+  }, [fetchMissingEntities, entityStore, mode, dataSource]);
```
loweisz, line 355: This might be dangerous to set the dataSource as dependency to the `useEffect` in which we `setDataSource` in multiple places. 
This could lead to infinite loops as we had quite recently

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -113,40 +116,47 @@ export function VisualEditorContextProvider({
...
-      console.error(error);
-    } finally {
-      // Important to set this as it is the only state variable that triggers a rerendering
-      // of the components (changes inside the entityStore are not part of the state)
-      setInitialEntitiesFetched(true);
-      setFetchingEntities(false);
-    }
-  }, [dataSource, entityStore]);
+  const fetchMissingEntities = useCallback(
+    async (newDataSource?: typeof dataSource) => {
```
loweisz, line 124: why not directly `CompositionDataSource` type?

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
```diff
@@ -86,7 +86,10 @@ export function VisualEditorContextProvider({
   const [isDragging, setIsDragging] = useState(false);
   const selectedNodeId = useRef<string>('');
   const [locale, setLocale] = useState<string>(initialLocale);
-  const [areInitialEntitiesFetched, setInitialEntitiesFetched] = useState(false);
+  // Set to true when entities were fetched from the parent app.
+  // Reset to false when we receive a tree update and need to validate
+  // again whether all necessary entities are fetched.
+  const [areEntitiesFetched, setEntitiesFetched] = useState(false);
   const [isFetchingEntities, setFetchingEntities] = useState(false);
```
loweisz, line 93: Maybe instead of two states we can use a status state that hold enums like `FETCHING` `ISFETCHED` etc?

loweisz (changes requested):
Some nits, but please check that we don't have another infinite loop with the `useEffect`

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: Good question. I was considering this as well. The reason why I didn't go for this was that it wouldn't have the same behavior or less readability IMO. Here are four different cases possible and all of them will happen throughout the lifetime of the component.
1. `areEntitiesFetched=false` and `isFetchingEntities=false` -> initial state; we can start fetching in the effect; don't render bindings or design components yet
2. `areEntitiesFetched=false` and `isFetchingEntities=true` -> we're fetching, so don't trigger another fetch call
3. `areEntitiesFetched=true` and `isFetchingEntities=false` -> render bindings & design components; when necessary trigger another fetch call
4. `areEntitiesFetched=true` and `isFetchingEntities=true` -> still allow rendering while refetching. This can happen when bindings or design components change.

So the enum would have four states. Giving them descriptive names might be a bit harder and it's probably not as readable as these explicit booleans. Note that there is no performance issue in having separate state variables and setting them one by one. Setting both booleans sequentially will still end in only one re-rendering.

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: Huh, dunno why I did this as the type is already imported and quite simple to write out. Will replace it 👍

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
Chaoste: Not writing out all dependencies in effects can also pose risks as the effect will eventually not pick up the latest value of a state variable.
Also, note that `fetchMissingEntities` depends on `dataSource`. So if `dataSource` changes, it will already update the callback and thus call again the effect. So practically, this shouldn't cause any difference in the actual execution.
Additionally, note in this screenshot that this effect itself is just adding and removing event listeners. So the state is only changed when an event was triggered, not when the effect was called.

Screenshot 2023-12-27 at 11 46 20

loweisz: That is true, I was just a bit cautious because something like that was the reason that we had infinite get requests. 
But you are right that the useCallback should have the same effect, so hopefully nothing has changed because of this

packages/experience-builder-sdk/src/blocks/editor/VisualEditorContext.tsx
loweisz: Ahh, I thought they would be exclusive so `isFetchingEntities` can never be true when `areEntitiesFetched ` is also true.

loweisz (approved):
Thanks for the clarifications, looks good so far :)

Chaoste:
Test links SPA-1711

PR #219, opened 2024-01-05
Release@next []

chasepoirier:


packages/core/src/fetchers/fetchExerpienceEntry.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+        client: undefined,
+        experienceTypeId: 'books',
+        locale: 'en-US',
+        identifier: { slug: 'slug' },
+      });
+    } catch (e) {
+      expect((e as Error).message).toBe(
+        'Failed to fetch experience entities. Required "client" parameter was not provided'
+      );
+    }
```
Chaoste, line 26: [request change]
This test won't function correctly. If the function doesn't throw an error, we don't call `expect` and thus don't make the test as failed. Also `expect` calls should never be conditionally but also have to be called. Please have a look at the docs of `toThrow` ([here](https://jestjs.io/docs/expect#tothrowerror)) and `rejects` ([here](https://jestjs.io/docs/expect#rejects)).

```suggesti
    await expect(async () => {
      await fetchExperienceEntry({
        // @ts-expect-error intentionally setting it to undefined
        client: undefined,
        experienceTypeId: 'books',
        locale: 'en-US',
        identifier: { slug: 'slug' },
      });
    }).rejects.toThrow('Failed to fetch experience entities. Required "client" parameter was not provided');
```

The same goes for the other broken tests in this file.

packages/core/src/fetchers/fetchReferencedEntities.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+        // @ts-expect-error intentionally setting it to undefined
+        client: undefined,
+        experienceEntry: compositionEntry,
+        locale: 'en-US',
+      });
+    } catch (e) {
+      expect((e as Error).message).toBe(
+        'Failed to fetch experience entities. Required "client" parameter was not provided'
+      );
+    }
```
Chaoste, line 25: [request change]
Same feedback as for the tets in `fetchExperienceEntry.spec.ts` that try/catch clauses are not working for jest tests.

packages/core/src/fetchers/fetchExerpienceEntry.spec.ts
Chaoste, line 1: [nitpick]
There is a typo in this file name. It should be `feetchExperienceEntry.spec.ts`.

packages/core/src/fetchers/fetchReferencedEntities.ts
```diff
@@ -0,0 +1,64 @@
+import { ExperienceEntry } from '@/types';
+import { ContentfulClientApi, Entry, Asset } from 'contentful';
+import { isExperienceEntry } from '@/utils';
+
+type fetchReferencedEntitiesArgs = {
```
Chaoste, line 5: [nitpick]
Types are always written with a capital first letter (at least at Contentful). This is especially nice to consider as this is a public repo and the types are consumed by SDK users.
```suggestion
type FetchReferencedEntitiesArgs = {
```

Chaoste: Same feedback for `fetchExperienceEntryArgs`, `fetchExperienceArgs`, etc.

packages/experience-builder-sdk/src/hooks/useExperienceBuilder.ts
```diff
@@ -23,7 +23,7 @@ type UseExperienceBuilderProps = {
 };
 
 /**
- * @deprecated This hook is deprecated. We re-designed the SDK to enable client or server side rendering
+ * @deprecated This hook is deprecated. Use fetchBySlug or fetchById instead
```
Chaoste, line 26: I guess this will be a major version bump. How about we drop this file completely as part of the new release?

packages/experience-builder-sdk/src/hooks/useFetchById.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+        id,
+        experienceTypeId,
+        localeCode,
+      },
+    });
+
+    await waitFor(() => {
+      expect(result.current).toBeDefined();
+      expect(result.current.isLoading).toBe(true);
+    });
```
Chaoste, line 43: [nitpick]
`waitFor` is not meant to be used with multiple assertions and can lead to unwanted issues:
```suggestion
    await waitFor(() => expect(result.current?.isLoading).toBe(true));
```
See here for more explanation on this: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback

packages/experience-builder-sdk/src/hooks/useFetchById.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      initialProps: {
+        client: clientMock,
+        mode: 'preview' as ExternalSDKMode,
+        id,
+        experienceTypeId,
+        localeCode,
+      },
+    });
+
+    await waitFor(() => {
```
Chaoste, line 57: [nitpick]
Same problem as with in the comment above. Especially this `waitFor` is a very brittle check that can easily break.

packages/experience-builder-sdk/src/hooks/useFetchBySlug.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      initialProps: {
+        client: clientMock,
+        mode: 'preview' as ExternalSDKMode,
+        slug,
+        experienceTypeId,
+        localeCode,
+      },
+    });
+
+    await waitFor(() => {
```
Chaoste, line 40: [nitpick]
Same feedback about invalid multiple assertions in `waitFor`

Chaoste: This is multiple times happening in this test file.

packages/experience-builder-sdk/src/hooks/useFetchExperience.test.ts
```diff
@@ -27,249 +26,223 @@ describe('useFetchExperience', () => {
...
-    await waitFor(() => {
-      expect(result.current.error?.message).toBe(
-        'Either slug or id must be provided to useFetchExperience'
+    try {
+      await act(() =>
+        // @ts-expect-error undefined is not alloed through types, but it can still happen if invoked from a plain js
+        res.result.current.fetchBySlug({ experienceTypeId, slug: undefined, localeCode })
       );
+    } catch (e) {
+      expect((e as Error).message).toBe(
```
Chaoste, line 220: [nitpick]
Same feedback as above that `try/catch` doesn't work in jest tests as intended. Use `toThrow` instead.

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
```diff
@@ -1,41 +1,21 @@
...
-export const useFetchExperience = ({
-  mode,
-  client,
-  slug,
-  id,
-  experienceTypeId,
-  localeCode,
-}: useClientSideExperienceFetchersProps) => {
+/**
+ * @deprecated please use `useFetchBySlug` or `useFetchById` hooks instead
```
Chaoste, line 16: I guess we can remove this one as well as part of the major release?

Chaoste (changes requested):

packages/visual-editor/src/components/DropZone/Dropzone.tsx
Chaoste, line 1: @[nitpick] Since dropzone is usually considered one word, the folder should also be written `Dropzone` with a lower case `z`. Not that git itself is not properly handing casings of file and folder names. The hack here will be to use a temp name as intermediate step:
`git rm DropZone/ tempName/ && git commit -am "style: rename dropzone temp" && git rm tempName/ Dropzone/ && git commit -am "style: rename temp to dropzone"`

Chaoste (approved):
Thanks for providing this PR, re-structuring our logic and merging all these improvements :)
I initially requested changes on a few nitpicks and not working unit tests but I guess it's better to merge this one first and then create additional PRs based on those as this is not blocking.

Note: at least we might wanna consider removing deprecated methods as part of this major release.

packages/core/src/fetchers/createExperience.spec.ts
```diff
@@ -1,7 +1,8 @@
 import type { Entry } from 'contentful';
-import { compositionEntry } from '../../test/__fixtures__/composition';
-import { assets, entries } from '../../test/__fixtures__/entities';
+import { compositionEntry } from '../test/__fixtures__/composition';
+import { assets, entries } from '../test/__fixtures__/entities';
 import { createExperience } from './createExperience';
+import { describe, it, expect } from 'vitest';
```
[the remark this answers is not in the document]
elylucas: The core package uses vitest, and by default, you import the vitest functions. There were a lot of issue trying to get jest to work with esmodules and they worked out of the box in vitest. The api is very similar to jest, so it shouldn't be much of a learning curve to use it here. I'd suggest updating the sdk package to vitest sometime in the future as well (or move to Cypress Component tests for the CT tests).

packages/core/src/fetchers/fetchExerpienceEntry.spec.ts
elylucas: These tests were brought over from the other project, so they would of had the issue there as well. I'll take a look at the suggestion and improve it.

packages/experience-builder-sdk/src/hooks/useFetchExperience.ts
elylucas: This will be the first release where it is deprecated. I'd usually wait at least one release to let users move over before it is removed, but since this is alpha its not as big of a deal. Might be good to chat about it as a whole.

packages/visual-editor/src/components/DropZone/Dropzone.tsx
Chaoste: Ah yes, I meant `mv` not `rm` ^^

PR #222, opened 2024-01-08
fix: design tokens in visual editor [ALT-200]

primeinteger:
Changes:
* Fixes resolving design token values in the visual editor
* Moves `sendMessage(OUTGOING_EVENTS.DesignTokens` to run when the SDK is connected (suggested by Ely) - instead of immediately when `defineDesignTokens({ ... })` is called by the user's app
* Changes the `DesignTokensDefinition` type to show tooltip hints about the expected keys/props for developers


packages/core/src/types.ts
```diff
@@ -272,7 +272,15 @@ export type Composition = {
   componentSettings?: ExperienceComponentSettings;
 };
 
-export type DesignTokensDefinition = { [key: string]: string | DesignTokensDefinition };
+export type DesignTokensDefinition = {
+  spacing?: Record<string, string>;
+  sizing?: Record<string, string>;
+  colors?: Record<string, string>;
+  borders?: Record<string, { size: string; color: string }>;
```
ryunsong-contentful, line 279: can you add one more property called `style: string`, also can you change `size` to `width`

packages/core/src/types.ts
```diff
@@ -272,7 +272,15 @@ export type Composition = {
...
+  fontSize?: Record<string, string>;
+} & {
+  [key: string]: string | Record<string, string>;
```
ryunsong-contentful, line 282: We may want to make this a recursive type so that there can be multiple nestings. Maybe we can make this change in the future so either adding a comment or actually making a recursive type here should suffice

ryunsong-contentful:
```
Moves sendMessage(OUTGOING_EVENTS.DesignTokens to run when the SDK is connected (suggested by Ely) - instead of immediately when defineDesignTokens({ ... }) is called by the user's app
```
Makes sense to me! Will approve once you make the type changes as they would currently break the user-interface stuff if this gets in

ryunsong-contentful (approved):
LGTM

PR #224, opened 2024-01-08
chore: update experience-builder-sdk readme for design tokens

ryunsong-contentful:


primeinteger (commented):
Since all this design token documentation was added to a Wiki page, it may be better to add a minimal example of defining design tokens in the existing setup example code block in this readme to make it a short introduction 🤔

PR #228, opened 2024-01-09
feat: add cfBorder designTokens resolver [ALT-208]

ryunsong-contentful:
Resolves values for borders

### DIFF
--- a/packages/core/src/registries/designTokenRegistry.ts
+++ b/packages/core/src/registries/designTokenRegistry.ts
@@ -13,12 +13,22 @@ export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition
 
 const templateStringRegex = /\${(.+?)}/g;
 
-export const getDesignTokenRegistration = (breakpointValue: string) => {
+export const getDesignTokenRegistration = (breakpointValue: string, variableName: string) => {
   if (!breakpointValue) return breakpointValue;
 
   let resolvedValue = '';
   for (const part of breakpointValue.split(' ')) {
-    const tokenValue = templateStringRegex.test(part) ? resolveSimpleDesignToken(part) : part;
+    let tokenValue = '';
+    if (variableName === 'cfBorder' && templateStringRegex.test(part)) {
+      tokenValue = resolveBorderDesignToken(part);
+      const { width, color, style } =
+        (tokenValue as unknown as DesignTokensDefinition['borders']) || {
+          width: '1px',
+          style: 'outside',
+          color: 'rgba(0,0,0,1)',
+        };
+      tokenValue = `${width} ${style} ${color}`;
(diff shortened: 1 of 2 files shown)


packages/core/src/registries/designTokenRegistry.ts
```diff
@@ -32,3 +42,11 @@ const resolveSimpleDesignToken = (templateString: string) => {
   const resolvedValue = spacingValues[designKeys[1]] as string;
   return resolvedValue || '0px';
 };
+
+const resolveBorderDesignToken = (templateString: string) => {
+  const nonTemplateValue = templateString.replace(templateStringRegex, '$1');
+  const designKeys = nonTemplateValue.split('.');
+  const designValues = designTokensRegistry[designKeys[0]] as DesignTokensDefinition;
+  const resolvedValue = designValues[designKeys[1]] as string;
```
primeinteger, line 50: I think `as string` should be changed to match the resolved value being the borders object from DesignTokenDefinition.

```suggestion
  const resolvedValue = designValues[designKeys[1]] as DesignTokensDefinition['borders'];
```

packages/core/src/registries/designTokenRegistry.ts
```diff
@@ -13,12 +13,22 @@ export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition
...
+    if (variableName === 'cfBorder' && templateStringRegex.test(part)) {
+      tokenValue = resolveBorderDesignToken(part);
+      const { width, color, style } =
+        (tokenValue as unknown as DesignTokensDefinition['borders']) || {
+          width: '1px',
+          style: 'outside',
+          color: 'rgba(0,0,0,1)',
+        };
+      tokenValue = `${width} ${style} ${color}`;
+    } else tokenValue = templateStringRegex.test(part) ? resolveSimpleDesignToken(part) : part;
```
primeinteger, line 31: I think the type casting can be avoided here.  It seems to stem from the return type being wrong for `resolveBorderDesignToken` (string instead of the border object)

Here's a possible way to rewrite this without any type casting (although I didn't test this)

```suggestion
    if (variableName === 'cfBorder' && templateStringRegex.test(part)) {
      const borderTokenValue = resolveBorderDesignToken(part);
      const { width, color, style } = borderTokenValue || {
        width: '1px',
        style: 'outside',
        color: 'rgba(0,0,0,1)',
      };
      resolvedValue += `${width} ${style} ${color} `;
      continue;
    }
    const tokenValue = templateStringRegex.test(part) ? resolveSimpleDesignToken(part) : part;
    resolvedValue += `${tokenValue} `;
```

primeinteger (changes requested):
Type casting 💔

PR #231, opened 2024-01-09
fix(visual-editor): remove error around passing editor props to normal elements [ALT-255]

elylucas:


primeinteger (approved):
Looking good! 🚀

PR #235, opened 2024-01-10
feat: add cfBorder designTokens resolver [ALT-208]

ryunsong-contentful:
Resolves values for borders


primeinteger (approved):
Super clean! 🚀

PR #240, opened 2024-01-10
fix(visual-editor): canvas interactions flickering on changes [ALT-241]

chasepoirier:
Flickering issues with canvas interactions were caused by always updating the entire component tree in the visual editor when it received the `CompositionUpdated` PostMessage from the user_interface.

This PR adds additional complexity to the visual editor so that instead of updating the entire component tree, which caused a rerender of the entire layout, it instead finds the specific differences between the current tree and the incoming updated tree from the user_interface.

Based on the diff, it then only updates the changed nodes in the tree.

Additional logic was also added to commit certain actions to the tree ahead of receiving the actual updated tree via PostMessage.

For example, if I drag a component to reorder them on the canvas, we commit that action to the visual editor's tree at the same time we send a PostMessage of the change to the user_interface. Once the `CompositionUpdated` PostMessage is received back, we check the diff again and if there is no diff, we prevent an update to the tree which prevents any additional unneeded rerenders.


package-lock.json
Chaoste, line 1: This is a rather big but probably unnecessary change to the lock file. I see that all integrity checks were removed. I assume that this is based on an old version of npm?
Would be great if you could undo this change with `git checkout -- origin/next package-lock.json` to keep the git history a bit clean :)

On the long run, I think we should fix the package manager to a single version.

packages/visual-editor/src/utils/getTreeDiff.ts
```diff
@@ -0,0 +1,195 @@
...
+interface CompareNodeParams {
+  currentNode?: CompositionComponentNode;
+  updatedNode?: CompositionComponentNode;
+  currentNodeCount: number;
+  updatedNodeCount: number;
+  originalTree: CompositionTree;
+  differences: Array<TreeDiff | null>;
+}
+
+function compareNodes({
```
Chaoste, line 89: This is looking great and I can't wait to see those improvements in action!

If I'm not mistaken this is currently not covering updated entries/ assets that linked via `dataSource`. In [this recent PR](https://github.com/contentful/user_interface/pull/19090/files#diff-4e4cdedb6684dd71349bf556a416fe99ac358a33e84c3809f5ffa447339bd48aR50), I added the logic that would call `treeUpdated` to cause a rerendering as `updatedEntity` somehow didn't cause a rerendering.

Can we check here also if the linked entities in the entity store changed somehow? Otherwise, we would need to find out how to rerender the correct node on `updatedEntity` which is currently not working.

Just want to call this out as it looks like this change will break the feature that I shipped in that linked PR.

Chaoste (commented):
Really great improvement! The code looks reasonable to me but I didn't spend yet time on testing and careful line-by-line review - so I didn't approve it yet but hopefully, someone else can do that on your team :)

package-lock.json
chasepoirier: Nice catch, just synced the package-lock with `next`

packages/visual-editor/src/utils/getTreeDiff.ts
chasepoirier: @Chaoste Gotcha I'll test this out today and see how we could get that logic covered in this PR.

Just to clarify, the use case that currently doesn't work would be:
- I bind let's say a "header" text component to another entity
- I change the content of that entity and publish it
- I should see the changed content immediately in my experience

packages/visual-editor/src/hooks/useCanvasInteractions.ts
```diff
@@ -0,0 +1,75 @@
...
+    addChild(destination.index, destination.droppableId, node);
+
+    onDrop({
+      data: tree,
+      componentType: draggableId,
+      destinationIndex: destination.index,
+      destinationZoneId: destination.droppableId,
+    });
+
+    return;
```
primeinteger, line 49: nit: this return doesn't do anything since it's the last line in the method.  should give the same result if it was <kbd>delete</kbd>'d

packages/visual-editor/src/hooks/useCanvasInteractions.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    onComponentMoved({
+      nodeId: draggableId.replace('draggable-', ''),
+      destinationIndex: destination.index,
+      destinationParentId: destination.droppableId,
+      sourceIndex: source.index,
+      sourceParentId: source.droppableId,
+    });
+
+    return;
```
primeinteger, line 71: <kbd>delete</kbd>

packages/visual-editor/src/types/treeActions.ts
```diff
@@ -0,0 +1,54 @@
+import { CompositionComponentNode } from '@contentful/experience-builder-core/types';
+
+export type TreeDiffType =
+  | 'remove_node'
+  | 'add_node'
+  | 'move_node'
+  | 'update_node'
+  | 'reorder_node'
+  | 'replace_node';
```
primeinteger, line 9: Could be nice to switch this type to an Enum since each string is used in a few places

packages/visual-editor/src/utils/treeHelpers.ts
```diff
@@ -0,0 +1,175 @@
...
+    // we iterated over the old children and didn't introduce a change.
+    // that means the child node is added to the end of the array.
+    addChildNode(
+      oldChildren.length,
+      node.data.id,
+      updatedChildren[updatedChildren.length - 1],
+      node
+    );
+  }
+  return;
```
primeinteger, line 78: <kbd>delete</kbd>

packages/visual-editor/src/utils/treeHelpers.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    // we only care about the first instance of an unequal node
+    if (changed) {
+      return;
+    }
+    changed = true;
+    removeChildNode(i, nodeId, node);
+  });
+
+  return;
```
primeinteger, line 108: <kbd>delete</kbd>

primeinteger (approved):
This looks great!  I can't wait to see the improvement this will make in the UI

packages/visual-editor/src/utils/getTreeDiff.ts
chasepoirier: @Chaoste just tested this scenario with both entries and assets:

- Linked text to a blog page entry, changed the content of the blog page, instantly saw the changes in the layout
- Linked an image to an existing asset, changed the file in the asset, instantly saw the changed file in the layout

No changes were required so this seems to function properly as is :)

The reason it works is a bit of a mystery to me. It does seem like the proper components are re-rendered but it's not because of a diff in the tree. I tried debugging a bit with [wdyr](https://github.com/welldone-software/why-did-you-render) but wasn't able to determine the reason. We may want to look into this more at some point

packages/visual-editor/src/types/treeActions.ts
chasepoirier: Yeah good call, went ahead and did this

PR #253, opened 2024-01-11
fix(visual-editor): enable dropzones on custom components

chasepoirier:


packages/test-app/src/Page.tsx
```diff
@@ -1,9 +1,33 @@
 import { useParams } from 'react-router-dom';
 import '@contentful/experience-builder-components/styles.css';
 import './styles.css';
-import { ExperienceRoot, useFetchBySlug } from '@contentful/experience-builder';
+import { ExperienceRoot, defineComponents, useFetchBySlug } from '@contentful/experience-builder';
 import { useContentfulClient } from './hooks/useContentfulClient';
 import { useContentfulConfig } from './hooks/useContentfulConfig';
+import Duplex from './components/Duplex';
```
elylucas, line 7: did you mean this change for the test-app to get in? We haven't really discussed it but my thoughts are the test-app would stay mostly unchanged unless its something for the app itself.

elylucas (approved):
had a question about the test app but besides that looks good

packages/test-app/src/Page.tsx
chasepoirier: Yeah I added it because we don't really have a way to test custom components from within the test-app. It's currently only the builtin components but I could remove it and or rename it to something that makes more sense

chasepoirier: @elylucas I tried refactoring this a bit to make more sense in a more generic use case as a test-app. Lmk your thoughts, I think eventually we'll want to add a basic DesignToken definition too so we can easily test design tokens.

elylucas (approved):

packages/visual-editor/src/hooks/useComponentProps.tsx
```diff
@@ -189,5 +189,9 @@ export const useComponentProps = ({
     ...omit(props, CF_STYLE_ATTRIBUTES, ['cfHyperlink', 'cfOpenInNewTab']),
   };
 
+  if (definition.children) {
```
Chaoste, line 192: Hey @chasepoirier did you check whether this affects assembly noides? We set `children: true` for them to render their children within an experience but they should not be able to accept new ones. So it's rather a fixed set of children. I suspect that we need to check here that the definition is not an assembly one.

chasepoirier: I believe Assembly nodes still function as expected after this PR. The reason being because I had modified the component used for DesignComponents which looks like this:

```ts
// Feel free to do any magic as regards variable definitions for design components
// Or if this isn't necessary by the time we figure that part out, we can bid this part farewell
export const DesignComponent: React.FC<DesignComponentProps> = (props) => {
  if (props.editorMode) {
    const { node } = props;

    return props.renderDropzone(node, {
      ['data-test-id']: 'contentful-container',
      ['data-cf-node-id']: node.data.id,
      ['data-cf-node-block-id']: node.data.blockId,
      ['data-cf-node-block-type']: node.type,
      id: 'design-component',
      className: props.className,
      style: designComponentStyle,
    });
  }
  // Using a display contents so design component content/children
  // can appear as if they are direct children of the div wrapper's parent
  return <div data-test-id="design-component" {...props} style={designComponentStyle} />;
};
```

Since the return is a `renderDropzone` instead of the `children` prop in editorMode the changes made don't have an effect

PR #256, opened 2024-01-12
refactor: rename design component to assembly [SPA-1730]

Chaoste:
As already called out last year before christmas, we are about to rename "design component" everywhere to "assembly". After I aligned with Chase regarding possible merge conflicts with hybrid editor, I decided to push this back until it's save without risking too many conflicts.

This PR replaces now **all** occurrences of the term "design component" in the whole mono repo.

The only exceptions are the constants for the tree and the messaging. We already have new ones and deprecated the old ones. Still, we will have to support both constants to have a smooth transition. I added a note that they will be dropped by version 5.

## Important step after merging
Similar to other major SDK changes in the past, we need to make sure that the user interface is using the latest SDK before the clients do.


packages/visual-editor/src/hooks/useComponent.tsx
```diff
@@ -7,10 +7,13 @@ import type {
 import { useMemo } from 'react';
 import { useComponentProps } from './useComponentProps';
 import { builtInComponents } from '@/types/constants';
-import { DESIGN_COMPONENT_NODE_TYPE } from '@contentful/experience-builder-core/constants';
-import { ContentfulContainer, DesignComponent } from '@contentful/experience-builder-components';
-import { resolveDesignComponent } from '@/utils/designComponentUtils';
-import { componentRegistry, createDesignComponentRegistration } from '@/store/registries';
+import {
+  DESIGN_COMPONENT_NODE_TYPE,
+  ASSEMBLY_NODE_TYPE,
```
SofiaMargariti, line 12: Has `ASSEMBLY_NODE_TYPE` been added in `@contentful/experience-builder-core`? Will this cause any errors if not on the latest version of the SDK?

Chaoste: Yes, you can see this in die diff at `/core/constants.ts` (all constants are exported from this file)

packages/visual-editor/src/utils/assemblyUtils.ts
```diff
@@ -65,24 +66,24 @@ export const deserializeDesignComponentNode = ({
...
       componentInstanceUnboundValues,
       componentInstanceDataSource,
     })
   );
 
   return {
-    // separate node type identifiers for design components and their blocks, so we can treat them differently in as much as we want
-    type: isDesignComponent ? DESIGN_COMPONENT_NODE_TYPE : DESIGN_COMPONENT_BLOCK_NODE_TYPE,
+    // separate node type identifiers for assemblies and their blocks, so we can treat them differently in as much as we want
+    type: isAssembly ? ASSEMBLY_NODE_TYPE : ASSEMBLY_BLOCK_NODE_TYPE,
```
SofiaMargariti, line 86: If using an older version of the SDK here the ASSEMBLY_NODE_TYPE will be undefined, which will cause issues later on, correct? Does this mean this is a breaking change and can only work on latest SDK?

SofiaMargariti (approved):
Looks good 🚀

PR #260, opened 2024-01-16
fix(experience-builder): only accept new assembly event [SPA-1730]

Chaoste:
We figured out that changing the payload of the iframe message can't just be addressed by a renaming with supporting old and new event type. The previous renaming leads to breaking changes for the SDK which we want to avoid.

For a smoother SDK migration time, we should instead use only the new event shape in the newer SDK and only support the older event shape in the older SDK (<=4.0.0-alpha.10).

To support both ways, the web app will send old and new events through the message channel until we drop the support for the deprecated functions.

TLDR; how the migration will work:
- Old SDK accepts only designComponentsAdded/ designComponentsRegistered (with param `designComponents`)
- New SDK accepts only assembliesAdded/ assembliesRegistered (with param `assemblies`)
- Experience Builder will send both events at once until we discontinue support for the old versions (starting with major version 5)

### DIFF
--- a/packages/visual-editor/src/hooks/useEditorSubscriber.ts
+++ b/packages/visual-editor/src/hooks/useEditorSubscriber.ts
@@ -182,6 +182,9 @@ export function useEditorSubscriber() {
           setLocale(locale);
           break;
         }
+        case INCOMING_EVENTS.DesignComponentsRegistered:
+          // Event was deprecated and support will be discontinued with version 5
+          break;
         case INCOMING_EVENTS.AssembliesRegistered: {
           const { assemblies }: { assemblies: ComponentRegistration['definition'][] } = payload;
 
@@ -193,6 +196,9 @@ export function useEditorSubscriber() {
           });
           break;
         }
+        case INCOMING_EVENTS.DesignComponentsAdded:
+          // Event was deprecated and support will be discontinued with version 5
+          break;
         case INCOMING_EVENTS.AssembliesAdded: {
           const {
             assembly,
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/hooks/useEditorSubscriber.ts
```diff
@@ -183,6 +183,8 @@ export function useEditorSubscriber() {
           break;
         }
         case INCOMING_EVENTS.DesignComponentsRegistered:
+          // Event was deprecated and support will be discontinued with version 5
+          break;
```
dimitrycf, line 187: Should we add explicit `console.warn(`event deprecatred ...yadayada`) ?  so that we can see if something is using it?

dimitrycf: I usually try to avoid swallowing errors as they eventually bite you where you don't expect.

But anyways this is merely a recommendation and feel free to act upon your own will :)

dimitrycf (approved):
LGTM, unblocking PR, but see the comment though ;)

packages/visual-editor/src/hooks/useEditorSubscriber.ts
Chaoste: Luckily, we already log each message in the console and only log additional warnings for unexpected ones. This one is expected to happen, so I would like to avoid polluting the logs with a separate log :)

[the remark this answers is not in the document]
Chaoste: The web app will send both for the time being. So I figured it would work nicely by ignoring the old event type for newer versions of the SDK while older SDK versions only "understand" the old event type.

PR #263, opened 2024-01-16
chore: use exact version for deps when resolving local path deps []

Chaoste:
In our colorful coin demo project, I observer the following issue:
We updated to the latest alpha version of the SDK but under the hood it resolves a deeper dependency from a month ago. This is since the range defined by the prefix `^` doesn't function as expected for alpha releases. In the example below it resolved `^0.0.2-alpha.15` to `0.0.2-dev-20240114T223125.0` which contains an outdated set of `INCOMING_EVENTS` and thus logged warnings and caused errors in the application.

Screenshot 2024-01-16 at 13 35 14

Screenshot 2024-01-16 at 13 38 19

https://github.com/lerna/lerna/tree/main/libs/commands/version#--exact

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -41,9 +41,9 @@ jobs:
           git config user.email "${{ github.actor}}@users.noreply.github.com"
 
           if [ ${{ github.ref_name }} = development ]; then
-            npx lerna publish prerelease --conventional-commits --conventional-prerelease --preid dev-$(date +%Y%m%dT%H%M%S) --no-changelog --no-git-tag-version --dist-tag dev --no-push --yes
+            npx lerna publish prerelease --exact --conventional-commits --conventional-prerelease --preid dev-$(date +%Y%m%dT%H%M%S) --no-changelog --no-git-tag-version --dist-tag dev --no-push --yes
           elif [ ${{ github.ref_name }} = next ]; then
-            npx lerna publish prerelease --conventional-commits --conventional-prerelease --preid alpha --yes --no-private --dist-tag latest
+            npx lerna publish prerelease --exact --conventional-commits --conventional-prerelease --preid alpha --yes --no-private --dist-tag latest
           else
             npx lerna publish --conventional-commits --conventional-graduate --yes --no-private --create-release github --dist-tag latest
           fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,8 @@ jobs:
       - vault/configure-lerna
       - run: echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > ~/.npmrc
       - run: yarn build # This should not be needed, but without it it will not contain any files
-      - run: yarn lerna version --no-private --conventional-commits --conventional-prerelease --preid next --create-release github --yes
+      # Use the exact dependency version for local path specifiers. e.g. this avoids resolving `^0.0.2-alpha.15` to `0.0.2-dev-20240114T223125.0`
+      - run: yarn lerna version --exact --no-private --conventional-commits --conventional-prerelease --preid next --create-release github --yes
       - run: yarn lerna publish from-git --pre-dist-tag next --yes
 
 workflows:


elylucas (changes requested):
A few questions:

1)  Could you explain the situation here? Locally, the packages in the monorepo reference each other via file paths, not versions ( see [here](https://github.com/contentful/experience-builder/blob/c92704621d3b7bacc658c12cd087e81d3e195519/packages/experience-builder-sdk/package.json#L28) for example). So I'm not sure as it would pull an errant version as it just uses the one in the monorepo.

2) --exact doesn't seem to be a valid option to lerna version/publish. I haven't tried myself but curious what it does or if it gets passed into another command.

3) We are using GH Actions now in the sdk repo, so we'll need to update the lerna commands in the action files [here](https://github.com/contentful/experience-builder/blob/development/.github/workflows/publish.yaml#L44) instead if we go forward with the change

Chaoste:
> A few questions:
> 
> 1. Could you explain the situation here? Locally, the packages in the monorepo reference each other via file paths, not versions ( see [here](https://github.com/contentful/experience-builder/blob/c92704621d3b7bacc658c12cd087e81d3e195519/packages/experience-builder-sdk/package.json#L28) for example). So I'm not sure as it would pull an errant version as it just uses the one in the monorepo.
> 2. --exact doesn't seem to be a valid option to lerna version/publish. I haven't tried myself but curious what it does or if it gets passed into another command.
> 3. We are using GH Actions now in the sdk repo, so we'll need to update the lerna commands in the action files [here](https://github.com/contentful/experience-builder/blob/development/.github/workflows/publish.yaml#L44) instead if we go forward with the change

Thanks @ely for the extensive feedback!

1. In the screenshot you can see that the [colorful coin demo](https://github.com/contentful/colorful-coin-eb) depends on `@contentful/experience-builder` which under the hood requires the `visual-editor@^0.0.2-alpha.15`. This version range is then resolved by npm to a temporary `dev` release of the package as we don't enforce the exact version. Especially for the alpha releases, we should make sure that all deps are on the latest version. In the `package.json`, I saw that we're using [local path specifiers](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#local-paths) which is resolved during `lerna version` so that later when someone installs one of the packages, they don't point to a non-existing local sibling package. So `file://...` will be replaced by a proper version for that. Unfortunately, the documentation by lerna is pretty sparse on this behavior but there is [this blog post](https://www.linkedin.com/pulse/things-i-wish-had-known-when-started-javascript-monorepo-gorej/) where the behavior is described like this:
> A word on releasing
> All local npm [local path specifiers](https://docs.npmjs.com/files/package.json#local-paths) are temporarily resolved to the > version range of the package they point to before being packed into the tarball.

2. I can see here in [the docs](https://github.com/lerna/lerna/tree/main/libs/commands/version#--exact) and in [this test](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/src/lib/version-command.spec.ts#L573) that the option `--exact` is supported and tested.

3. Thanks for the callout. Will update there as well!

elylucas (approved):

elylucas:
> > A few questions:
> > 
> > 1. Could you explain the situation here? Locally, the packages in the monorepo reference each other via file paths, not versions ( see [here](https://github.com/contentful/experience-builder/blob/c92704621d3b7bacc658c12cd087e81d3e195519/packages/experience-builder-sdk/package.json#L28) for example). So I'm not sure as it would pull an errant version as it just uses the one in the monorepo.
> > 2. --exact doesn't seem to be a valid option to lerna version/publish. I haven't tried myself but curious what it does or if it gets passed into another command.
> > 3. We are using GH Actions now in the sdk repo, so we'll need to update the lerna commands in the action files [here](https://github.com/contentful/experience-builder/blob/development/.github/workflows/publish.yaml#L44) instead if we go forward with the change
> 
> Thanks @ely for the extensive feedback!
> 
> 1. In the screenshot you can see that the [colorful coin demo](https://github.com/contentful/colorful-coin-eb) depends on `@contentful/experience-builder` which under the hood requires the `visual-editor@^0.0.2-alpha.15`. This version range is then resolved by npm to a temporary `dev` release of the package as we don't enforce the exact version. Especially for the alpha releases, we should make sure that all deps are on the latest version. In the `package.json`, I saw that we're using [local path specifiers](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#local-paths) which is resolved during `lerna version` so that later when someone installs one of the packages, they don't point to a non-existing local sibling package. So `file://...` will be replaced by a proper version for that. Unfortunately, the documentation by lerna is pretty sparse on this behavior but there is [this blog post](https://www.linkedin.com/pulse/things-i-wish-had-known-when-started-javascript-monorepo-gorej/) where the behavior is described like this:
> 
> > A word on releasing
> > All local npm [local path specifiers](https://docs.npmjs.com/files/package.json#local-paths) are temporarily resolved to the > version range of the package they point to before being packed into the tarball.
> 
> 2. I can see here in [the docs](https://github.com/lerna/lerna/tree/main/libs/commands/version#--exact) and in [this test](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/src/lib/version-command.spec.ts#L573) that the option `--exact` is supported and tested.
> 3. Thanks for the callout. Will update there as well!

Gotcha, thanks for the explanation. Also, sorry I completely missed the docs on exact.

.github/workflows/publish.yaml
```diff
@@ -41,9 +41,9 @@ jobs:
           git config user.email "${{ github.actor}}@users.noreply.github.com"
 
           if [ ${{ github.ref_name }} = development ]; then
-            npx lerna publish prerelease --conventional-commits --conventional-prerelease --preid dev-$(date +%Y%m%dT%H%M%S) --no-changelog --no-git-tag-version --dist-tag dev --no-push --yes
+            npx lerna publish prerelease --exact --conventional-commits --conventional-prerelease --preid dev-$(date +%Y%m%dT%H%M%S) --no-changelog --no-git-tag-version --dist-tag dev --no-push --yes
           elif [ ${{ github.ref_name }} = next ]; then
-            npx lerna publish prerelease --conventional-commits --conventional-prerelease --preid alpha --yes --no-private --dist-tag latest
+            npx lerna publish prerelease --exact --conventional-commits --conventional-prerelease --preid alpha --yes --no-private --dist-tag latest
           else
             npx lerna publish --conventional-commits --conventional-graduate --yes --no-private --create-release github --dist-tag latest
```
elylucas, line 48:
```suggestion
            npx lerna publish --exact --conventional-commits --conventional-graduate --yes --no-private --create-release github --dist-tag latest
```

elylucas: I think you'd want to do it on GA releases too, right?

Chaoste: I wasn't sure about that as exact versions in public libraries are generally not a good thing. I thought that at least for the dev/ alpha version is crucial for those to exactly specify all related deps. On the official release, I would try to relying on version ranges.
As described in the PR description, the original challenge was that it resolved `^0.0.2-alpha.15` to `0.0.2-dev-20240114T223125.0`. This means that even the latest alpha branch is eventually "lower" than any dev branch and thus leads to very outdated references. On the GA releases, I wouldn't expect something like this to happen. Not sure though, so let's keep an eye out on this and add it later if this really the case.

PR #265, opened 2024-01-17
Store & Rerender Live Updated Assemblies (aka "Assembly livesync") [SPA-1697]

dimitrycf:
## Purpose

 This feature is related to two repositories
 - user_interface ([PR](https://github.com/contentful/user_interface/pull/19187))
 - experience-builder ([PR](https://github.com/contentful/experience-builder/pull/265))

Keep in mind that to test this you'd need to build `experience-buidler-demo` with the dependencies from `experience-builder` from the [PR's branch](https://github.com/contentful/experience-builder/tree/fix/assembly-livesync-SPA1697) 

Ticket: https://contentful.atlassian.net/browse/SPA-1697

### How listening to updates of assemblies

There's a pubsub mechanism which allows to establish a connection with backend (I presume websockets) and receive events when entity is updated (in any way or form). 

The `useAssemblyUpdatePubsub()` hook inside of Experience Builder is listening to this event. Most of the work that this hook takes is to handle filtering and also bring this pubSubClient and it's events  into React land. (And making sure we do not have stale callbacks or stale .usedComponents array).

Keep in mind that we need to send a message to the SDK as well to inform it about new tree. However SDK's responsibility doesn't end up in merely receiving new {tree data structure} and rerendering canvas. 

I needed to extend protocol message 'compositionTreeUpdated' with additional key: `forceUpdateEntities: Link[]`. Anything included in this array would be force-updated (not taken from cache).

Also there was some {smart tree update logic based on diffs} which was introduced by @chasepoirier . I reused the same approach to selective node update to only update {assembly embed nodes}. (multiple instance).

It also needs to refetch those assemblies (they are cached in EntityStore (aka EditorEntityStore ( aka EditorModeEntityStore))). 

Hence EditorEntityStore and EditorModeEntityStore now has additional `skipCache: true` flag. 

### Deletion event

In Luiz's hook, there's mention for `DeletedEntry` and `DeletedAsset` which I interpreted as the fact that there's a pub-sub event to inform about deletion of an assembly. However, I wasn't able to trigger those events by causing deleting an assembly (eg. on flinkly). So I am not sure whether this some remnant of some non-existing behaviour in this hook.

However, in case there's a deleted assembly inside of an experience, then this error message is already shown (just putting here screenshot for reference).

image

### DIFF
--- a/packages/core/src/entity/EditorEntityStore.ts
+++ b/packages/core/src/entity/EditorEntityStore.ts
@@ -54,17 +54,32 @@ export class EditorEntityStore extends EntityStoreBase {
     return id.length === 1 ? id[0] : id.join(this.cacheIdSeperator);
   }
 
-  private async fetchEntity(type: 'Asset', ids: string[]): Promise<Array<Asset>>;
-  private async fetchEntity(type: 'Entry', ids: string[]): Promise<Array<Entry>>;
+  private async fetchEntity(
+    type: 'Asset',
+    ids: string[],
+    skipCache: boolean
+  ): Promise<Array<Asset>>;
+  private async fetchEntity(
+    type: 'Entry',
+    ids: string[],
+    skipCache: boolean
+  ): Promise<Array<Entry>>;
   private async fetchEntity(
     type: 'Asset' | 'Entry',
-    ids: string[]
+    ids: string[],
+    skipCache: boolean = false
   ): Promise<Array<Entry> | Array<Asset>> {
-    const { missing, resolved } = this.getEntitiesFromMap(type, ids);
+    let missing: string[];
+    if (!skipCache) {
+      const { missing: missingFromCache, resolved } = this.getEntitiesFromMap(type, ids);
+      if (missingFromCache.length === 0) {
+        // everything is already in cache
+        return resolved as Array<Entry> | Array<Asset>;
+      }
 
-    if (missing.length === 0) {
-      // everything is already in cache
-      return resolved as Array<Entry> | Array<Asset>;
+      missing = missingFromCache;
+    } else {
+      missing = [...ids];
     }
 
     const cacheId = this.getCacheId(missing);
@@ -130,9 +145,9 @@ export class EditorEntityStore extends EntityStoreBase {
     return this.getEntitiesFromMap(type, ids).resolved as Array<Entry> | Array<Asset>;
   }
 
-  public async fetchAsset(id: string): Promise<Asset | undefined> {
+  public async fetchAsset(id: string, skipCache: boolean = false): Promise<Asset | undefined> {
     try {
-      return (await this.fetchAssets([id]))[0];
+      return (await this.fetchAssets([id], skipCache))[0];
     } catch (err) {
       // TODO: move to debug utils once it is extracted
       console.warn(`Failed to request asset ${id}`);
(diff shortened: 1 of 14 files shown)


Chaoste (changes requested):
Please have a look at [the review](https://github.com/contentful/user_interface/pull/19187#pullrequestreview-1832425809) in the sibling PR. If I'm not mistaken, we don't need to introduce this special behavior in the SDK as we already have a separate message for that (`UPDATED_ENTITY`)

packages/core/src/entity/EditorEntityStore.ts
Chaoste, line 1: With the new approach we won't need the argument `skipCache` anymore. I would like to keep unused code out if that's fine by you :)

packages/visual-editor/src/hooks/useEditorSubscriber.ts
```diff
@@ -171,16 +175,19 @@ export function useEditorSubscriber() {
                 ...changedNode.data.unboundValues,
               });
             }
+
+            // Update the tree when all necessary data is fetched and ready for rendering.
```
Chaoste, line 179: Should we move this back? I'm not sure if there is still an intention in repeating this code here.

packages/visual-editor/src/store/tree.ts
```diff
@@ -46,6 +56,38 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
...
+            // Cannot use `structuredClone()` as node is probably a Proxy object with weird references
+            updateNode(node.data.id, cloneDeepAsPOJO(node), draftState.tree.root);
+          }
+        });
+      })
+    );
+  },
+
+  /**
+   * Force updates entire tree. Usually shouldn't be used as updateTree()
```
Chaoste, line 80: As this shouldn't be used in production, I would like to claim this clear at the front:
```suggestion
   * WARNING! Please note that this is for debugging purposes only:
   * Force updates entire tree. Usually shouldn't be used as updateTree()
```

packages/visual-editor/src/store/tree.ts
```diff
@@ -111,3 +156,7 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
     );
   },
 }));
+
+function cloneDeepAsPOJO(obj) {
```
Chaoste, line 160: As POJO might be a bit unclear to other devs, I would like to explain what the intention behind this helper is:
```suggestion
// Serialize and deserialize an object again to remove all functions and references.
// Some people refer to this as "Plain Old JavaScript Object" (POJO) as it solely contains plain data.
function cloneDeepAsPOJO(obj) {
```

packages/visual-editor/src/store/tree.ts
```diff
@@ -64,6 +106,9 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
 
     // The current and updated tree are the same, no tree update required.
     if (!treeDiff.length) {
+      console.debug(
+        `[exp-builder.visual-editor] updateTree(): During smart-diffing no diffs. Skipping tree update.`
```
Chaoste, line 110: If we want to keep this log, I would like to align it with the other logs in this repo:
```suggestion
        `[exp-builder.visual-editor::updateTree] During smart-diffing no diffs. Skipping tree update.`
```

packages/visual-editor/src/utils/treeHelpers.ts
```diff
@@ -7,6 +7,12 @@ export function updateNode(
   node: CompositionComponentNode
 ) {
   if (node.data.id === nodeId) {
+    console.log(`:::updateNode() updating node.data`, {
+      oldData: node.data,
+      newData: updatedNode.data,
+      isShallowEqual: node.data === updatedNode.data,
+      isDeepEqual: isEqual(node.data, updatedNode.data),
+    });
```
Chaoste, line 15: Do we want to keep this log? If so, I'd love to align this with the other logs as well
```suggestion
    console.debug(`[exp-builder.visual-editor::updateNode] updating node.data`, {
      oldData: node.data,
      newData: updatedNode.data,
      isShallowEqual: node.data === updatedNode.data,
      isDeepEqual: isEqual(node.data, updatedNode.data),
    });
```

packages/visual-editor/src/utils/treeTraversal.ts
Chaoste, line 1: It's good that we have this now in the monorepo available as well. Food for thought: We could move this to the core package as we probably could import this in the web app as well and drop the duplicate code over there.

Chaoste (approved):
I hove some concerns but it's all nitpicks. In general, this would be ready for merging if you decide to skip those.

PR #266, opened 2024-01-17
fix: replace node completely with new assembly node [SPA-1775]

Chaoste:
When creating a new assembly out of an existing container, the tree diff logic was making two mistakes:
- It detected not only the "replace" action but also the "remove" action which led to not rendering the new node at all
- The replace action itself did not replace the node fully which caused a weird mix of the original block node and new assembly node with warning logs and missing designs for the container inside the new assembly node.

I addressed the first problem by explicitly handling this case. The second one I could resolve by capturing the index of the node that should be replaced. I could then replace the node as a whole via its parent and the index.

## Screenshot of the adjusted tree diff action
Screenshot 2024-01-17 at 15 07 23


chasepoirier (approved):
Really nice job, this looks like a great fix

PR #270, opened 2024-01-17
fix: add dnd cypress stub [SPA-1712]

Spring3:
Adding a div with event handlers allowed to skip the "native" dnd and simulate events in cypress.

All that is required is setting the `window.Cypress` property.

You can see the progress on the video below. Sorry for the music - I forgot to disable mic when recording.

https://github.com/contentful/experience-builder/assets/4171202/67f92c5f-e2ce-4eaa-a9c9-06e3210b4c33


SofiaMargariti (approved):
👍

ChidinmaOrajiaku (approved):

PR #271, opened 2024-01-18
chore: move EntityStore into core and remove visual-sdk package

elylucas:


packages/core/src/entity/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum PostMessageMethods {
+  REQUEST_ENTITIES = 'REQUEST_ENTITIES',
+  REQUESTED_ENTITIES = 'REQUESTED_ENTITIES',
+}
```
primeinteger, line 4: Looks like this same enum is defined in `packages/core/src/constants.ts` as well

primeinteger (approved):
Looking good!  Found one small thing mentioned below

packages/core/src/entity/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum PostMessageMethods {
```
Chaoste, line 1: Not necessary to be done in this PR but just wanna call out that we should avoid enums as long as they are not necessary. Our [TS standards doc](https://contentful.atlassian.net/wiki/spaces/ENG/pages/4189094335/TypeScript+Standards#Enums) recommends using union types or object literals.

Chaoste (approved):
Great to see that we're dropping again the `visual-sdk` 👍🏻

PR #272, opened 2024-01-18
fix(core): remove lodash dep to get rid of weird reference errors

elylucas:
We were getting some weird lodash import errors in user interface when the core dep was referenced locally. We were using only two relatively simple methods from lodash (times and random), so I replaced those with custom ones and removed the package. This helped fix the error we were seeing.

### DIFF
--- a/packages/core/src/utils/utils.ts
+++ b/packages/core/src/utils/utils.ts
@@ -1,5 +1,3 @@
-import { times } from 'lodash-es';
-import { random } from 'lodash-es';
 import {
   CompositionTree,
   CompositionComponentNode,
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/utils.ts
```diff
@@ -121,6 +119,9 @@ export const generateRandomId = (letterCount: number): string => {
   const NUMS = '0123456789';
   const ALNUM = NUMS + LETTERS;
 
+  const times = (n: number, callback: () => string) => Array.from({ length: n }, callback);
+  const random = (min: number, max: number) => Math.floor(Math.random() * (max - min + 1)) + min;
```
ryunsong-contentful, line 123: Why do we calculate it like this?

ryunsong-contentful (approved):
LGTM - just a question on the rng for my own knowledge

packages/core/src/utils/utils.ts
elylucas: Not sure Im following, are you asking about the internal implementation of random? Its fairly typical for a js random function that i've seen, but if you have suggestions im up for it

PR #273, opened 2024-01-18
fix: adjust where the hover label goes for components and containers in xb

ryunsong-contentful:
Screenshot 2024-01-18 at 11 03 38 AM

Screenshot 2024-01-18 at 10 56 19 AM

### DIFF
--- a/packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
+++ b/packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
@@ -3,7 +3,8 @@ import { CSSProperties, ReactNode, SyntheticEvent } from 'react';
 import { Draggable } from '@hello-pangea/dnd';
 import classNames from 'classnames';
 import styles from './styles.module.css';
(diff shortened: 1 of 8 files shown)


packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
```diff
@@ -63,7 +65,9 @@ export const DraggableComponent = ({
           onClick={onClick}>
           {!isSelected ? (
             <div
-              className={classNames(styles.overlay, {
+              className={classNames({
+                [styles.overlayContainer]: isContainer,
+                [styles.overlay]: !isContainer,
```
primeinteger, line 70: I just had an idea: you could make the `.overlay` styles always present here (static), and apply the `.overlayContainer` styles conditionally based on your `isContainer` boolean.. Then the actual css within `.overlayContainer` only needs to be the styles you want to change/override from `.overlay` (in this case only the `top` style is different and all the other styles are duplicates (which could just be inherited instead of duplicated)

```suggestion
              className={classNames(styles.overlay, {
                [styles.overlayContainer]: isContainer,
```

PR #275, opened 2024-01-19
fix: adjust where the hover label goes for components and containers in xb

ryunsong-contentful:
Screenshot 2024-01-19 at 2 23 36 PM
Screenshot 2024-01-19 at 2 23 16 PM
Screenshot 2024-01-19 at 2 23 14 PM


packages/visual-editor/src/components/Draggable/canvasToolsUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+  CompositionUnboundValues,
+  CompositionVariableValueType,
+  Link,
+} from '@contentful/experience-builder-core/types';
+import { ASSEMBLY_DEFAULT_CATEGORY } from '@contentful/experience-builder-core/constants';
+
+export type ExperienceDataSource = CompositionDataSource;
+export type ExperienceUnboundValues = CompositionUnboundValues;
+export type ExperienceProperty = CompositionComponentPropValue;
+export type ExperiencePropertyLiteral = CompositionVariableValueType;
```
primeinteger, line 15: These type aliases seem unnecessary to me.  We could probably stick with the Composition-named types or rename them at the source if the names need to change to start with Experience.

packages/visual-editor/src/components/Draggable/canvasToolsUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+   * If for any reason, the element's top is negative, we show the tooltip at the bottom
+   */
+  if (coordinates.top < 0) {
+    newTooltipStyles['top'] = coordinates.height;
+  }
+
+  return newTooltipStyles;
+};
+
+export const getParentNodes = (treeNode: TreeNode, nodeId: string): TreeNode[] | null => {
```
primeinteger, line 148: This method doesn't seem to be used anywhere. Should it be removed?

chasepoirier (approved):
Lgtm 🚀 

More of a nice to have, but it would be good to get some unit tests in for the `canvasToolsUtils`

packages/visual-editor/src/components/Draggable/canvasToolsUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+  Link,
+} from '@contentful/experience-builder-core/types';
+import { ASSEMBLY_DEFAULT_CATEGORY } from '@contentful/experience-builder-core/constants';
+
+export type ExperienceDataSource = CompositionDataSource;
+export type ExperienceUnboundValues = CompositionUnboundValues;
+export type ExperienceProperty = CompositionComponentPropValue;
+export type ExperiencePropertyLiteral = CompositionVariableValueType;
+
+export type TreeNode = {
```
primeinteger, line 17: The `TreeNode` type only seems to be used by `getParentNodes()` which isn't used anywhere from what I can tell.  Maybe all of this can be trimmed out?

primeinteger (changes requested):
Would you consider trimming out the unused code from the the new `canvasToolsUtils.ts` file?  It looks like there's an unused method and some types that can be eliminated.

primeinteger (approved):
Clean AF ❤️

PR #278, opened 2024-01-22
fix(visual-sdk): only show bg color of dropzone for components being dragged in

elylucas:


packages/visual-editor/src/store/draggedItem.ts
```diff
@@ -1,13 +1,13 @@
 import { DragStart, DragUpdate } from '@hello-pangea/dnd';
 import { create } from 'zustand';
 
-export type DraggedItem = (DragStart & Partial<DragUpdate>) | undefined;
+export type DraggedItem = DragStart & Partial<DragUpdate>;
```
elylucas, line 4: making the DraggedItem type a bit more reusable elsewhere

packages/visual-editor/src/hooks/usePlaceholderStyle.ts
```diff
@@ -136,15 +136,21 @@ export const usePlaceholderStyle = () => {
         clientX = offsetLeft - translateValue;
       }
 
+      const draggingNewComponent = !!draggedItem.source.droppableId.startsWith('component-list');
+      const shouldShowBackgroundColor = draggingNewComponent;
```
primeinteger, line 140: small nit pick: `draggingNewComponent` var could be renamed to avoid the alias definition
```suggestion
      const shouldShowBackgroundColor = !!draggedItem.source.droppableId.startsWith('component-list');
```

primeinteger (approved):
Nice update for dropzone styles!

PR #279, opened 2024-01-23
feat: section component [ALT-205]

primeinteger:
* Adds a built-in Section component (full-width component using the same code as the Container component)
* Changes the Container component definition to include a default max-width of 1192px
* Adds a `rootComponent?: boolean` prop to the ComponentDefinition type, which is used by the UI to determine which components can be dropped into the canvas root

Here's a short demo video (recorded before I changed the container default max-width from 960px to 1192px)

https://github.com/contentful/experience-builder/assets/8539634/92663997-be80-410e-921e-4b3b50639055


chasepoirier (approved):
Nice work! Code looks great, love the additional clean up you always include into your PRs.

One question, how is the behavior when dropping a container on the root instead of inside a section? Does it center the container on the page without the section still?

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
```diff
@@ -6,45 +6,32 @@ import { ContentfulContainerAsHyperlink } from './ContentfulContainerAsHyperlink
 import type { ContentfulContainerAsHyperlinkProps } from './ContentfulContainerAsHyperlink';
 import { combineClasses } from '../../utils/combineClasses';
 
-export const ContentfulContainer: React.FC<ContentfulContainerAsHyperlinkProps> = (
-  sectionProps
-) => {
-  const { className, editorMode, children } = sectionProps;
+export const ContentfulContainer: React.FC<ContentfulContainerAsHyperlinkProps> = (props) => {
+  const { className, editorMode, children, cfHyperlink } = props;
```
elylucas, line 10: slight nitpick, but I generally like to destructure the props in the function params. Its slightly cleaner (imo) and you don't have an additional props param that might cause some confusion later in scope of the component.

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
```diff
@@ -6,45 +6,32 @@ import { ContentfulContainerAsHyperlink } from './ContentfulContainerAsHyperlink
...
-        className={className}
-        editorMode={editorMode}
-        cfHyperlink={sectionProps.cfHyperlink}
-        cfOpenInNewTab={sectionProps.cfOpenInNewTab}
-        {...sectionProps}>
-        {children}
-      </ContentfulContainerAsHyperlink>
-    );
+  if (cfHyperlink) {
+    return <ContentfulContainerAsHyperlink {...props}>{children}</ContentfulContainerAsHyperlink>;
```
elylucas, line 13: this could lead to unintentionally spreading some props onto the component that don't belong, which could cause additional rendering and confusion. We've also ran into a few React errors/warnings about props being spread onto elements that don't accept them. Unless we are sure the child component takes the same props as the parent I would suggest manually adding each variable.

packages/components/src/components/ContentfulContainer/ContentfulContainer.css
```diff
@@ -1,10 +1,10 @@
-#ContentfulContainer.defaultStyles {
+.ContentfulContainer {
   position: relative;
   display: flex;
   box-sizing: border-box;
 }
 
-#ContentfulContainer.defaultStyles::-webkit-scrollbar {
+.ContentfulContainer::-webkit-scrollbar {
```
elylucas, line 7: nitpick, but looks like we have a kebab casing going on for most css classes, so we might want to make this 'contentful-container'

packages/components/src/components/ContentfulContainer/ContentfulContainerAsHyperlink.tsx
```diff
@@ -84,7 +80,7 @@ export const ContentfulContainerAsHyperlink: React.FC<ContentfulContainerAsHyper
   // return (
```
elylucas, line 80: some commented out code that can be removed

packages/core/src/constants.ts
```diff
@@ -61,9 +61,9 @@ export const VISUAL_EDITOR_EVENTS = {
 };
 
 export const VISUAL_EDITOR_CONTAINER_ID = 'cf-visual-editor';
-export const CONTENTFUL_SECTION_ID = 'contentful-section'; // TODO: remove me once all customers are using 2023-09-28 schema version
+export const CONTENTFUL_SECTION_ID = 'contentful-section';
 export const CONTENTFUL_CONTAINER_ID = 'contentful-container';
-export const CONTENTFUL_SECTION_NAME = 'Section'; // TODO: remove me once all customers are using 2023-09-28 schema version
+export const CONTENTFUL_SECTION_NAME = 'Section';
```
elylucas, line 66: might want to be consistent on the  namings here (kebab case vs proper).

packages/core/src/definitions/components.ts
```diff
@@ -1,12 +1,21 @@
 import type { ComponentDefinition } from '@/types';
-
-import { CONTENTFUL_COMPONENT_CATEGORY, CONTENTFUL_CONTAINER_ID } from '@/constants';
-import { containerBuiltInStyles } from './styles';
+import * as Constants from '@/constants';
```
elylucas, line 2: probably not a big deal here, but import * can cause havok on tree shaking. I kinda like being able to see whats being used in a file as well by glancing at the imports. If we are running into issues where the imports are exploding (i know they are in some places, we can look into some refactoring of the constants to help).

packages/visual-editor/src/hooks/useComponentProps.tsx
```diff
@@ -182,21 +190,17 @@ export const useComponentProps = ({
...
     className,
     editorMode: true,
     node,
     renderDropzone,
     'data-cf-node-id': node.data.id,
     'data-cf-node-block-id': node.data.blockId,
     'data-cf-node-block-type': node.type,
-    // TODO: do we really need lodash just for this?
     ...omit(props, CF_STYLE_ATTRIBUTES, ['cfHyperlink', 'cfOpenInNewTab']),
+    ...(definition.children ? { children: renderDropzone(node) } : {}),
```
elylucas, line 202: nice! I bet this cleans up some of the TS wokiness caused downstream from here.

packages/core/src/utils/components.ts
```diff
@@ -0,0 +1,4 @@
+import { CONTENTFUL_CONTAINER_ID, CONTENTFUL_SECTION_ID } from '../constants';
+
+export const isContentfulStructureComponent = (componentId?: string) =>
```
elylucas, line 3: nice method to clean up some of the logic elsewhere

elylucas (approved):
Great job! I left some code styling nitpick comments, but take them or leave them.

primeinteger:
> One question, how is the behavior when dropping a container on the root instead of inside a section? Does it center the container on the page without the section still?

Currently when you drop a container into the root it is left aligned.  If you drop a Container into a Section, the Container is centered by default, and the Alignment options on the Section can be used to alter the alignment of the Container.

This one one of the things I'm about to meet with Travis on to discuss solutions.  I'm thinking we may way to auto-wrap Containers in a Section when they are dropped on the root, but probably not in other scenarios.

primeinteger:
> Great job! I left some code styling nitpick comments, but take them or leave them.

Thanks for the feedback!  I'll try to incorporate many of your suggestions before merging this

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
primeinteger: This component was already spreading via `{...sectionProps}` before I refactored.  I think I'll switch these props to be individually passed in instead of spreading though, you raised some good points about spreading here

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
primeinteger: The way the `ContentfulContainerAsHyperlinkProps` is defined conditionally based on `editorMode` makes it difficult to move this destructuring to the function params.  I think I'll leave this as-is

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
primeinteger: I think I'm going to leave this one as-is and make a story for future refactoring the ContentfulContainer component code.

Spreading the props seems to work better with the way the `ContentfulContainerAsHyperlinkProps` type is currently defined conditionally based on `editorMode`.  I attempted to refactor this and pass in all the props explicitly without spreading but it becomes challenging with the conditional prop types, and there's tradeoffs to consider w/ refactoring the conditional prop type

primeinteger: Created https://contentful.atlassian.net/browse/ALT-331 to capture some ideas for additional refactoring. Feel free to update this ticket with any other ideas!

PR #280, opened 2024-01-23
style: move all assembly checks to core utils []

Chaoste:
Enrich the core and SDK module by putting `checkIsAssemblyNode` and `checkIsAssemblyDefinition` into the utils. Also deprecated `checkIsAssembly` as this naming will be confusing when talking about the different entities around assemblies.


elylucas (commented):
🤙

elylucas (approved):

PR #282, opened 2024-01-23
Fix/assembly livesync spa1697 rebased 20240123hh16mm00

dimitrycf:


Spring3:
@dimitrycf should we close this PR?

PR #284, opened 2024-01-24
fix: allows re-parenting of components, however does not allow for new containers to be created while dragging across parents

ryunsong-contentful:
Main light bulb idea was to treat existing components on the canvas more like dragging components from the component-list.

This PR allows all components to be dragged across parents however it lacks one feature (which might even be debatable if we want this or not).

The feature it lacks is the ability to create new containers as you're dragging the component.


primeinteger (approved):
Awesome! It's nice this turned out to be fairly small adjustment 🎉

PR #285, opened 2024-01-25
feat: implement columns component with presets and custom options [ALT-43]

chasepoirier:
This PR implements the columns component with both presets and custom options enabled.

Related user_interface PR: https://github.com/contentful/user_interface/pull/19310

Follow up work is needed to enable gap, wrap, and drag to resize columns. Tickets are groomed and in Team Altitude's backlog to complete.

## Using the columns component
https://github.com/contentful/experience-builder/assets/23040747/6ff6ed23-8309-4613-b272-680c7a31045d

## Delivery mode of completed layout
Screenshot 2024-01-24 at 4 56 46 PM

PR #286, opened 2024-01-25
feat: Container default margin auto [ALT-326]

primeinteger:
Updating the Container definition to have default margin value `0 auto 0 auto`

This includes some changes from #279 for the `useComponentProps` hook:
* Some of the component's generated `cfStyles` are separated into a `editorWrapperClass`
* The `editorWrapperClass` is added to the `<DraggableComponent />`'s outermost div so the Draggable element has the correct size when nesting smaller containers which don't take up the full space inside of a larger container.

Containers that aren't full-width can be aligned different ways using margin auto:

Screenshot 2024-01-24 at 10 00 31 PM
Screenshot 2024-01-24 at 10 00 44 PM
Screenshot 2024-01-24 at 10 01 12 PM


packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
```diff
@@ -65,15 +65,15 @@ export const DraggableComponent = ({
   }, [coordinates]);
 
   return (
-    <div ref={draggableRef}>
+    <div className={classNames('Draggable', className)} ref={draggableRef}>
```
chasepoirier, line 68: In #285 I actually changed up this logic a bit so that we could get rid of this wrapping div that was causing layout issues in editor mode.

Since this `div` was only introduced for the Tooltip coordinates, I opted to add a data attribute to the `div` on line 76 here which can be used to fetch and calculate the same logic that this ref was doing before.

We may want to compare this solution to this PR to see if it would work for this scenario too

chasepoirier (changes requested):

packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
chasepoirier: Sounds good, PR is merged so you can move forward with this one!

primeinteger: Updated & reconciled, everything works great w/ tooltip placement and selected component outlines

PR #287, opened 2024-01-25
chore(visual-editor): resolve build warnings [SPA-1497]

Chaoste:
## Change 1:
I added a description of how to set up the `test-app` to make the transition easier for others in the Sparks team.

## Change 2: Typescript Warning ContentfulContainer
Screenshot 2024-01-25 at 13 12 28
To resolve this quickly, I just added a casting as the `props` types are not sufficient and it would be hard to extract the right case.

## Change 3: Remove circular dependency
Screenshot 2024-01-25 at 12 06 25
While this was not blocking in any way, I would rather not have such code patterns in our code.

At first, I tried the "[internal module pattern](https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de)" (a bit similar to `index.ts` but only meant for the internal usage) but it was still complaining now about dep chains like `Dropzone.tsx > internal.ts > Dropzone.tsx`.

Instead, I tried another pattern where the upper layer hands over the necessary logic/ functionality to its children. Here I resolved this by creating the function `renderDropzone` inside `Dropzone.tsx` and then passing it through to its children. Luckily, this resolved the warning.

## Addition
While being at it, I moved the hooks `useComponent` and `useComponentProps` to the place where its used, so inside `components/Dropzone`. I think we should watch out for keeping related code in one place instead of splitting them up by their type (`hook`/ `components`) as in the long run, this leads to duplicate folder hierarchy. For similar reasons, tests live next to the code they're testing.

I also replaced a default export with a named export as I think default exports are a bad pattern. Occurrences of the same component might call it differently every time which for example, makes it hard to find all places where a certain component is rendered.


SofiaMargariti (commented):
The Readme changes look good to me, but for the refactoring I would wait to get Altitude's input as well.
About the `packages/visual-editor/src/hooks/useComponentProps.tsx` -> `packages/visual-editor/src/components/Dropzone/useComponentProps.tsx` move,  I understand this is not a pattern we use in the web app but it is a valid way to organise the code and with this move we're creating an inconsistency where we have two ways of organising hooks.
I am also not a fan of having duplicate folder structures (in the future), but it's important to keep things consistent. Maybe we can have a discussion if we want to move the rest of the hooks around before starting moving just a single one.

packages/test-app/README.md
```diff
@@ -2,6 +2,17 @@
 
 This apps intended purpose is to quickly test the functionality of the EB when developing locally. There is also a deployed instance of the app that can be used to test EB in a production.
 
+## How to Start
+
+1. Create a `.env` in this package with the environment variables described in chapter [Config](#config).
+2. Make sure that the preview URL of the experience type is set to `http://localhost:5173/{entry.fields.slug}`.
```
elylucas, line 8: Step 2 isn't necessary, the visual editor now uses the experience selected in the editor instead of relying on the app to fetch it.

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
```diff
@@ -125,6 +128,22 @@ export function Dropzone({
     node?.data.blockId
   );
 
+  // To avoid a circular dependency, we create the recursive rendering function here and trickle it down
+  const renderDropzone = useCallback(
```
elylucas, line 132: could this be moved to its own hook to keep DropZone more cohesive?

packages/visual-editor/src/components/Dropzone/Dropzone.types.ts
```diff
@@ -0,0 +1,6 @@
+import { CompositionComponentNode } from '@contentful/experience-builder-core/types';
+
+export type RenderDropzoneFunction = (
```
elylucas, line 3: If the `renderDropZone` was in its own hook, this could be added to that hook to keep the type and the definition together.

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
```diff
@@ -125,6 +128,22 @@ export function Dropzone({
...
+    (node: CompositionComponentNode, props?: Record<string, unknown>) => {
```
elylucas, line 133: Should explicitly type this to the `RenderDropZone` function instead of relying on duck typing.

elylucas (approved):
Nice changes! I left a few comments that were more nitpicks than anything. The only one I would consider more is moving the renderDropZone function into its own file.

packages/visual-editor/src/components/Draggable/DraggableComponent.tsx
```diff
@@ -38,7 +38,7 @@ export const DraggableComponent = ({
   onMouseUp?: (e: SyntheticEvent) => void;
   onMouseOver?: (e: SyntheticEvent) => void;
   onMouseOut?: (e: SyntheticEvent) => void;
-  coordinates: Rect;
+  coordinates: Rect | null;
```
primeinteger, line 41: I noticed there's a couple more components like Tooltip and DraggableChildComponent that define this coordinates prop as `Rect` instead of `Rect | null` - if you change those instances it will prevent the same build warnings in those components as well.

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
primeinteger, line 132: ! solution to avoid the circular dependency warning!

packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -63,11 +69,12 @@ export const useComponent = ({ node: rawNode, resolveDesignValue }: ComponentPar
     node,
     areEntitiesFetched,
     resolveDesignValue,
+    renderDropzone,
     definition: componentRegistration.definition,
   });
 
   // Only pass editor props to built-in components
-  const { editorMode, renderDropzone, ...componentProps } = props;
+  const { editorMode, renderDropzone: _renderDropzone, ...componentProps } = props;
```
primeinteger, line 77: I think `renderDropzone` is destructured from props here so all the remaining props that we want to keep are spread into the `componentProps` variable (which does get used).  This change confused me at first because I wasn't seeing anywhere that `_renderDropzone` gets used.  Do you think this change is necessary?

primeinteger (approved):
Very nice cleanup effort! 🎉

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
Chaoste: If we have a separate file the renders the `Dropzone` we get again a circular dependency. I wasn't able to get rid of this warning than doing it in the same file 🤷‍♂️

elylucas: Ahh, I see. The `renderDropZone` function uses a `DropZone` itself, causing a circular dep between the two. Oh well, it can't all be perfect 😹 .

packages/test-app/README.md
Chaoste: You mean it's not necessary to add the slug, right? Still we need to point to `localhost:5173` :)

Chaoste: I guess it will still be necessary to render test-app in preview mode.

packages/visual-editor/src/components/Dropzone/useComponent.ts
Chaoste: Yes, it's the same for `editorMode`. We don't want to pass them to all kind of components as this creates warnings in the console IIRC.

Chaoste: Problem being here that we define `renderDropzone` already before, so I had to rename it :/

primeinteger: Got it, that makes sense now

PR #288, opened 2024-01-25
fix: duplicate assembly rendering [SPA-1783]

Chaoste:
When detaching an assembly node, the tree-diffing logic did not recognize that this was technically a "replacement". Based on the total node count in the tree it always figured that a node was "added" and thus didn't remove the previous assembly node but just added the new detached container.

I tested multiple ways of adding, removing, re-ordering, and replacing nodes to ensure that this had no side-effects on other rendering scenarios.


primeinteger (approved):
Nice fix and simplification!

PR #290, opened 2024-01-25
chore: add catalog info for backstage integration [SPA-1497]

Chaoste:
To show this mono repo as well as the SDK in Roadie, we need to provide catalog info configurations. Those are by no mean perfect. I looked at similar repositories to define at least the minimal required attributes.

### DIFF
--- a/catalog-info.yml
+++ b/catalog-info.yml
@@ -1,0 +1,24 @@
+apiVersion: backstage.io/v1alpha1
+kind: System
+metadata:
+  name: experience-builder
+  description: A monorepo for all NPM packages related to Experience Builder.
+  annotations:
+    circleci.com/project-slug: github/contentful/experience-builder
+    github.com/project-slug: contentful/experience-builder
+    contentful.com/ci-alert-slack: prd-experience-builder-alerts
+    backstage.io/source-location: url:https://github.com/contentful/experience-builder/
+spec:
+  type: library
+  lifecycle: production
+  owner: group:team-sparks
+
+---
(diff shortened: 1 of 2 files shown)


catalog-info.yml
```diff
@@ -0,0 +1,24 @@
...
+  description: A monorepo for all NPM packages related to Experience Builder.
+  annotations:
+    circleci.com/project-slug: github/contentful/experience-builder
+    github.com/project-slug: contentful/experience-builder
+    contentful.com/ci-alert-slack: prd-experience-builder-alerts
+    backstage.io/source-location: url:https://github.com/contentful/experience-builder/
+spec:
+  type: library
+  lifecycle: production
+  owner: group:team-sparks
```
anwaar931, line 14: Can we add multiple teams here? `team-altitude`

anwaar931 (approved):

catalog-info.yml
Chaoste: Unfortunately, Backstage wants to have one ultimate owner in here 😅
https://backstage.io/docs/features/software-catalog/descriptor-format/#specowner-required
> In Backstage, the owner of a component is the singular entity (commonly a team) that bears ultimate responsibility for the component, and has the authority and capability to develop and maintain it. They will be the point of contact if something goes wrong, or if features are to be requested.

chasepoirier: @Chaoste good idea on creating a shared experience-builder chapter, I think we should pursue doing that

chasepoirier (approved):

PR #294, opened 2024-01-26
fix: update hover labels to be more like onclick labels

ryunsong-contentful:
## Purpose
Make the hover labels match more closely with the active labels in terms of height and width. Kept the gray500 color because the color scheming will be handled during the dark mode ticket.

Screenshot 2024-01-26 at 8 52 59 AM
Screenshot 2024-01-26 at 8 52 53 AM


Chaoste (approved):
Nice one!

elylucas (approved):
🎉

PR #295, opened 2024-01-26
feat: update editor canvas background so that it's clear to users where the canvas lies [ALT-292]

ryunsong-contentful:
## Purpose
EB PR [https://github.com/contentful/experience-builder/pull/295]
Two parts (User Interface + EB changes) - this is because we need the entire IFrame (which is defined in User Interface) to be background transparent. Then in EB we make the relevant div with id "root" background white which allows us to show the user what part of the canvas is being used or not.
https://www.loom.com/share/c068512d3b8447dd9ca20ffc23dfc06d?sid=576ded82-c454-47f0-911b-9d0eef565055

Images below show the background for desktop, tablet, and mobile
Desktop
Screenshot 2024-01-26 at 9 17 36 AM
Tablet
Screenshot 2024-01-26 at 9 17 40 AM
Mobile
Screenshot 2024-01-26 at 9 17 44 AM

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -68,7 +68,7 @@ export const builtInStyles: Partial<
     type: 'Text',
     group: 'style',
     description: 'The background color of the section',
-    defaultValue: 'rgba(255, 255, 255, 0)',
+    defaultValue: 'rgba(255, 255, 255, 1)',
   },
   cfWidth: {
     displayName: 'Width',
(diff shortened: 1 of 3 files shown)


Chaoste (commented):
Worth mentioning that this only addresses our internal test-app and doesn't resolve the confusion for our customers

ryunsong-contentful:
> Worth mentioning that this only addresses our internal test-app and doesn't resolve the confusion for our customers

Yup, I'm wondering what is best here - I'm thinking we have need a wrapper around the experience that gives it the background white. Basically something in between the iFrame and the app that the developer defines.

PR #296, opened 2024-01-29
fix: update Heading component built in styles [ALT-154]

primeinteger:
* Enables more built in style options in the design sidebar for the default Heading component.
* Updates the `useComponentProps` hook:
  * Outputs `wrapperProps` object containing the `wrapperClass` and `data-cf-node` attributes
  * Adjusts the component styles w/ `margin: 0; width: 100%; height: 100%; maxWidth: none` after moving these styles from the component over to the editor wrapper styles.

Screenshot 2024-01-29 at 3 10 46 PM


packages/components/src/components/Heading/index.ts
```diff
@@ -7,14 +7,52 @@ export const HeadingComponentDefinition: ComponentDefinition = {
   id: 'heading',
   name: 'Heading',
   category: 'Contentful',
-  builtInStyles: ['cfMargin', 'cfPadding'],
+  builtInStyles: [
+    'cfMargin',
+    'cfPadding',
+    'cfFontWeight',
```
ryunsong-contentful, line 13: Are font size, line height, and bold not valid options because they get overridden no matter what you picked due to the default values in the variables section?

primeinteger: If those 3 options are added to `builtInStyles` array, the objects in `variables` will be ignored (so it will get the standard default values instead of custom ones.

ryunsong-contentful (approved):
Any thoughts on double checking the attributes with the team during standup tomorrow about the attributes we're allowing/not allowing to change?

chasepoirier (commented):
I think we should think about applying the following props to the wrapper draggable component instead of the underlying component:

```ts
    'data-cf-node-id': node.data.id,
    'data-cf-node-block-id': node.data.blockId,
    'data-cf-node-block-type': node.type,
```
This will also help with implementation of underlying components as they won't require `spreading` props on the component, instead we can handle that set up for users with the wrapping draggable component.

Instead of this return in `useComponentProps`:

```ts
  return { props: componentProps, editorWrapperClass };
```

we could instead do this:
```ts
  return { componentProps, wrapperProps }
```

The `wrapperProps` object could then be spread onto the draggable component div

primeinteger:
> I think we should think about applying the following props to the wrapper draggable component instead of the underlying component

I pushed updates to split those three `data-cf-node` props into a new `wrapperProps` object.

I'm not sure if that was expected to fix the styling problems for nested containers and such, but it doesn't look like it did so I need to keep the other changes for `editorWrapperStyles` which solve those issues.

primeinteger:
> I think we should think about applying the following props to the wrapper draggable component instead of the underlying component

@chasepoirier - I pushed updates to split those three `data-cf-node` props into a new `wrapperProps` object.

I'm not sure if that was expected to fix the styling problems for nested containers and such, but it doesn't look like it did so I need to keep the other changes for `editorWrapperStyles` which solve those issues.

primeinteger:
I pushed additional updates to the `useComponentProps` hook to fix the Heading block width settings to behave consistently in Editor mode vs Preview/Delivery mode.  

I think the changes landed in a better spot with that update (less of the ternary logic for structure component vs not). 

Now all components will split the editor styles vs component styles in the same way (which is basically to move the `margin`, `width`, `maxWidth` and `height` to the editor wrapper (draggable) element, and _normalizing_ these styles on the component element to fill the wrapper space (`margin: 0; width: 100%; height: 100%; maxWidth: none`)

chasepoirier:
> I pushed additional updates to the `useComponentProps` hook to fix the Heading block width settings to behave consistently in Editor mode vs Preview/Delivery mode.
> 
> I think the changes landed in a better spot with that update (less of the ternary logic for structure component vs not).
> 
> Now all components will split the editor styles vs component styles in the same way -- basically after moving the `margin`, `width`, `maxWidth` and `height` to the editor wrapper (draggable) element, the styles are adjusted on the component element to fill the wrapper space (`margin: 0; width: 100%; height: 100%; maxWidth: none`)

Nice I like this solution a lot, good work!

chasepoirier (approved):

PR #297, opened 2024-01-29
feat: updates button component definition to hold default stylings rather than using css modules [ALT-312]

ryunsong-contentful:
## Purpose
Similar to @primeinteger's PR https://github.com/contentful/experience-builder/pull/296 where he updated the header. This one's for Button.
Screenshot 2024-01-29 at 4 54 26 PM

NOTES:
1. This PR should be merged after adrian's PR.
2. We need to discuss having a cfBorderRadius property


packages/components/src/components/Button/index.ts
```diff
@@ -7,13 +7,78 @@ export const ButtonComponentDefinition: ComponentDefinition = {
...
+      description: 'The width of the button.',
+      defaultValue: '92px',
+    },
+    cfHeight: {
+      displayName: 'Height',
+      type: 'Text',
+      group: 'style',
+      description: 'The height of the button.',
+      defaultValue: '43px',
+    },
```
primeinteger, line 70: I'm a little worried about the fixed height/width.  Is there a default value along the lines of "auto" that would allow the height and width to be dynamic based on the text content and padding of the button instead?

primeinteger: Or maybe if the default size of the button was wider like 200px by 40px that would be nice way to give a little more room for text changes without having to introduce something new to allow an "auto" width option

primeinteger (approved):
🚀

PR #299, opened 2024-01-30
chore: bail rollup and lerna when TS errors []

Chaoste:
Currently, a TS error does not yield an error when building with rollup.
Using the config [noEmitOnError](https://www.npmjs.com/package/@rollup/plugin-typescript#noemitonerror), TS will now return an error that can be consumed by rollup and leads to a subsequent abortion of rollup and lerna.

Video: https://www.loom.com/share/97ba8ed8671f4d5187514b88332bfcd9

Related bug: https://github.com/rollup/rollup-plugin-typescript/issues/43

### DIFF
--- a/packages/components/tsconfig.json
+++ b/packages/components/tsconfig.json
@@ -11,12 +11,13 @@
     "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
     "forceConsistentCasingInFileNames": true,
+    "noEmitOnError": true,
     "baseUrl": ".",
     "paths": {
       "@components/*": ["./src/components/*"],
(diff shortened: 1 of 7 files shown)


dimitrycf (approved):
Seems to work great and now aborts build on errors!

image

Chaoste:
> Actually once I cleared npm cache and rebuilt SDK again it showed this error:
> 
> image
> 
> We can quick fix it by allowing the `coordinates` type to accept `null` (as we don't control de facto value which comes there; and the `coordinates` is only used in the file for triggering an effect.
> 
> image

PR #316, opened 2024-01-30
feat(components): built-in styles for columns [ALT-276, ALT-277]

chasepoirier:


packages/core/src/definitions/styles.ts
```diff
@@ -319,19 +319,137 @@ export const containerBuiltInStyles: Partial<
...
+        },
+      ],
+    },
     type: 'Text',
-    defaultValue: '6',
     group: 'style',
-  } as ComponentDefinitionVariable<'Text'>,
-  cfHeight: {
-    displayName: 'Height',
+    description: 'The horizontal alignment of the section',
```
ryunsong-contentful, line 344: description: should it be `section` or `column`?

packages/core/src/definitions/styles.ts
```diff
@@ -319,19 +319,137 @@ export const containerBuiltInStyles: Partial<
...
+    group: 'style',
+    description: 'Wrap objects',
+    defaultValue: 'nowrap',
+  },
+  cfBorder: {
+    displayName: 'Border',
+    type: 'Text',
+    group: 'style',
+    description: 'The border of the section',
+    defaultValue: '0px outside rgba(255, 255, 255, 1)',
```
ryunsong-contentful, line 404: Isn't this just a white border? Should we just make it transparent like the `cfBackgroundColor`?

packages/core/src/definitions/styles.ts
```diff
@@ -343,7 +461,21 @@ export const columnsBuiltInStyles: Partial<
...
+  },
+  cfWidth: {
+    displayName: 'Width',
+    type: 'Text',
+    group: 'style',
+    description: 'The width of the columns',
+    defaultValue: 'fill',
+  },
+  cfMaxWidth: {
+    displayName: 'Max Width',
```
ryunsong-contentful, line 474: Should be `Max width` (w is lower-case)

ryunsong-contentful (approved):
LGTM

PR #319, opened 2024-01-31
fix: update Text component built in styles and add hyperlink options [ALT-358]

primeinteger:
## Purpose

Changes to the Text component

* Enable hyperlink options (url, target)
* Enable additional built-in styles (background, border, size)
* Default text from "Lorem ipsum" to "Text"

Changes to the Heading component

* Default height value from "fill" (inherited) to "fit-content" (override)


chasepoirier (approved):
LGTM 🤖

PR #320, opened 2024-01-31
fix: breakpoints now correctly affect the editor ui properly

ryunsong-contentful:
## Purpose
Breakpoints now properly display the correct view for desktop, mobile, and tablet

### DIFF
--- a/packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
+++ b/packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
@@ -22,21 +22,23 @@ import { useCallback, useEffect, useState } from 'react';
  * and then decending by screen width. For mobile-first designs, the order would be ascending
  */
 export const useBreakpoints = (breakpoints: Breakpoint[]) => {
-  const [mediaQueryMatchers, initialMediaQueryMatches] = mediaQueryMatcher(breakpoints);
-
-  const [mediaQueryMatches, setMediaQueryMatches] =
-    useState<Record<string, boolean>>(initialMediaQueryMatches);
-
   const fallbackBreakpointIndex = getFallbackBreakpointIndex(breakpoints);
+  const [mediaQueryMatchers, currentQueryMatches] = mediaQueryMatcher(breakpoints);
+  const [mediaQueryMatches, setMediaQueryMatches] =
+    useState<Record<string, boolean>>(currentQueryMatches);
+
+  const activeBreakpointIndex = getActiveBreakpointIndex(
+    mediaQueryMatches,
+    fallbackBreakpointIndex
+  );
 
   // Register event listeners to update the media query states
   useEffect(() => {
-    const eventListeners = mediaQueryMatchers.map(({ id, signal }) => {
-      const onChange = () =>
-        setMediaQueryMatches((prev) => ({
-          ...prev,
-          [id]: signal.matches,
-        }));
+    const eventListeners = mediaQueryMatchers.map(({ signal }) => {
+      const onChange = () => {
+        const currentQueryMatches = mediaQueryMatcher(breakpoints)[1];
+        setMediaQueryMatches(currentQueryMatches);
+      };
       signal.addEventListener('change', onChange);
       return onChange;
     });
@@ -46,13 +48,7 @@ export const useBreakpoints = (breakpoints: Breakpoint[]) => {
         mediaQueryMatchers[index].signal.removeEventListener('change', eventListener);
       });
     };
-  }, [mediaQueryMatchers]);
-
-  const activeBreakpointIndex = getActiveBreakpointIndex(
-    breakpoints,
-    mediaQueryMatches,
-    fallbackBreakpointIndex
-  );
+  }, [breakpoints, fallbackBreakpointIndex, mediaQueryMatchers, mediaQueryMatches]);
 
   const resolveDesignValue: ResolveDesignValueType = useCallback(
     (
--- a/packages/visual-editor/src/hooks/useBreakpoints.ts
+++ b/packages/visual-editor/src/hooks/useBreakpoints.ts
@@ -22,22 +22,23 @@ import { useCallback, useEffect, useState } from 'react';
  * and then decending by screen width. For mobile-first designs, the order would be ascending
  */
 export const useBreakpoints = (breakpoints: Breakpoint[]) => {
-  const [mediaQueryMatches, setMediaQueryMatches] = useState<Record<string, boolean>>({});
-
   const fallbackBreakpointIndex = getFallbackBreakpointIndex(breakpoints);
+  const [mediaQueryMatchers, currentQueryMatches] = mediaQueryMatcher(breakpoints);
+  const [mediaQueryMatches, setMediaQueryMatches] =
+    useState<Record<string, boolean>>(currentQueryMatches);
+
+  const activeBreakpointIndex = getActiveBreakpointIndex(
+    mediaQueryMatches,
+    fallbackBreakpointIndex
+  );
 
   // Register event listeners to update the media query states
   useEffect(() => {
-    const [mediaQueryMatchers, initialMediaQueryMatches] = mediaQueryMatcher(breakpoints);
-    // Store the media query state in the beginning to initialise the state
-    setMediaQueryMatches(initialMediaQueryMatches);
-    const eventListeners = mediaQueryMatchers.map(({ id, signal }) => {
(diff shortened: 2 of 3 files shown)


primeinteger (approved):
It looks like there's a duplicate `useBreakpoints` hook in `sdk/src/hooks/useBreakpoints.ts`

Should the same change be applied to that hook as well?

packages/visual-editor/src/hooks/useBreakpoints.ts
```diff
@@ -50,7 +50,7 @@ export const useBreakpoints = (breakpoints: Breakpoint[]) => {
 
   const activeBreakpointIndex = getActiveBreakpointIndex(
     breakpoints,
-    mediaQueryMatches,
+    initialMediaQueryMatches || mediaQueryMatches,
```
Chaoste, line 53: Mmmmh, this looks a bit shady to be honest.

This would mean that we never consume `mediaQueryMatches` which is the variable that gets updated when resizing the canvas. So this change would mean that we only show the designs of the initial canvas size and don't care anymore when you switch between desktop/ table/ mobile view.

If we really wanted to remove this feature, we should rather drop the whole state variable and effect instead of having this shady overwrite.

Chaoste (changes requested):

packages/visual-editor/src/hooks/useBreakpoints.ts
elylucas: We should probably move this deterministic logic into the `getActiveBreakPointIndex` method that comes from core so the logic can be reused by the other breakpoint functionality we have in various places.

elylucas (changes requested):

packages/visual-editor/src/hooks/useBreakpoints.ts
Chaoste: We have the exact same logic already in the SDK in the similar called file `useBreakpoints` [here](https://github.com/contentful/experience-builder/blob/6c7b72264e91029b78e349e37cf6195b2dc4b9eb/packages/experience-builder-sdk/src/hooks/useBreakpoints.ts#L24). So yeah, sounds like something to pull into core. Though, I'm wondering a bi right now why we didn't go with native media queries in delivery mode - the current approach is a very JS-heavy one which might not work always for the customer as expected I think. Not sure if this would affect SSG applications in any way actually.

ryunsong-contentful:
Okay updated the fix so that it's much cleaner.
Basically the issue was that the listeners are being set to the write property. However, the `setMediaQueryMatches` was using the wrong matcher. Probably due to some copy pasta error from the past since the names are so similar.

In anycase, I cleaned up the useEffect so that it works as it was originally intended.

Additionally, moving the logic to the core package is more complex because @chasepoirier pointed out to me that we want to avoid react dependencies in core which would be introduced if we moved the useBreakpoint listener logic. So for this issue, I'll create a task in our backlog to track this so that we can workshop and groom this when appropriate.

elylucas (approved):

chasepoirier:
@ryunsong-contentful overall looks great, I tested locally and breakpoints began working properly so nice job!

Small thing, can you add this change into the `useEditorSubsciber.ts` file so that the component stays selected after changing breakpoints?

```ts
       case INCOMING_EVENTS.CanvasResized:
       case INCOMING_EVENTS.SelectComponent: {

      }
```

packages/core/src/utils/breakpoints.ts
```diff
@@ -49,21 +41,14 @@ export const mediaQueryMatcher = (breakpoints: Breakpoint[]) => {
...
-    isMatch: mediaQueryMatches[id] ?? true,
-  }));
-
-  // Find the last breakpoint in the list that matches (desktop-first: the narrowest one)
-  const mostSpecificIndex = findLast(breakpointsWithMatches, ({ isMatch }) => isMatch)?.index;
-  return mostSpecificIndex ?? fallbackBreakpointIndex;
+  if (Object.keys(mediaQueryMatches).length === 0) return fallbackBreakpointIndex;
+  if (mediaQueryMatches.mobile) return 2;
+  if (mediaQueryMatches.tablet) return 1;
+  if (mediaQueryMatches.desktop) return 0;
```
Chaoste, line 50: We wrote the breakpoint logic so that it would allow a dynamic number instead of solely the three default ones. It would be great if we didn't remove this functionality as we would have to re-introduce it again later.

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
```diff
@@ -22,21 +22,23 @@ import { useCallback, useEffect, useState } from 'react';
...
   useEffect(() => {
-    const eventListeners = mediaQueryMatchers.map(({ id, signal }) => {
-      const onChange = () =>
-        setMediaQueryMatches((prev) => ({
-          ...prev,
-          [id]: signal.matches,
-        }));
+    const eventListeners = mediaQueryMatchers.map(({ signal }) => {
+      const onChange = () => {
+        const initialMediaQueryMatches = mediaQueryMatcher(breakpoints)[1];
```
Chaoste, line 39: What is meant by "initial" here? As far as I see, this value is regenerate on every change, so it's the "current" value I guess.

Therefore, it might make more sense to rename `currentQueryMatches` in line 26 back to `initialQueryMatches`. it is recalculated on every render but we only use it as the initial value for the state value.

packages/core/src/utils/breakpoints.ts
Chaoste: Also, note that your current logic doesn't care about whether it's `desktop-first` vs. `mobile-first`. It looks like this introduces yet another bug :/

Was any change in this file necessary for the fix? It looks like the actual fix only happened in `useBreakpoints`. If so, I would ask you to leave the logic in here unchanged to keep it adjustable to different breakpoint definitions :)

Chaoste (changes requested):
I think that we're introducing new bugs with the current changes in the utils file. I'm suggesting to revert those changes to still allow "desktop-first" styles and also a dynamic amount of breakpoints.

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
```diff
@@ -22,21 +22,23 @@ import { useCallback, useEffect, useState } from 'react';
...
 
   // Register event listeners to update the media query states
   useEffect(() => {
-    const eventListeners = mediaQueryMatchers.map(({ id, signal }) => {
-      const onChange = () =>
-        setMediaQueryMatches((prev) => ({
-          ...prev,
-          [id]: signal.matches,
-        }));
+    const eventListeners = mediaQueryMatchers.map(({ signal }) => {
```
Chaoste, line 37: The media query matchers are called on every single stage change of every single breakpoint. So e.g. if I switched from "desktop" to "mobile", you would receive two signal callbacks: one that sets "desktop" to "matches: false" and one that sets "table" to "matches: false". So we would be doing duplicate work by re-calculating everything on every single signal callback.
Can you maybe provide more background on why the previous signals actually did not work?

packages/core/src/utils/breakpoints.ts
ryunsong-contentful: Hmm for some reason when I was testing it on Friday this logic was giving the wrong output, but now I did a fresh build of everything, it seems to work as expected. So I will revert this

Chaoste:
@ryunsong-contentful 
As I built this whole logic in the very beginning, I wanted to dig into this problem and find out what was going on. I believe to have found all bug fixes in there and filed a separate PR. I hope it is okay for you that I jumped on this:
https://github.com/contentful/experience-builder/pull/339

packages/experience-builder-sdk/src/hooks/useBreakpoints.ts
ryunsong-contentful: The issue came from state not updating correctly. Basically the `mediaQueryMatches` variable in the original code failed to update and would always have `{ tablet: true }` when selecting the mobile breakpoint. However, `initialMediaQueryMatches` would have the correct latest value. So the key is to always get the latest breakpoint values which happens when you call `mediaQueryMatcher`

Chaoste: Yes, you're right. The event listener is called once for mobile and once for tablet. However, after the first call we remove the listener and thus the second one doesn't get triggered. I described this also in the other PR: https://github.com/contentful/experience-builder/pull/339

PR #321, opened 2024-02-01
fix: allow built-in components to be disabled [ALT-267]

primeinteger:
## Purpose

Allow Contentful's built-in components to be disabled through an option when calling the `defineComponents` function.

I'll add the documentation below to the Wiki after the PR is merged.

## Configuring Contentful Built-In Components

Contentful provides several built-in components such as Heading, Text, Rich Text, Image, and Button. You can enable or disable these components when defining your custom components using the `defineComponents` function.

### Enabling All Built-In Components

To enable all built-in components, simply call `defineComponents` with your custom component definitions. The built-in components are enabled by default.

```ts
defineComponents(
  [
    /* custom component definitions */
  ]
);
```

### Disabling All Built-In Components

To disable all built-in components, pass an options object to `defineComponents` with `enabledBuiltInComponents` set to an empty array.

```ts
defineComponents(
  [
    /* custom component definitions */
  ],
  {
    enabledBuiltInComponents: [],
  }
);
```

### Enabling Specific Built-In Components

You can also choose to enable only specific built-in components. To do this, pass an array of the IDs of the components you want to enable to `enabledBuiltInComponents`.

In the following example, all built-in components except the Image component are enabled:

```ts
import { CONTENTFUL_COMPONENTS } from '@contentful/experience-builder';

defineComponents(
  [
    /* custom component definitions */
  ],
  {
    enabledBuiltInComponents: [
      CONTENTFUL_COMPONENTS.button.id,
      CONTENTFUL_COMPONENTS.heading.id,
      CONTENTFUL_COMPONENTS.richText.id,
      CONTENTFUL_COMPONENTS.text.id,
    ],
  }
);
```

Note: The Image component is disabled in this example because its ID (`CONTENTFUL_COMPONENTS.image.id`) is not included in the `enabledBuiltInComponents` array.


packages/core/src/constants.ts
```diff
@@ -63,11 +63,22 @@ export const CONTENTFUL_SECTION_ID = 'contentful-section'; // TODO: remove me on
...
+export const CONTENTFUL_BUTTON_ID = 'button';
+export const CONTENTFUL_HEADING_ID = 'heading';
+export const CONTENTFUL_IMAGE_ID = 'image';
+export const CONTENTFUL_RICH_TEXT_ID = 'richText';
+export const CONTENTFUL_TEXT_ID = 'text';
 export const CONTENTFUL_SECTION_NAME = 'Section'; // TODO: remove me once all customers are using 2023-09-28 schema version
 export const CONTENTFUL_COLUMNS_NAME = 'Columns';
 export const CONTENTFUL_SINGLE_COLUMN_NAME = 'Column';
 export const CONTENTFUL_CONTAINER_NAME = 'Container';
+export const CONTENTFUL_BUTTON_NAME = 'Button';
```
Chaoste, line 75: [nitpick]
Maybe, the time has come to rename or re-arrange those component-related constants as will expose all of this on the SDK and customers could use those.

**Idea 1: Create a map** (preferred)
```
export const CONTENTFUL_COMPONENT_IDS = {
    Button: 'button',
    // ...
    Container: 'contentful-container',
}
/** @deprecated will be replaced by CONTENTFUL_COMPONENT_IDS.Container in v5 */
export const CONTENTFUL_CONTAINER_ID = CONTENTFUL_COMPONENT_IDS.Container;

// --> Similar object literal for names
```

**Idea 2: Use a more specific suffix**
```
export const CONTENTFUL_COMPONENT_BUTTON_ID = 'button';
```

packages/core/src/constants.ts
```diff
@@ -63,11 +63,22 @@ export const CONTENTFUL_SECTION_ID = 'contentful-section'; // TODO: remove me on
 export const CONTENTFUL_CONTAINER_ID = 'contentful-container';
 export const CONTENTFUL_COLUMNS_ID = 'contentful-columns';
 export const CONTENTFUL_SINGLE_COLUMN_ID = 'contentful-single-column';
+export const CONTENTFUL_BUTTON_ID = 'button';
```
Chaoste, line 66: [nitpick]
It's by no means necessary but for looking at the persisted data and SDK messages, it would be helpful to use the same prefix as for the container
```suggestion
export const CONTENTFUL_BUTTON_ID = 'contentful-button';
```

(same below)

Chaoste (approved):
Nice one!
Just a small naming feedback on the constants that we make availabe on the outside

packages/core/src/constants.ts
primeinteger: I think it's intentional to keep the optional built-in component IDs more generic than the structure components in case the customer application wants to register their own button to override the contentful built-in button.

If we change our built-in component's IDs to have the `contentful-` prefix, most likely the customer would never override our button (since they would have to register their component as `contentful-button` to do that.  Instead, they would have to disable/opt-out of the contentful button.

Thoughts?

packages/core/src/constants.ts
primeinteger: I like the first idea.  I originally started making an Enum (similar to your map idea) but wasn't sure how to navigate the deprecation side and figured it would add more scope to the effort (such as variable replacement and UI changes) to switch all of the constants to use a new Enum instead.  I can do that as it would probably be better for the long term.

primeinteger: I'm thinking about this more and I could see an object being better than an enum because we could organize related props together like this:
```ts
export const CONTENTFUL_COMPONENTS = {
  button: {
    id: 'button',
    name: 'Button',
    category: CONTENTFUL_DEFAULT_CATEGORY,
  },
  container: {
    id: 'contentful-container',
    name: 'Container',
    category: CONTENTFUL_COMPONENT_CATEGORY
  },
  // ...
}
```

packages/core/src/constants.ts
elylucas: I think with the ability to disable the builtins now, that overriding is less important than it was. If we kept the current names, I could see a scenario where a user registers their own button, but gets confused on why our button suddenly disappeared. Prefixing the component names with something would help solve that, and its more understood how components would show up with the disable feature.

packages/core/src/constants.ts
elylucas: The constants only get exposed if they are reexported from the SDK package, but yes, putting some more organization around them by namespacing or enums is a good call.

Chaoste: In the TS guild, we once agreed that we want to avoid enums if we don't need an exclusive feature from it:
https://contentful.atlassian.net/wiki/spaces/ENG/pages/4189094335/TypeScript+Standards#Enums

packages/core/src/constants.ts
primeinteger: If we change the built-in component ids it will probably break experiences that have added these components.  I'll test that and see what happens later today.

primeinteger: Renaming component ids does impact experiences. I think I'll skip changing the ids for now.

Screenshot 2024-02-01 at 3 06 37 PM

PR #322, opened 2024-02-01
Container and Section labels [ALT-355]

ryunsong-contentful:
## Purpose
Adds background label to the Container and Section component.
Also cuts the Container label short on either end if the container itself has too small of a width
Container label no longer shows up when it's non-empty

Screenshot 2024-02-01 at 2 16 36 PM
Screenshot 2024-02-01 at 9 50 55 AM
Screenshot 2024-02-01 at 9 51 05 AM

### DIFF
--- a/packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
+++ b/packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
@@ -5,6 +5,7 @@ import { Flex } from '../Layout/Flex';
 import { ContentfulContainerAsHyperlink } from './ContentfulContainerAsHyperlink';
 import type { ContentfulContainerAsHyperlinkProps } from './ContentfulContainerAsHyperlink';
 import { combineClasses } from '../../utils/combineClasses';
+import { CONTENTFUL_SECTION_ID } from '@contentful/experience-builder-core/constants';
 
 export const ContentfulContainer: React.FC<ContentfulContainerAsHyperlinkProps> = (props) => {
   const { className, editorMode, children, cfHyperlink } = props;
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
```diff
@@ -38,10 +38,18 @@ export const ContentfulContainer: React.FC<ContentfulContainerAsHyperlinkProps>
...
-  return renderDropzone(node, {
-    ['data-test-id']: 'contentful-container',
-    id: 'ContentfulContainer',
-    className: combineClasses(className, 'defaultStyles'),
-    WrapperComponent: Flex,
-  });
+  const isEmpty = !node.children.length;
+
+  return (
+    <div className="cf-container-wrapper">
```
elylucas, line 44: could this be done without wrapping it in another div container?

elylucas (changes requested):

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
ryunsong-contentful: Maybe we can discuss this as a parking lot in standup. This is what the `SingleColumn.tsx` component does so I followed in that same vein. 

Off the top of my head I can't think of anything but maybe we can come up with an alternate solution during the parking lot

ryunsong-contentful: Addressed via your idea of only having a wrapper div when it's empty

elylucas (approved):
noice!

PR #326, opened 2024-02-02
fix: smaller dropzone indicator and container height adjustments [ALT-356]

primeinteger:
## Purpose

* Changed the height behavior of structure components
  * Structure components start with `min-height: 120px` when first dropped into the canvas and are empty
  * Once an element is added to the component, the height adjusts to match the combined height of all children
  * If the user changes the container height to a fixed height in the design settings, the fixed height value will be used instead
  * This behavior is consistent with what the user sees in the editor canvas vs what renders in preview/delivery mode
  * Demo:

  https://github.com/contentful/experience-builder/assets/8539634/ebe877c0-2d17-42fc-bbd7-1acb7ec04048

* Decreased the size of drop zone indicators
  * Height of horizontal drop zone indicators decreased from 74px -> 40px
  * Width of vertical drop zone indicators decreased from 92px -> 40px
  * Small containers will be expanded while dragging by using the `onBeforeCapture` responder to attach a `min-height: 80px` style to containers
  * Demo:
  
  https://github.com/contentful/experience-builder/assets/8539634/4e530910-7f52-4ec9-a3de-d0881d17b616


elylucas:
I think the size of the old dropzone indicators felt more natural as they matched the size of the card being dropped in. The larger sizes give more affordance to the user as well. Was this change approved by design?  If not, we should test it a bit to make sure it feels good.

primeinteger:
> I think the size of the old dropzone indicators felt more natural as they matched the size of the card being dropped in. The larger sizes give more affordance to the user as well. Was this change approved by design? If not, we should test it a bit to make sure it feels good.

Yeah, the change to drop zone size wasn't part of the original request from design, but it seemed a necessary part of the solution to allow containers to adjust their height to fit content. I did work closely with design and had their approval on these change.

The larger 74px drop zone hitbox doesn't work when the height of the container is less than 100px.  I wanted to decrease the drop zones hitbox height to 40px to align with the future height of the component cards planned in future designs, but even that height has issues with containers holding a single button or text component.  Using 20px tall drop zone hitboxes seemed like the tallest we could allow to still work with the new container height adjustment logic.

I had experimented with some other options like dynamically attaching min-height and padding to containers only when dragging over the canvas, or only dragging over the individual container, but these solutions create even more issues with calculating drop zone positions when the content is animating/moving around and leads to instances of the canvas rapidly flashing drop zones and hover states.

elylucas:
> > I think the size of the old dropzone indicators felt more natural as they matched the size of the card being dropped in. The larger sizes give more affordance to the user as well. Was this change approved by design? If not, we should test it a bit to make sure it feels good.
> 
> Yeah, the change to drop zone size wasn't part of the original request from design, but it seemed a necessary part of the solution to allow containers to adjust their height to fit content. I did work closely with design and had their approval on these change.
> 
> The larger 74px drop zone hitbox doesn't work when the height of the container is less than 100px. I wanted to decrease the drop zones hitbox height to 40px to align with the future height of the component cards planned in future designs, but even that height has issues with containers holding a single button or text component. Using 20px tall drop zone hitboxes seemed like the tallest we could allow to still work with the new container height adjustment logic.
> 
> I had experimented with some other options like dynamically attaching min-height and padding to containers only when dragging over the canvas, or only dragging over the individual container, but these solutions create even more issues with calculating drop zone positions when the content is animating/moving around and leads to instances of the canvas rapidly flashing drop zones and hover states.

Cool thanks for the explanation

elylucas (approved):

packages/visual-editor/src/components/Dropzone/styles.module.css
```diff
@@ -7,7 +7,10 @@
   outline: 2px solid transparent;
   width: 100%;
   transition: outline 0.2s;
-  min-height: 120px;
+}
+
+.isExpanded {
+  min-height: 80px;
```
chasepoirier, line 13: Thoughts on adding something like `padding: 12px` to this drag state?

primeinteger: I think it's a good suggestion and more padding could help with visualizing nested containers.

I tried some different `padding` and `margin` values here and unfortunately it throws off the drop zone positioning for some reason.  I went down the rabbit hole a little bit looking at `usePlaceholderStyle` but didn't see a way to fix the drop zone positions after `padding` or `margin` is added. 🤔 lets consider it for future updates

Screenshot 2024-02-05 at 5 13 38 PM

PR #339, opened 2024-02-06
fix(experience-builder): rendering breakpoint-specific design values [SPA-1772]

Chaoste:
## Purpose
There were at least two bugs when having breakpoint-specific styles:

## Bug 1
When opening the experience builder, we used to render the mobile value first. There were two problems related to that:

#### Fix 1
The "matches" state variable is an empty object initially. It is a state variable that receives the `initialMediaQueryMatches` only once. Though, on the first run the breakpoint definitions are not yet defined, so the initial value becomes `{}`.
**Solution**: Fill the state variable ones in the effect and trigger the effect only when the breakpoint definitions changed.

#### Fix 2
Now when the "matches" variable is `{}`, the util function just assumes `true` for every breakpoint. Due to the specificity priority for desktop-first, we ended up assuming we're on mobile.
**Solution:** We only assume `true` for the fallback breakpoint but otherwise `false`.

## Bug 2
When switching from "desktop" to "mobile" (or vice versa), it didn't end up in the right breakpoint state.

#### Fix
The signal callbacks for "tablet" and "mobile" get called separately for each change. So in this scenario, we have two callback invocations. However, the first one did already change the state which retriggered the effect. Before rerunning the effect, it removed all signal listeners before the second one could alter the state.
**Solution**: The effect reruns and initializes the signal listeners only when the breakpoint definition changes. This works but we need to be aware later on that we might not have the latest reference to the state variable. Currently, we don't need we change the state via `set((prev) => ...)`.

## Screen recording
I added logs to show the proper state change and callback invocations.

https://github.com/contentful/experience-builder/assets/9327071/6e0c4da0-4ef9-455b-8797-b82c22a821fc

### DIFF
--- a/packages/core/src/utils/breakpoints.ts
+++ b/packages/core/src/utils/breakpoints.ts
@@ -58,7 +58,7 @@ export const getActiveBreakpointIndex = (
     id,
     index,
     // The fallback breakpoint with wildcard query will always match
-    isMatch: mediaQueryMatches[id] ?? true,
+    isMatch: mediaQueryMatches[id] ?? index === fallbackBreakpointIndex,
   }));
 
   // Find the last breakpoint in the list that matches (desktop-first: the narrowest one)
--- a/packages/visual-editor/src/hooks/useBreakpoints.ts
+++ b/packages/visual-editor/src/hooks/useBreakpoints.ts
@@ -22,21 +22,22 @@ import { useCallback, useEffect, useState } from 'react';
  * and then decending by screen width. For mobile-first designs, the order would be ascending
  */
 export const useBreakpoints = (breakpoints: Breakpoint[]) => {
-  const [mediaQueryMatchers, initialMediaQueryMatches] = mediaQueryMatcher(breakpoints);
-
-  const [mediaQueryMatches, setMediaQueryMatches] =
-    useState<Record<string, boolean>>(initialMediaQueryMatches);
+  const [mediaQueryMatches, setMediaQueryMatches] = useState<Record<string, boolean>>({});
 
   const fallbackBreakpointIndex = getFallbackBreakpointIndex(breakpoints);
 
   // Register event listeners to update the media query states
   useEffect(() => {
+    const [mediaQueryMatchers, initialMediaQueryMatches] = mediaQueryMatcher(breakpoints);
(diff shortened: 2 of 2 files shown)


Spring3 (approved):
Checked locally - works in both editor and preview/delivery. 

Surprisingly, if I resized the browser window in editor mode while on canvas view, the styles from let's say "tablet" would not apply to the iframe when the canvas size would match the tablet screen size, but outside of editor mode, it would behave as expected. See the video attached to see exactly what I mean. Shouldn't it behave in a similar way?

https://github.com/contentful/experience-builder/assets/4171202/67c40b17-c693-4436-85e2-c9c7e8c55d42

Chaoste:
> Surprisingly, if I resized the browser window in editor mode while on canvas view, the styles from let's say "tablet" would not apply to the iframe when the canvas size would match the tablet screen size, but outside of editor mode, it would behave as expected. See the video attached to see exactly what I mean. Shouldn't it behave in a similar way?

During the implementation back then, we discussed this with Rui and Sarah. We aligned on the fact that the selected breakpoint in the top menu should always be visible. Otherwise we open a can of worms overcome these mixed states. So when you select desktop, the iframe will at least have the minimum necessary width of the desktop size. That's why you can also scroll horizontally on the canvas. Otherwise, you might not be able to test larger media settings on a typical MacBook.

Spring3:
> > Surprisingly, if I resized the browser window in editor mode while on canvas view, the styles from let's say "tablet" would not apply to the iframe when the canvas size would match the tablet screen size, but outside of editor mode, it would behave as expected. See the video attached to see exactly what I mean. Shouldn't it behave in a similar way?
> 
> During the implementation back then, we discussed this with Rui and Sarah. We aligned on the fact that the selected breakpoint in the top menu should always be visible. Otherwise we open a can of worms overcome these mixed states. So when you select desktop, the iframe will at least have the minimum necessary width of the desktop size. That's why you can also scroll horizontally on the canvas. Otherwise, you might not be able to test larger media settings on a typical MacBook.

nah it makes sense, I was just thinking that perhaps for some users it might be easier to use native Chrome's or browser's tools to change the viewpoint and then it would perhaps make sense to adjust the contents of the canvas to respect the new dimentions and viewports

But ok, good to know that it was a conscious decision

PR #340, opened 2024-02-06
chore: add dev scripts for local development

elylucas:
## Purpose

A recent change to add the tsconfig 'noEmitOnError' config value caused the rollup watch server to hault when there was any type of compilation error.

This PR adds new "dev" scripts to each of main EB packages to fire up the rollup watch server with a dev mode argument, which will disable the 'noEmitOnError' option, allowing the process to still report the error but continue running. 

When building the application (npm run build), the 'noEmitOnError' option will still be used and will error out if there is a compilation error.

To use the new scripts, go into the packages folder, and run 'npm run dev', which will build the application and then continually watch for file changes, which will retrigger a build.

### DIFF
--- a/packages/storybook-addon/package.json
+++ b/packages/storybook-addon/package.json
@@ -40,7 +40,7 @@
     "*.d.ts"
   ],
   "scripts": {
-    "dev:": "echo noop",
+    "dev": "echo noop",
     "clean": "rimraf ./dist",
     "prebuild": "npm run clean",
     "build": "tsup",
--- a/packages/storybook-addon/package.json
+++ b/packages/storybook-addon/package.json
@@ -40,6 +40,7 @@
     "*.d.ts"
   ],
   "scripts": {
+    "dev:": "echo noop",
     "clean": "rimraf ./dist",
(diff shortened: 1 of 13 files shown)


primeinteger (approved):
Nice! Thanks for making these changes! This will be very helpful for local development

packages/storybook-addon/package.json
```diff
@@ -40,6 +40,7 @@
     "*.d.ts"
   ],
   "scripts": {
+    "dev:": "echo noop",
```
Chaoste, line 43:
```suggestion
    "dev": "echo noop",
```

Chaoste: Or was this intentional? (same q for the other noop package scripts)

Chaoste (approved):
Nice improvement! Does this mean, I can run `npm run dev` inside the `test-app` and get all changes right away from all deps?

packages/storybook-addon/package.json
elylucas: no copy pasta

PR #341, opened 2024-02-06
feat(experience-builder-sdk): allow delivery mode in editor

elylucas:
## Purpose

Currently, the client app needs to be in 'preview' mode in order to correctly load in the editor. The concept of 'preview' and 'delivery' are Contentful API concepts that don't directly apply to editing a page in Experience Builder. Since the content editor page is always what is in preview mode (regardless of what the provided mode is), it seemed arbitrary to constrain the app to only work in preview mode, or even need a mode at all to work.

I went in to take the requirement of the app being in preview mode, and noticed that all the other uses of `mode` in the packages were mostly superfluous, any useful ones were just checking that the mode was a valid one. This showed that the mode wasn't truly needed in the SDK, so most of this PR is removing the usage of node throughout the code base. It was mostly just passing the mode around and in tests.

All the public API endpoints now mark mode deprecated when passed in as a parameter or prop to component, and most of its internal usage has been removed.


packages/experience-builder-sdk/src/hooks/useEditorModeSwitch.ts
```diff
@@ -1,27 +1,16 @@
...
-  mode: InternalSDKMode;
-  switchToEditorMode: () => void;
-}) => {
+export const useEditorModeSwitch = ({ switchToEditorMode }: { switchToEditorMode: () => void }) => {
   const hasConnectEventBeenSent = useRef(false);
 
-  // switch from preview mode to editor mode
+  // switch to editor mode
   useEffect(() => {
-    if (mode !== 'preview') {
```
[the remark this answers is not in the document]
elylucas: I reworked this logic into a new `useDetectEditorMode` hook, give it a look

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (typeof window !== 'undefined' && !window.__EB__?.isEditorMode) {
+      window.addEventListener('message', onMessage);
+      sendMessage(OUTGOING_EVENTS.Connected);
+
+      setTimeout(() => {
+        if (!receivedMessage.current) {
+          // if message is not received back in time, set editorMode back to false
+          setIsEditorMode(false);
+        }
+      }, 100);
```
Chaoste, line 42: I'm not sure if 100 ms will be that reliable. The web app can be quite slow for some spaces and browsers.
What would currently happen if it takes longer? I guess we show an empty screen for those 100ms and then try to fetch preview data before we finally go into editor mode, right?

**Suggestion for a follow-up ticket:**
I guess this would become relevant for customers who render the preview inside an iframe. In this case, we render an empty screen for 100ms before correctly switching to the preview. Having this 100ms pause or making it even longer doesn't sound like a good option either. Maybe we can make the initial check a bit smarter to avoid this timeout right away, e.g. we could add a query parameter in the preview URL like `cfEditor=true` so this logic here can check for it and doesn't need to rely on `inIframe` which is not a sufficient condition to assume an editor mode.

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const onMessage = (event: MessageEvent) => {
+      if (doesMismatchMessageSchema(event)) {
+        return;
+      }
+      const eventData = tryParseMessage(event);
+
+      if (eventData.eventType === INCOMING_EVENTS.RequestEditorMode) {
+        setIsEditorMode(true);
+        receivedMessage.current = true;
+        if (typeof window !== 'undefined') {
```
Chaoste, line 24: [nitpick]
We could move this check upward to line 15 and make it a guard clause as it must be defined for a working editor mode:
```
useEffect(() => {
  if (typeof window === 'undefined') return;
```

Chaoste (approved):
The changes look good to go. Just added some general comments on the current code but not mandatory for this PR

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+  doesMismatchMessageSchema,
+  sendMessage,
+  tryParseMessage,
+} from '@contentful/experience-builder-core';
+import { INCOMING_EVENTS, OUTGOING_EVENTS } from '@contentful/experience-builder-core/constants';
+
+export const useDetectEditorMode = () => {
+  //Assume we are in editor mode initially if we are in iframe
+  const [isEditorMode, setIsEditorMode] = useState(inIframe());
```
Spring3, line 11: we used to have this check in the very beginning, but it led to the Hydration error when next.js app was configured to run via SSR, because on server it would have value `false` while on client it would have value `true`

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -0,0 +1,57 @@
...
+        return;
+      }
+      const eventData = tryParseMessage(event);
+
+      if (eventData.eventType === INCOMING_EVENTS.RequestEditorMode) {
+        setIsEditorMode(true);
+        receivedMessage.current = true;
+        if (typeof window !== 'undefined') {
+          //Once we definitely know that we are in editor mode, we set this flag so future postMessage connect calls are not made
+          window.__EB__.isEditorMode = true;
```
Spring3, line 26: since the word `builder` is removed from the product name, should we pick a different alias instead of `EB`?

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
Spring3: We should test if this delay introduces an un-avoidable blink for 100ms duration. We'd want to avoid side-effects like this

packages/experience-builder-sdk/src/hooks/useFetchById.ts
```diff
@@ -3,25 +3,23 @@ import type { ContentfulClientApi } from 'contentful';
...
 
-export const useFetchById = ({
-  id,
-  localeCode,
-  client,
-  experienceTypeId,
-  mode,
-}: UseFetchByIdArgs) => {
+export const useFetchById = ({ id, localeCode, client, experienceTypeId }: UseFetchByIdArgs) => {
+  const isEditorMode = useDetectEditorMode();
```
Spring3, line 18: A hook doesn't persist it's internal state across different usages, meaning that the hook used here and the hook used in `ExperienceRoot` will both start with fresh initial state and write up message listeners.

This potentially could lead to an out-of-sync situation when let's say this hook will receive value true and `ExperienceRoot` will receive value false, because the message sent from the web app would be received by one of them before the other one registers an event listener.

Perhaps using zustand for it would be a more secure way

The same applies to the other places which use this hook. RequestEditorMode event is sent once and with this approach we should have a guarantee that all hooks will receive the following handshake and have their state in sync

packages/visual-editor/src/components/RootRenderer/DNDProvider.tsx
```diff
@@ -42,12 +44,17 @@ export const DNDProvider = ({ children }: Props) => {
     });
   };
 
+  const beforeCapture: OnBeforeCaptureResponder = () => {
+    setDraggingOnCanvas(true);
```
Spring3, line 48: adding another event handler might affect the tests on our web app's side and might require an update on that side

Spring3 (changes requested):
Blocking it for the time being to first check the hydration error case.

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
elylucas: This code in the old branch wasn't really doing anything as the isInIframe check was setting editorMode to true and it wouldn't be set back to false if this check didn't go through.

The only time this would be a potential issue is outside of editor, if the user is running their own experience in their own iframe. The current code would always think it would be editor mode and thus we wouldn't support users running apps in their own iframes. This is a failsafe that flips editormode back to off if we falsely identify it in the isInIFrame check.

In development, I was seeing this callback happening in under 5ms, so I felt giving it 100ms was a good balance between being safe and not delaying the users app (when only running in their own iFrame) too long.

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
elylucas: Because this is synchronous check and the hook is at the top of PreviewEditor, it should have the same behavior, but I will look into that.

packages/experience-builder-sdk/src/hooks/useFetchById.ts
elylucas: The hook itself has a bit of global state (thats the purpose of the __EB__ var on global) to deal with this issue. I didn't want to introduce a state lib that could potentially collide with userland code for one variable.

packages/visual-editor/src/components/RootRenderer/DNDProvider.tsx
elylucas: these changes aren't mine and are unrelated to this PR. Not sure why they are there, I'll look into it.

elylucas: I rebased my commits on development again and these unreleated commits are gone now

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
elylucas: we aren't making a concerted effort to rename EB in the codebase at this time, so I think this is fine.

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
Chaoste: Gotcha, you're right that we don't explicitly support this scenario. Also, we probably don't have specific public documentation on this. If there is a place for this in the helper docs, let's add that the preview of experiences is not supported to render inside iframes.

PR #342, opened 2024-02-06
chore: introduce depcruiser and remove orphan code []

Chaoste:
## Purpose

We used to have quite a lot of duplicate code in the past after introducing the monorepo. This adds the dep cruiser to do a manual check on whether there is any orphan module in the code base. Additionally, it shows warnings for circular deps.

Later on, the circular deps should also throw errors and we should add it to the pipeline.


packages/components/src/utils/constants.ts
```diff
@@ -17,5 +17,3 @@ const constants = {
       'https://images.ctfassets.net/tofsejyzyo24/7cOHx0p0J1gUCFMPcoXnFg/73144735e3db345b101509bf707a56b1/Duplex.png',
   },
 };
-
-export default constants;
```
elylucas, line 21: was this change to make the constants a named export a depcruiser recommendation or something else? I'm not against it just curious.

elylucas (commented):
Nice! I didn't know a tool like this existed. Left a question more than anything. Also curious if this should be added to the GHA builds?

packages/components/src/utils/constants.ts
Chaoste: I thought that it's related but it turned out it isn't ^^ I can revert but it's also my personal recommendation to not use default exports and probably most of the coding guidelines I saw recommend the same. A default export can and will be renamed often on import which makes it hard to find all occurrences of a certain function or component later on.

Chaoste:
> Nice! I didn't know a tool like this existed. Left a question more than anything. Also curious if this should be added to the GHA builds?

I added in the description as a future addition but actually you're right. Let's add it right away :) Just did that

anwaar931 (approved):

PR #345, opened 2024-02-07
fix: properly remove event listener on scroll [ALT-395] [SPA-1817]

ChidinmaOrajiaku:
## Purpose
 Properly remove event listener on scroll

## Approach
We had the capture option true when adding the event listener, but never specified on remove that the event listener to be removed is registered as a capturing event listener.

Before PR

https://github.com/contentful/experience-builder/assets/30434146/4abf57e1-916b-470e-9653-f81781ae6eba

After PR

https://github.com/contentful/experience-builder/assets/30434146/c8ae529c-96b9-4e26-b89c-1d0797d934b8

### DIFF
--- a/packages/visual-editor/src/hooks/useEditorSubscriber.ts
+++ b/packages/visual-editor/src/hooks/useEditorSubscriber.ts
@@ -361,7 +361,7 @@ export function useEditorSubscriber() {
     window.addEventListener('scroll', onScroll, { capture: true, passive: true });
 
     return () => {
-      window.removeEventListener('scroll', onScroll);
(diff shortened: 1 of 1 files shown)


Chaoste (approved):

Chaoste:
This is really unexpected. I wouldn't have thought that this is necessary on removal

ChidinmaOrajiaku:
> This is really unexpected. I wouldn't have thought that this is necessary on removal

Me too! Until I saw the problem of several selected node ids sent to the webapp - and this happened for every node that had been selected. That looked sus, then I read it here

Screenshot 2024-02-07 at 10 52 59

PR #348, opened 2024-02-07
feat: experience validator [SPA-1700]

SofiaMargariti:
## Purpose
Add new package for schema validations of experiences/patterns. Main purpose of the validators is to be used in CMA to validate experience entries upon save. This is also the reason for implementing this into a separate package and not into the core package. We want to keep this as lightweight as possible to avoid unnecessary dependencies being brought into `content_api`.

## Approach
A new package `@contentful/experiences-validators` is introduced. The package file structures looks like this:
```
src
|-> schemas
|    |-> version1
|    |   |-> experience.ts 
|    |-> version2
|    |   |-> experience.ts
|.   |-> latest.ts
|-> validator.ts
|-> index.ts
|-> types.ts
```
`validator.ts` exports the main validation function that is responsible for applying the correct schema validator based on the `schemaVersion` of the experience entry.
`schema/<schema_version>/experience.ts` defines the experience schema for each version.
`schemas/latest.ts` is an alias for the latest version. Currently `2023-09-28`.
Types of the basic data structures, e.g. `unboundValues`, `breakpoints`, etc are inferred by the Zod schema and exported from `types.ts`. Currently only the types of the latest version are exported. If in the future we are going to need types of multiple versions we can expose them here as needed.

#### v2023_09_01 validations
Schema validation for the latest schema version can be found in [packages/validators/src/schemas/v2023_09_28/experience.ts](https://github.com/contentful/experience-builder/pull/348/files#diff-871a3c5fed8fdf002359c4074334c7c63a623984b3e8fbc22022f519560d6927R1)

#### New types
Duplicate types has been removed from `packages/core/types.ts` and replaced with imports from `packages/validators`. core package is then reexporting those types with their old names to avoid breaking changes with renamings. (Although we should at some point remove use of `composition` term)


packages/validators/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "prebuild": "npm run clean",
+    "build": "rollup -c rollup.config.mjs",
+    "preview": "vite preview",
+    "test": "vitest",
+    "test:coverage": "vitest run --coverage"
+  },
+  "files": [
+    "dist/**/*.*"
+  ],
+  "typesVersions": {
```
Chaoste, line 25: I'm just reading that this kinda the old way of defining types exports and than one should use `exports` instead/ additionally (see [here](https://stackoverflow.com/a/75738953/4816930)). Looking at the `core` module, I think this would a valid config:
```suggestion
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js"
    }
  },
  "typesVersions": {
```
We might not need to explicitly define the `types` property as eventually, it will point automatically to the `index.d.ts` based on the exported `index.js`. We can still go this way though to be safe.

Chaoste: For reference, here you see that we have the `index.d.ts` which has the exact same content as the `types.d.ts`
Screenshot 2024-02-12 at 16 06 25

packages/validators/public/vite.svg
Chaoste, line 1: Do we need this icon really?

packages/validators/rollup.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+  {
+    input: 'src/index.ts',
+    output: [
+      {
+        dir: 'dist',
+        format: 'esm',
+        sourcemap: true,
+      },
+    ],
+    plugins: [nodeResolve(), typescript({ tsconfig: './tsconfig.json' })],
```
Chaoste, line 14: Ely added this new dev-specific config a few days ago that appears to be missing here.
```suggestion
    plugins: [
      nodeResolve(),
      typescript({ tsconfig: './tsconfig.json', noEmitOnError: process.env.DEV ? false : true }),
    ],
```

packages/validators/rollup.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+    external: [/node_modules\/(?!tslib.*)/],
+  },
+  //typings
+  {
+    input: 'src/index.ts',
+    output: [{ dir: 'dist', format: 'esm', preserveModules: true }],
+    plugins: [dts({ tsconfig: './tsconfig.json' })]
```
Chaoste, line 21:
```suggestion
    plugins: [
      dts({
        tsconfig: './tsconfig.json',
        compilerOptions: { noEmitOnError: process.env.DEV ? false : true },
      }),
    ],
```

packages/validators/rollup.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+  },
+  {
+    input: 'src/types.ts',
+    output: [{ dir: 'dist', format: 'esm', preserveModules: true }],
+    plugins: [dts({ tsconfig: './tsconfig.json' })]
```
Chaoste, line 26:
```suggestion
    plugins: [
      dts({
        tsconfig: './tsconfig.json',
        compilerOptions: { noEmitOnError: process.env.DEV ? false : true },
      }),
    ],
```

packages/validators/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "*": {
+      "types": [
+        "./dist/types.d.ts"
```
Chaoste, line 28:
```suggestion
        "./dist/index.d.ts"
```
As mentioned above we have the same content already in the `index.d.ts` which is a more appropriate export file for this module.

packages/validators/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@contentful/experiences-validators",
+  "private": true,
+  "version": "0.0.0",
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/types.d.ts",
```
Chaoste, line 7:
```suggestion
  "types": "./dist/index.d.ts",
```
See in the other comments on `typesVersions` and `exports` why this might make more sense.

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,186 @@
...
+  variables: z.record(propertyKeySchema, ComponentPropertyValueSchema),
+});
+
+export type ComponentTreeNode = z.infer<typeof baseComponentTreeNodeSchema> & {
+  children: ComponentTreeNode[];
+};
+
+const ComponentTreeNodeSchema: z.ZodType<ComponentTreeNode> = baseComponentTreeNodeSchema.extend({
+  children: z.lazy(() => ComponentTreeNodeSchema.array()),
+});
```
Chaoste, line 84: [Aligning the casing for schema variables with a capital letter]
I guess the "base" schema is a workaround as recursiveness doesn't play nicely with zod and type inference? If that's true, I added a comment already to make this clear.
```suggestion
// Use helper schema to define a recursive schema with its type correctly below
const BaseComponentTreeNodeSchema = z.object({
  definitionId: z.string(),
  variables: z.record(propertyKeySchema, ComponentPropertyValueSchema),
});

export type ComponentTreeNode = z.infer<typeof BaseComponentTreeNodeSchema> & {
  children: ComponentTreeNode[];
};

const ComponentTreeNodeSchema: z.ZodType<ComponentTreeNode> = BaseComponentTreeNodeSchema.extend({
  children: z.lazy(() => ComponentTreeNodeSchema.array()),
});
```

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+    if (q2 === '*') {
+      return 1;
+    }
+
+    const q1PixelValue = parseInt(q1.replace(/px|<|>/, ''));
+    const q2PixelValue = parseInt(q2.replace(/px|<|>/, ''));
+
+    return q1PixelValue > q2PixelValue ? -1 : 1;
+  });
```
Chaoste, line 133: [nitpick] we could extract the values already upfront and then only sort them.
```suggestion
  // Extract the queries boundary by removing the special characters around it
  const queries = value.map((bp) => bp.query === '*' ? bp.query : parseInt(bp.query.replace(/px|<|>/, ''));
  // sort updates queries array in place so we need to create a copy
  const originalQueries = [...queries];
  queries.sort((q1, q2) => {
    if (q1 === '*') {
      return -1;
    }
    if (q2 === '*') {
      return 1;
    }
    return q1 > q2 ? -1 : 1;
  });
```

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,186 @@
...
+    type: z.literal('ComponentValue'),
+    key: z.string(),
+  }),
+]);
+
+export type ComponentPropertyValue = z.infer<typeof ComponentPropertyValueSchema>;
+
+const BreakpointSchema = z.object({
+  id: z.string(),
+  query: z.string(),
```
Chaoste, line 61: In the refinement, we require a specific format for this string. Can we specify this in here maybe with a template string or regex?
e.g. TS notation for template literals: ```'*' | `<${number}px` ```

packages/validators/test/fixtures/v2023_09_28/experience.ts
Chaoste, line 1: We usually store fixtures in many projects in a folder called `__fixtures__`. It would be great if we follow that pattern. Example see here the [docs](https://github.com/contentful/user_interface/blob/a7e8a525d0742ddff675d6bc90a69c46823a7edf/docs/guides/testing-jest.md#L67) on user interface

Chaoste: We actually call it like that already in the other packages.

packages/validators/test/v2023_09_28/experienceValidator.spec.ts
```diff
@@ -0,0 +1,708 @@
...
+      });
+
+      (
+        [
+          ['definitionId', 'string'],
+          ['variables', 'object'],
+          ['children', 'array'],
+        ] as const
+      ).map(([attribute, type]) =>
+        it(`fails if '${attribute}' is missing`, () => {
```
Chaoste, line 183: In jest, such lists of test are usually defined via `it.each` - see the docs [here](https://jestjs.io/docs/api#testeachtablename-fn-timeout).

packages/validators/test/v2023_09_28/experienceValidator.spec.ts
Chaoste, line 1: I love how extensively we're testing the validators :)

Just one note: 708 LOC in one file is quite hard to jump right into, e.g. if someone else needs to adjust the later one. Splitting by describe blocks into two or three files might make it easier to navigate to the actual test case that someone wants to check out / change.

Chaoste (changes requested):
This is really great to see. As creating a whole new module is a good chance to make things right, I added quite a lot of (partially tentative) feedback/ suggestions.

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,186 @@
+import { z } from 'zod';
+import { SchemaVersions } from '../schemaVersions';
+
+const propertyKeySchema = z.string().regex(/^[a-zA-Z0-9-_.]{1,64}$/);
```
DanweDE, line 4: Can we restrict this further to what [nanoid](https://github.com/ai/nanoid) uses? I know this is the regex for CF entity sys IDs, but I would rather reserve characters like `.` for now, unless we already generate IDs using it in the app today. A smaller key length should be sufficient as well, considering these IDs serve the purpose of a simple lookup within the same doc.

```suggestion
const propertyKeySchema = z.string().regex(/^[A-Za-z0-9_-]{1,21}$/);
```

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,186 @@
...
+  z.object({
+    type: z.literal('ComponentValue'),
+    key: z.string(),
+  }),
+]);
+
+export type ComponentPropertyValue = z.infer<typeof ComponentPropertyValueSchema>;
+
+const BreakpointSchema = z.object({
+  id: z.string(),
```
DanweDE, line 60: can we use the same regex as for `propertyKeySchema` here?

packages/validators/test/fixtures/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,305 @@
...
+              cfHorizontalAlignment: {
+                type: 'DesignValue',
+                valuesByBreakpoint: {
+                  desktop: 'center',
+                },
+              },
+              cfMargin: {
+                type: 'DesignValue',
+                valuesByBreakpoint: {
+                  desktop: '0 auto 0 auto',
```
DanweDE, line 108: I didn't see any design value with token values, would be nice to have some in the fixtures, even though we don't do any special validation against those right now.

packages/components/package-lock.json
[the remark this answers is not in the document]
SofiaMargariti: Yes, I'll remove it

packages/core/src/types.ts
```diff
@@ -1,13 +1,32 @@
...
+  ComponentTree,
+} from '@contentful/experiences-validators/types';
+// TODO: Remove references to 'Composition'
+export {
+  DataSource as CompositionDataSource,
+  UnboundValues as CompositionUnboundValues,
+  ComponentSettings as ExperienceComponentSettings,
+  ComponentPropertyValue as CompositionComponentPropValue,
+  ComponentTreeNode as CompositionNode,
+  PrimitiveValue as CompositionVariableValueType,
```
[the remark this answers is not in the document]
SofiaMargariti: Good suggestion, I am updating the file to mark them as deprecated.
I was also debating whether I should use a prefix here, I think you have a good point about `TreeNode` so I'll add the prefix.

packages/validators/.gitignore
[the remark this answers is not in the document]
SofiaMargariti: Yes, was probably autogenerated.

packages/validators/src/schemas/v2023_09_28/experience.ts
SofiaMargariti: @DanweDE I am using this schema for both the user specified property names (variables) and the nanoid ids. Limiting to 21 is short for user property names, no? I could define two different Schemas here if we want to be more strict.

I agree about removing the `.`, so I'll remove that.

packages/validators/src/schemas/v2023_09_28/experience.ts
SofiaMargariti: Yes, good callout

SofiaMargariti:
@Chaoste @DanweDE thank you for the review, I believe I have addressed all the comments, so feel free to have another look.

packages/validators/src/schemas/v2023_09_28/experience.ts
SofiaMargariti: In the end, I added two different regex validations, one for nanoid ids (21 chars) and one for property names (64 chars), but let me know if you have any more comments.

packages/validators/src/schemas/schemaVersions.ts
```diff
@@ -0,0 +1,4 @@
+import { z } from "zod";
+
+export const SchemaVersions = z.union([z.literal('2023-09-28'), z.literal('2023-06-27'), z.literal('2023-07-26'), z.literal('2023-08-23')]);
+export type SchemaVersions = z.infer<typeof SchemaVersions>;
```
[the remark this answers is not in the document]
SofiaMargariti: I am reexporting that type in `core`, so `experience-builder-sdk/src/constants` already uses the new type.

I wanted to keep `core` as the main source of types so we don't have to import types from two different places.

Chaoste (approved):
Amazing work! Thanks for incorporating all the feedback and making this truly a great contribution :)

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,185 @@
+import { z } from 'zod';
+import { SchemaVersions } from '../schemaVersions';
+
+const uuidKeySchema = z.string().regex(/^[a-zA-Z0-9-_]{1,21}$/);
+const propertyKeySchema = z.string().regex(/^[a-zA-Z0-9-_]{1,64}$/);
```
DanweDE, line 5: is there a real need for allowing 64 here, or can we go with 21 or 32 here as well?

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -0,0 +1,185 @@
...
+export type ComponentTreeNode = z.infer<typeof BaseComponentTreeNodeSchema> & {
+  children: ComponentTreeNode[];
+};
+const ComponentTreeNodeSchema: z.ZodType<ComponentTreeNode> = BaseComponentTreeNodeSchema.extend({
+  children: z.lazy(() => ComponentTreeNodeSchema.array()),
+});
+
+const ComponentSettingsSchema = z.object({
+  variableDefinitions: z.record(
+    z.string().regex(/^[a-zA-Z0-9-_]{1,86}$/), // Here the key is <variableName>_<nanoid_id> so we need to allow for a longer length
```
DanweDE, line 88: Are we already imposing any limits in the SDK, on custom components' variable definition?

If not, we should probably look into enforcing it at registration time already, to not end up violating anything indirectly as a consequence.

I think 16 or 32 would be a reasonable variable key length to start with.

DanweDE (approved):

packages/validators/src/schemas/v2023_09_28/experience.ts
SofiaMargariti: There is no real need for 64 char, we can go with 32 and we can relax this later if we see a need.
We don't have yet any enforcement on registration time, I'll create a small task to add that.

DanweDE: Thank you!
This makes me even think that in the long run, we might want to reconsider the `<variableName>_<nanoid_id>` convention, there's a recursion problem if we think pattern and design system components should be interchangeable in principle :thinking:

PR #351, opened 2024-02-07
chore: use LTS node v20.11 []

Chaoste:
The pipeline is failing as the latest semantic-version release required using the LTS version of node.
So let's switch to the LTS version! :)

https://github.com/contentful/experience-builder/actions/runs/7818433260/job/21328900644


elylucas (approved):
Bumping the LTS looks good. We can probably drop semantic-release from the sdk package as lerna is used to bump the versions now.

Chaoste:
> Bumping the LTS looks good. We can probably drop semantic-release from the sdk package as lerna is used to bump the versions now.

Can we actually just turn down the whole circleci integration? I think we still trigger CircleCI for no reasons

PR #354, opened 2024-02-08
chore: publish packages to github npm registry []

Chaoste:
In the web app, I'm not able to pull any version of `core` that was released in the last week. As we moved all private packages to `npm.pkg.github.com`, it looks like we need to move or public ones as well. According to the [docs](https://docs.npmjs.com/cli/v8/using-npm/registry#description), `.npmrc` does only allow defining a registry per namespace. In the user interface, we point the whole Contentful namespace to `npm.pkg.github.com` which forces us to move experience-builder to this registry as well.

I hope that mirroring works correctly so that customers can still install via the npm registry. We already have a similar config for almost all our customer-facing packages - see [code search](https://github.com/search?type=code&auto_enroll=true&q=org%3Acontentful+publishConfig+language%3AJSON&l=JSON). So I'm confident that this is the way to go.

# Update
Ely and I aligned on this and cleared up all questions. All packages in the `@contentful` scope have to be on `npm.pkg.github.com`. We have a custom script that mirrors releases to npmjs **once per hour**.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -2,16 +2,19 @@
   "name": "@contentful/experience-builder-components",
   "version": "0.0.2-alpha.23",
   "description": "A basic set of components to use with Experience Builder",
-  "homepage": "https://github.com/contentful/experience-builder-toolkit/tree/next/packages/components#readme",
+  "homepage": "https://github.com/contentful/experience-builder/tree/next/packages/components#readme",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/contentful/experience-builder-toolkit.git"
+    "url": "git+https://github.com/contentful/experience-builder.git"
   },
   "main": "./dist/index.js",
   "module": "./dist/index.js",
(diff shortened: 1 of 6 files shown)


elylucas:
@Chaoste , why is this needed? All our packages are published to NPM and are publicly available.

SofiaMargariti:
I see in some packages (forma36, rich-text) we have `"access": "public",`. Do we want the same here?
If we don't include this, that means that customers can only install from npmjs and only we, internally, can install from GH registry, is that correct?

Chaoste:
> I see in some packages (forma36, rich-text) we have `"access": "public",`. Do we want the same here? If we don't include this, that means that customers can only install from npmjs and only we, internally, can install from GH registry, is that correct?

I checked the npm docs and they state that `access` is by default `public`. So that's why I left it out.

SofiaMargariti (approved):
👍

PR #356, opened 2024-02-09
feat: clicking outside of the edited canvas sends post message to user interface

ryunsong-contentful:
## Purpose
Send post message to ui so that we know when a user clicks outside the canvas but in the iframe

- This is particularly useful for deselecting components (https://github.com/contentful/user_interface/pull/19646)


packages/visual-editor/src/components/RootRenderer/RootRenderer.tsx
```diff
@@ -32,11 +34,27 @@ export const RootRenderer: React.FC<Props> = ({ onChange }) => {
     if (onChange) onChange(tree);
   }, [tree, onChange]);
 
+  const wrapperRef = useRef<HTMLDivElement>(null);
```
elylucas, line 37: Whats wrapperRef doing? It seems likes its only used to check if the ref exists, but it seems like the div its attached to will always be there as its not conditionally rendered.

elylucas (approved):
had a question around the wrapperRef, but besides that nothing to hold this back. Nice job, the UI feels better already.

PR #367, opened 2024-02-12
feat: [SPA-1778] assembly prop selection working

loweisz:
Partner PR that should go live before this: https://github.com/contentful/user_interface/pull/19662

## Summary:
This PR adds the functionality to select individual elements from assemblies that also select the assembly, but also the individual element. 

This allows us to display in the web application the prop selector for the individual element and not for the entire assembly. 

## Approach

Previously, the assemblies had an Id that was selected with each click. The individual elements were rendered with a random id that was generated each time the assembly was rendered, so using this id was not possible. 

I changed that the id of each element is not deterministic and follows the approach of `${assemblyInstanceId}---${pathToTheElementInTheAssemblyTree}`. 

**Note:** I had to use `---` and `_` because we also use them in CSS classes. 

https://github.com/contentful/experience-builder/assets/34910067/25eb7de4-18d9-4c58-ae63-73d1770c8db9


packages/visual-editor/src/utils/assemblyUtils.spec.ts
```diff
@@ -172,6 +190,11 @@ describe('resolveAssembly', () => {
     const node: CompositionComponentNode = {
       type: ASSEMBLY_NODE_TYPE,
       data: {
+        assembly: {
+          componentId: 'random-node-id',
```
Spring3, line 194: I see the same ids used in the entire fixture. it would be good to actually randomize them to make sure that in future the fixture will better reflect the actual data

Spring3 (approved):
Works when tested.

The only weird issue I've noticed was the fact that the overlay duplicated the name of the component, which I think is not related to this PR

Screenshot 2024-02-14 at 17 41 55

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
```diff
@@ -59,21 +60,25 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
...
+        sendMessage(OUTGOING_EVENTS.ComponentSelected, {
+          nodeId: node.data.id,
+        });
+        return;
+      }
+
+      setSelectedNodeId(node.data.id);
       sendMessage(OUTGOING_EVENTS.ComponentSelected, {
-        nodeId,
+        assembly: node.data.assembly,
```
Chaoste, line 80: it looks like we now send this not only for assembly blocks but also normal nodes. Is this intended as `assembly` will be undefined in the latter scenario? If yes, then we can remove the unused variable `isAssemblyBlock` if I'm not mistaken.

packages/visual-editor/src/utils/assemblyUtils.spec.ts
```diff
@@ -64,6 +67,11 @@ describe('deserializeAssemblyNode', () => {
       type: ASSEMBLY_NODE_TYPE,
       parentId: 'root',
       data: {
+        assembly: {
```
Chaoste, line 70: [nitpick]
This test file is just a collection of dumb snapshot tests and not helpful at all to maintain our code or to understand what the util function is doing. You can tell that your new feature just require adding similar values all over the file without actually adding a separate test case that describes an tests any new behaviour that we added. Unit tests should test behaviour not snapshots. I hope we find the time soon to rewrite this entire file and make it right :(

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
loweisz: in line 71 we send the event without the assembly, so there is a difference between when it is an assembly block or not. Or am missunderstanding it? 🤔

packages/visual-editor/src/utils/assemblyUtils.spec.ts
loweisz: I'm not sure I understand the problem here, this id is technically only used once for every test case and just appears multiple times within the data as we just reference this element in the deeper children as well. 

Where would you see a case where we should use difference ids?

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
Chaoste: In line 71, we send the event when the node itself it an `assembly`, here it is either an `assembly block` or just a normal node. I was wondering about this third case.

PR #368, opened 2024-02-12
fix: adjust auto to Auto so that the design tab is consistent throughout

ryunsong-contentful:
## Purpose
For the margin to change to Auto instead of auto with the container component
Screenshot 2024-02-12 at 11 11 44 AM

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -315,7 +315,7 @@ export const containerBuiltInStyles: Partial<
     type: 'Text',
     group: 'style',
     description: 'The margin of the container',
-    defaultValue: '0 auto 0 auto',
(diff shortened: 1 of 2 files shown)


Chaoste (approved):
Looks indeed better in the UI:)

Today I learned that most CSS property values are case-insensitive.

> Properties: Property names defined by the W3C standards are case-insensitive. So, background-color and BackGround-COLOR are considered the same. That being said, it is recommended to use lowercase for property names to maintain readability and consistency.
> Property Values: In general, property values are case-insensitive, except for font names and URLs. So, red and RED are considered the same color. However, the values of the font-family, url, and attr are case-sensitive.
[Source](https://benjamincrozat.com/css-property-names-values-case-sensitive)

PR #370, opened 2024-02-13
fix(experience-builder-sdk): support ssr in nextjs pages router

elylucas:
This PR fixes some issues and introduces a workaround to get SSR working in NextJS projects that use the pages router (app router is still not supported in SSR).

There were a couple of issues to tackle here.

1) The Experience that gets returned from the fetchBy{slug/id} methods returned an EntityStore object that cannot be serialized by NextJS in `getServerSideProps` due to their strict JSON serialization rules (see [here](https://github.com/vercel/next.js/issues/11993)). In order to get around this, we need to serialize the experience ourselves in getServerSideProps and then reconstruct the object back into a proper class in the component. To do the later, I reintroduced the `createExperience` function back into userland. It now takes an overloaded parameter of a json package, and if thats passed in, we recreate the experience based on the existing json object.

2) Defaulting the mode to editorMode by default when in an iframe was causing a hydration issue when in the editor. 
This as due to the server thinking it was not in editorMode (because there is no window object to check if you are in an iFrame), and the client thinking it was.

To get around this, We now default editorMode to false on first render. This will send an initial pageload of the page (via ssr) in non editor mode, and the client will initially think its not an editor mode as well. But then will quickly get updated to editor mode on the client after it mounts.

### DIFF
--- a/packages/core/src/entity/EntityStore.ts
+++ b/packages/core/src/entity/EntityStore.ts
@@ -14,14 +14,32 @@ export class EntityStore extends EntityStoreBase {
   private _experienceEntry: Composition | undefined;
   private _unboundValues: CompositionUnboundValues | undefined;
 
-  constructor({ experienceEntry, entities, locale }: EntityStoreArgs) {
-    super({ entities, locale });
+  constructor(json: string);
+  constructor({ experienceEntry, entities, locale }: EntityStoreArgs);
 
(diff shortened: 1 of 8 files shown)


Spring3 (approved):
🎉

primeinteger (approved):
Clean work!  This looks great

PR #371, opened 2024-02-13
feat(components): column locking [ALT-357]

chasepoirier:
## Purpose
Adjusted component definitions to be booleans for column wrap / column lock props.


primeinteger (approved):
🔥

ryunsong-contentful (approved):

PR #372, opened 2024-02-13
fix(components): line breaks for text built-ins [ALT-307]

chasepoirier:
## Purpose
Add css properties so text components respect line breaks

Screenshot 2024-02-13 at 2 32 15 PM

### DIFF
--- a/packages/components/src/components/Heading/Heading.css
+++ b/packages/components/src/components/Heading/Heading.css
@@ -1,0 +1,3 @@
+.cf-heading {
+  white-space: pre-line;
+}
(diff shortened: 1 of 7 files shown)


primeinteger (commented):
I think it's a great solution but we recently removed a bunch of components css trying to eliminate the need for customers having to import styles to compliment the SDK.

Should this be an inline style in the components instead of adding css files?

chasepoirier:
> I think it's a great solution but we recently removed a bunch of components css trying to eliminate the need for customers having to import styles to compliment the SDK.
> 
> Should this be an inline style in the components instead of adding css files?

Good call out, @elylucas also mentioned this. What do ya'll think of how it's handled now in the latest PR?

primeinteger (approved):
This looks good to me! Better than my inline styles idea 😝

elylucas (approved):

PR #374, opened 2024-02-14
fix: handle colliding unbound value keys correctly in entity store [SPA-1836]

Chaoste:
## Purpose

When creating a new pattern out of a container, the custom unbound values would be persisted correctly and rendered in the editor. However, in preview mode, it would only render the default value.

This as our logic puts all unbound values from experience and assembly definitions in a single entity store. If the keys are the same, the assembly ones override the experience ones. As the former one contains only default values while the latter ones contains actual instance values, I changed the order in which we merge both maps. I added a comment to it as well in case, future us is wondering why.

## Approach

Also, I realized that we have some fallbacks like `entityStore?.unboundValues ?? unboundValues` which were not necessary. This is a leftover from the merged preview/ editor rendering. Now we only support preview mode and know that the entity store is always defined. `entityStore?.unboundValues` and `unboundValues` are the exact same object. The same holds for `breakpoints`, `usedComponents`, and `dataSource`. As we only pass the `entityStore`, I removed all the unnecessary additional props.

### DIFF
--- a/packages/core/src/entity/EntityStore.ts
+++ b/packages/core/src/entity/EntityStore.ts
@@ -53,16 +53,24 @@ export class EntityStore extends EntityStoreBase {
     return this._experienceEntry?.usedComponents ?? [];
   }
 
-  public updateUnboundValues(unboundValues: CompositionUnboundValues) {
-    this._unboundValues = { ...(this._unboundValues ?? {}), ...unboundValues };
+  /**
+   * Extend the existing set of unbound values with the ones from the assembly definition.
+   * When creating a new assembly out of a container, the unbound value keys are copied and
+   * thus the existing and the added ones have colliding keys. In the case of overlapping value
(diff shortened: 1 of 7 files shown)


Spring3 (approved):
Worked well when I checked it locally. 

Probably unrelated to changes in this PR, but once I converted the column on the left into an assembly, the styles got messed up making the column shrink in width.


Screenshot 2024-02-14 at 16 33 02

As you can see below, the instance of such component seems to not have the same sizing issue

Screenshot 2024-02-14 at 16 33 59

Chaoste:
> Could you confirm that it is unrelated to changes in this PR?
Thanks @Spring3 for testing. I also just tried this and was able to reproduce it on production. Further, it looks like this does not only happen for the initially created assembly node but for all assembly nodes. It seems to be a more general bug that SingleColumn doesn't work properly as assembly node yet. I will inform Altitude about this.

PR #385, opened 2024-02-16
refactor: rename experience-builder-visual-editor to experiences-visual-editor-react

ryunsong-contentful:
## Purpose
Rename experience builder sdk packages based on https://contentful.atlassian.net/wiki/spaces/PROD/pages/4480499829/Experience+Builder+package+rename+proposal

This PR in particular converts experience-builder-visual-editor to experiences-visual-editor-react

## Approach
Create feature branch called experience-release based on the development branch.
Ignored the `CHANGELOG.md` and `package-lock.json` files as those should be auto regenerated

### DIFF
--- a/packages/visual-editor/package.json
+++ b/packages/visual-editor/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@contentful/experience-builder-visual-editor",
-  "version": "0.0.2-alpha.26",
+  "name": "@contentful/experiences-visual-editor-react",
+  "version": "0.0.1-alpha.25",
   "type": "module",
   "main": "./dist/index.js",
   "module": "./dist/index.js",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2898,10 +2898,6 @@
       "resolved": "packages/storybook-addon",
       "link": true
     },
-    "node_modules/@contentful/experience-builder-visual-editor": {
-      "resolved": "packages/visual-editor",
-      "link": true
-    },
     "node_modules/@contentful/experiences-components-react": {
       "resolved": "packages/components",
       "link": true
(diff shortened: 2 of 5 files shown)


packages/visual-editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@contentful/experience-builder-visual-editor",
-  "version": "0.0.2-alpha.26",
+  "name": "@contentful/experiences-visual-editor-react",
+  "version": "0.0.1-alpha.25",
```
ryunsong-contentful, line 3: Guess we didn't check all the packages to find the latest version. The rest are on `alpha.25` do I need to have an extra PR at the end so we can backtrack and update all the package.json's to have `alpha.26`?

Chaoste: The alpha version is defined per package and went out of sync when only certain but not all packages were updated. This makes sense as when you only change something in the visual-editor you don't want to create a new release for `core`.
To avoid starting the new package with `alpha.25`, I suggest starting here again with `0.0.1-alpha.1`. As we do this for all packages, they would be at least for now in sync again :)

Chaoste (changes requested):

ryunsong-contentful:
@Chaoste This is something @elylucas and I discussed, we decided to reference some version off what we have currently in development. Hopefully we can just move forward for now and get this merged into the [experience-release] feature branch and then once I cut a PR for the entire [experience-release] feature branch into development, we can make a final call about making all the package.json versions a specific version that's all the same.

Chaoste:
> @Chaoste This is something @elylucas and I discussed, we decided to reference some version off what we have currently in development. Hopefully we can just move forward for now and get this merged into the [experience-release] feature branch and then once I cut a PR for the entire [experience-release] feature branch into development, we can make a final call about making all the package.json versions a specific version that's all the same.

Gotcha, will unblock this PR. I'm still strongly in favor of `alpha.1`. So I'd really like to understand why you initially went this way. Let's discuss this again in the PR that would merge this into `development` before it's too late to make a call on this.

ryunsong-contentful:
Hey @Chaoste, 100% agreed, in the dev-sync we decided to go with `alpha.1` tentatively for now. But I will make a slack post and ensure that the big PR against development will have eyes from both teams before merging as that was my original intent anyways

PR #397, opened 2024-02-19
Media asset changes

elylucas:


packages/core/src/types.ts
```diff
@@ -324,3 +324,21 @@ export type RequestedEntitiesMessage = {
...
+  | Record<string, AssetFile | undefined>
+  | Document
+  | OptimizedImageAsset
+  | undefined;
+
+export type OptimizedImageAsset = {
+  url: string;
+  srcSet?: string[];
+  sizes?: string;
+  file: AssetFile;
```
DanweDE, line 343: This might be a `string` though, if a specific URL was provided manually.
I would lean towards making this a generic `originalAsset` or `source` - hope they are not too ambiguous:

```suggestion
  originalAsset: {
    type: "ContentfulAssetFile" | "string" // could contain specific DAMs later
    value: AssetFile | string // should effectively be treated as `any`
  };
```

packages/core/src/types.ts
```diff
@@ -324,3 +324,21 @@ export type RequestedEntitiesMessage = {
...
+  | AssetFile
+  | Record<string, AssetFile | undefined>
+  | Document
+  | OptimizedImageAsset
+  | undefined;
+
+export type OptimizedImageAsset = {
+  url: string;
+  srcSet?: string[];
+  sizes?: string;
```
DanweDE, line 342: shouldn't `sizes` also be an array - accounting for different breakpoints - that we join?

packages/core/src/utils/transformers/transformImageAsset.ts
```diff
@@ -0,0 +1,67 @@
...
+  details: Required<AssetFile['details']>;
+}
+type ValidFormats = (typeof SupportedImageFormats)[number];
+
+const MAX_WIDTH_ALLOWED = 4000;
+
+export const transformImageAsset = (
+  file: AssetFile,
+  sizes?: string,
+  quality: number = 100,
```
DanweDE, line 15: Should this be an `options` object, where `quality` is just one possible factor?

packages/core/src/utils/transformers/isFormatSupported.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!canvas.toDataURL) return false;
+  const imageData = canvas.toDataURL('image/webp');
+  if (imageData.startsWith('data:image/webp')) {
+    return true;
+  }
+}
+
+//Most browsers don't support avif exporting in canvas yet, so have to use another method for avif
+async function supportsAvif() {
+  if (typeof window === 'undefined' || !window.createImageBitmap) return false;
```
DanweDE, line 36: This means if we do SSR, we only support these once hydration kicks in :thinking:

Another way could be [progressive enhancement based](https://darekkay.com/blog/avif-images/#progressive-enhancement) on the `picture` tag. But we would have to account for it in the `OptimizedImageAsset` then.

elylucas: ya its a good point, and I don't think this will be fully implemented for the GA release. I had to put it on the back burner.

packages/core/src/types.ts
DanweDE: I was thinking that whenever possible we should be utilizing the inherent responsiveness of HTML and CSS, as they are essentially breakpoint-aware by design (which can't be said about React props).

This goes somewhat hand-in-hand with the best possible support for SSR. If `sizes` change dynamically upon hydration, we're not leveraging the above to the fullest.

elylucas: I think there is some confusion on what the `sizes` variable here does. End the end, it populates the `sizes` attribute on the `img` tag. Users set this value on the design tab, and they are able to control which value gets passed in via our already existing breakpoint system. So, for instance, they can set different sizes for each of our breakpoints (desktop, tablet, and mobile).

DanweDE: Yes, I think I have a pretty good understanding of it.
My point is simply that whenever possible we should utilize media queries and create static HTML (provide a wholistic image tag `sizes` attribute value that doesn't have to be updated), rather than rely on stateful React (update `sizes` depending on the current screen size).

The `sizes` **variable** would function the same way - like you said, it uses the breakpoint system - but when actualy rendering via React, we would take all the breakpoint values and generate a single `sizes` combining provided sizes for all breakpoints for the image tag.

Anyways, probably not super important and something we could optimize later still.

PR #398, opened 2024-02-20
feat: Convert zod to contentful errors [SPA-1701]

SofiaMargariti:
## Purpose

Create mapping between zod errors and contentful errors. Also updated the test to check for the new format. For the mapping I am following closely what we're doing for Ajv error mapping [here](https://github.com/contentful/content-stack/blob/master/packages/errors/lib/ajv-to-contentful-error/index.js).

Now the validator function returns a list of errors in the format expected in our management API.

See also https://github.com/contentful/content_api/pull/9359 that integrates this validator into CMA.

In addition this PR includes a couple smaller changes:
* Compile the source code to commonJS format as content_api doesn't work with ESM.
* Don't rely on the caller to provide the schema version, but read it from the experience itself.
* Add `strict` option in a some of the schema to avoid unexpected keys
* Switch ComponentPropertyValueSchema to discriminatedUnion for clearer error messages.

### DIFF
--- a/packages/validators/package.json
+++ b/packages/validators/package.json
@@ -25,7 +25,8 @@
   "exports": {
     ".": {
       "import": "./dist/index.js",
-      "require": "./dist/index.js",
+      "require": "./dist/index.cjs",
+      "default": "./dist/index.js",
       "types": [
         "./dist/index.d.ts"
       ]
--- a/packages/validators/rollup.config.mjs
+++ b/packages/validators/rollup.config.mjs
@@ -1,18 +1,26 @@
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import typescript from '@rollup/plugin-typescript';
 import dts from 'rollup-plugin-dts';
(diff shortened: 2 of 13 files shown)


Chaoste (approved):
Just by observing the code, this looks fine to me. I didn't compare this with the existing error logic in our backend code but I guess this aligns with a similar type that is defined on the Content API side.

SofiaMargariti:
> Just by observing the code, this looks fine to me. I didn't compare this with the existing error logic in our backend code but I guess this aligns with a similar type that is defined on the Content API side.

Both our backend and @contentful/errors have very loose definitions for this kind of errors details, e.g `Record<string, any>`. I am following the same format we're using when converting Ajv errors in our API responses ([example](https://github.com/contentful/content-stack/blob/master/packages/errors/lib/ajv-to-contentful-error/index.js#L53C1-L60C5)) but there isn't any type defined anywhere I could find.

PR #400, opened 2024-02-20
refactor: rename experience-builder to experiences-sdk-react

ryunsong-contentful:
## Purpose
Rename experience builder sdk packages based on https://contentful.atlassian.net/wiki/spaces/PROD/pages/4480499829/Experience+Builder+package+rename+proposal

This PR in particular converts `experience-builder` to `experiences-sdk-react`

## Approach
Create feature branch called experience-release based on the development branch.
Ignored the `CHANGELOG.md` and `package-lock.json` files as those should be auto regenerated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ tmp/
 
 # Ingore .env local files
 .env
+.npmrc
--- a/package-lock.json
+++ b/package-lock.json
@@ -179,12 +179,6 @@
         }
       }
     },
-    "node_modules/@auto-it/core/node_modules/tr46": {
-      "version": "0.0.3",
-      "resolved": "https://registry.npmjs.org/tr46/-/tr46-0.0.3.tgz",
-      "integrity": "sha512-N3WMsuqV66lT30CrXNbEjx4GEwlow3v6rr4mCcv6prnfwhS01rkgyFdjPNBYd9br7LpXV1+Emh01fHnq2Gdgrw==",
-      "dev": true
-    },
     "node_modules/@auto-it/core/node_modules/ts-node": {
       "version": "10.9.2",
       "resolved": "https://registry.npmjs.org/ts-node/-/ts-node-10.9.2.tgz",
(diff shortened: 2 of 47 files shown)


.github/workflows/build.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
           restore-keys: cypress-${{ runner.os }}-
       - name: Authenticate with private NPM package
         run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+        run: echo "@contentful:registry=https://npm.pkg.github.com" >> ~/.npmrc
+        run: echo "//npm.pkg.github.com/:_authToken=${{secrets.GH_TOKEN}}" >> ~/.npmrc
```
Chaoste, line 32: This whole setup should not be necessary. This is a public repository that should only use public dependencies. Public packages would therefore be published to npmjs. Similar to our actual users, we should only fetch our node modules via npmjs to not mix up things in here.

Please be mindful to manually turn the package to "public" in the npmjs registry though :)

Chaoste (commented):
Changes look good except for the github registry part

.github/workflows/build.yaml
ryunsong-contentful: Sounds good - I'm going to make sure we publish it under the correct name as when you look it up it doesn't exist https://github.com/orgs/contentful/packages?tab=packages&q=experiences. So I'll have to rename it

Chaoste: Note that we would still push this to github thanks to the `.npmrc` in this repo :) But this config here is only about installing modules which should still happen via npm

Chaoste: Btw, so many packages with `experience` ^^ https://github.com/orgs/contentful/packages?q=experiences&tab=packages&q=experience

elylucas (approved):

PR #407, opened 2024-02-21
fix: updating dev branch after release []

Chaoste:
The release pipeline recently failed with [this error](https://github.com/contentful/experience-builder/actions/runs/7933196824/job/21662811162). I compared the commit history of both branches but couldn't find the reason for that. Maybe, it's related to using different merge strategies on our PRs?
However, this ensure that these two branches don't get out of sync.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -70,9 +70,10 @@ jobs:
             git push origin
             git checkout development
             git rebase next
-            git push origin
+            git push origin -f
           elif [ ${{ github.ref_name }} = next ]; then
             git checkout development
             git rebase next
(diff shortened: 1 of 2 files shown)


elylucas (commented):
Wouldn't it be better to fix this manually when it happens than risk losing commits?

Chaoste:
I'm closing this as we still don't know how this can happen and how we can avoid it. I really recommend adding branch protection for this. @elylucas do you by any chance have already a ticket for this in your backlog as it sounded like you are already aware of the missing bot configuration that we would need for that?

elylucas:
We do: https://contentful.atlassian.net/browse/ALT-432

PR #408, opened 2024-02-21
fix: add padding to dropzones when dragging [ALT-308]

primeinteger:
## Purpose

Ticket: https://contentful.atlassian.net/browse/ALT-308

Adds `10px` padding to dropzones when the user is dragging on the canvas.
* This makes it easier to work with nested containers by allowing components to be dropped next to containers that normally fill out the entire space of a parent container.

* Only Container, Section, and Single Column components will be updated with extra padding when the user is dragging components and re-parenting on the canvas.

* If the structure component includes padding set by the user in the design sidebar, only sides with less than `10px` padding will be updated to `10px` (higher padding values remain as-is).

Simple example showing how nested sections have extra padding when dragging:

https://github.com/contentful/experience-builder/assets/8539634/cccec5d9-dde6-4d94-acc1-53b593bc865d

More complex layout showing how dropzones can be activated with different row and column orientations:

https://github.com/contentful/experience-builder/assets/8539634/35e97b7c-e117-4b41-bbae-357e815d9582


packages/core/src/utils/components.ts
```diff
@@ -1,17 +1,26 @@
...
+  CONTENTFUL_COMPONENTS.container.id,
+  CONTENTFUL_COMPONENTS.singleColumn.id,
+];
+
+export const isContentfulStructureComponent = (
+  componentId?: string,
+  excludeStructureIds?: string[],
+) => {
+  const searchIds = excludeStructureIds
+    ? allStructureComponents.filter((id) => !excludeStructureIds?.includes(id))
```
ryunsong-contentful, line 15: Just food for thought, I'd imagine there would only ever be a few dozen or so max structure components to lookup/exclude based on how many times a user will put on the canvas. However, if there's ever a scenario or edge case a user somehow uses thousands of components we should consider using a set for a faster lookup O(1) over the O(n) we have here.

ryunsong-contentful (approved):

primeinteger:
I'm thinking that we should wait until after @chasepoirier's #414 is merged, and then evaluate if this padding is useful/necessary

packages/core/src/utils/components.ts
primeinteger: That's a very good point

primeinteger:
Chase incorporated these changes into his branch in progress for ALT-504

PR #409, opened 2024-02-21
Rename Experience Builder Sdks to use the Experiences term

ryunsong-contentful:
## Purpose
1. Rename the following packages specified in https://contentful.atlassian.net/wiki/spaces/PROD/pages/4480499829/Experience+Builder+package+rename+proposal
2. Reset all package versions to "0.0.1-alpha.1", (this include the version for Validators package)
3. The only package that is an exception to the above is the CLI package called `create-experience-builder`

## Approach
If you do a `cmd + shift + f` in VSCode, you should only see the `experience-builder` term be used when beign linked in READMEs or package.jsons, code specific namings, or in the `catalog-info.yml` files. In essence the term `experience-builder` should only exist in non-sdk contexts. Any further renamings should be done as a separate unit of work.


packages/visual-editor/src/store/entityStore.ts
```diff
@@ -21,7 +21,7 @@ export const useEntityStore = create<EntityState>((set) => ({
   },
   resetEntityStore(locale, entities = []) {
     console.debug(
-      `[exp-builder.sdk] Resetting entity store because the locale changed to '${locale}'.`
+      `[exp-builder.sdk] Resetting entity store because the locale changed to '${locale}'.`,
```
primeinteger, line 24: Should all the log messages that contain `exp-builder.sdk` and `exp-builder.visual-editor` be updated too?

primeinteger (approved):
I approve. I checked out the branch locally to review and it's all looking great! I found a couple small things to ask questions about below but those don't seem too critical.

packages/storybook-addon/src/components/SectionStyleInput.tsx
```diff
@@ -1,12 +1,12 @@
 import React, { useCallback, useEffect, useState } from 'react';
-import { ComponentDefinitionVariable } from '@contentful/experience-builder';
+import { ComponentDefinitionVariable } from '@contentful/experiences-sdk-react';
 
-// import { SECTION_STYLE_ATTRIBUTE_KEY } from '@contentful/experience-builder/dist/types';
+// import { SECTION_STYLE_ATTRIBUTE_KEY } from '@contentful/experiences-sdk-react/dist/types';
```
Chaoste, line 4:
```suggestion
```

Chaoste (approved):
Looks good to me - found code in your PR diff that can be removed but not related to your PR.
Looking forward to rename the folders and packages in READMEs soon :)

PR #410, opened 2024-02-22
fix: increase property key length []

SofiaMargariti:
## Purpose

Property keys for imported components have a limit of 32 characters (to be implemented) while property keys for patterns have a limit of 54 characters (<32-char-variable-name>_<21-char-nanoid-id>). Because we cannot distinguish between the two in the componentTree, we will use the larger limit for both.

Note: we should reconsider how we create the property keys for patterns to align with imported components, this is a quick fix to unblock the validations work.

### DIFF
--- a/packages/validators/src/schemas/v2023_09_28/experience.ts
+++ b/packages/validators/src/schemas/v2023_09_28/experience.ts
@@ -4,9 +4,15 @@ import { SchemaVersions } from '../schemaVersions';
 const uuidKeySchema = z
   .string()
   .regex(/^[a-zA-Z0-9-_]{1,21}$/, { message: 'Does not match /^[a-zA-Z0-9-_]{1,21}$/' });
+
+/**
+ * Property keys for imported components have a limit of 32 characters (to be implemented) while
+ * property keys for patterns have a limit of 54 characters (<32-char-variabl-name>_<21-char-nanoid-id>).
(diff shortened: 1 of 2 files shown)


packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -4,9 +4,15 @@ import { SchemaVersions } from '../schemaVersions';
...
+
+/**
+ * Property keys for imported components have a limit of 32 characters (to be implemented) while
+ * property keys for patterns have a limit of 54 characters (<32-char-variabl-name>_<21-char-nanoid-id>).
+ * Because we cannot distinguish between the two in the componentTree, we will use the larger limit for both.
+ */
 const propertyKeySchema = z
   .string()
-  .regex(/^[a-zA-Z0-9-_]{1,32}$/, { message: 'Does not match /^[a-zA-Z0-9-_]{1,32}$/' });
+  .regex(/^[a-zA-Z0-9-_]{1,54}$/, { message: 'Does not match /^[a-zA-Z0-9-_]{1,54}$/' });
```
DanweDE, line 15: yay, it's the [recursion problem](https://github.com/contentful/experience-builder/pull/348#discussion_r1491158528) :see_no_evil:

DanweDE (approved):

Spring3 (approved):

PR #412, opened 2024-02-22
chore: sync next with development branch [ci skip]

SofiaMargariti:
## Purpose

This branch attempts to sync next branch with development to resolve conflicts we're getting when trying to release a new version.


dimitrycf (approved):
LGTM

those are only package.json version bumps btw. I was expecting that in case we have syncing there would be more conflicts. But if that's all there is then great! :)

PR #413, opened 2024-02-22
fix: revise built-in component thumbnail urls [ALT-459]

primeinteger:
## Purpose

Clean up old references to thumbnail URL icons for built-in components.

New component icons are being implemented in the UI code via https://github.com/contentful/user_interface/pull/19882


lilbitner (approved):
Makes sense 👍 !

PR #414, opened 2024-02-23
fix(visual-editor): canvas usability and placeholders [ALT-449]

chasepoirier:
## Purpose
A significant refactor on the approach to how canvas interactions work. New functionality in the root `mousemove` event handler is used to recursively enable the most deeply nested dropzone in relation to the mouse position.

This allows the canvas to properly disable all dropzones that are not being moused over, including the parent dropzones to a nested dropzone which previously caused issues with reparenting.

In addition, hitboxes are used to enable specific dropzones based on the mouse position. Ex. moving the mouse between two components should allow the user to drop the component between the components instead of inside one.

With this refactored approach, there is no longer a need to add expanded styles during drag resulting in a more consistent canvas between the drag states and the non drag states.

The positioning of component indicators (placeholders) while dragging has also been refactored to consistently be placed in the proper spot instead of overlapping elements on the canvas.

Lastly, small UI changes were made while dragging components on the canvas to reduce clutter.

https://github.com/contentful/experience-builder/assets/23040747/25dd5227-ae6a-432d-9a1d-5f8dba71f3e3

## Related Bug Tickets that should be resolved with this refactor

- https://contentful.atlassian.net/browse/ALT-449
- https://contentful.atlassian.net/browse/ALT-338
- https://contentful.atlassian.net/browse/ALT-353
- https://contentful.atlassian.net/browse/ALT-482
- https://contentful.atlassian.net/browse/ALT-318
- https://contentful.atlassian.net/browse/ALT-405
- https://contentful.atlassian.net/browse/ALT-335


packages/components/src/components/ContentfulContainer/ContentfulContainer.css
```diff
@@ -14,12 +15,14 @@
 
 .cf-container-wrapper {
   position: relative;
+  /* outline-offset: -1px;
+  outline: 2px solid var(--exp-builder-gray400); */
```
primeinteger, line 19: Smash that <kbd>_Delete_</kbd> button

packages/components/src/components/ContentfulContainer/ContentfulContainer.tsx
```diff
@@ -33,7 +33,7 @@ export const ContentfulContainer: React.FC<ContentfulContainerAsHyperlinkProps>
     return renderDropzone(node, {
       ['data-test-id']: 'contentful-container',
       id: 'ContentfulContainer',
-      className: combineClasses(className, 'defaultStyles'),
+      className: combineClasses('contentful-container', className),
```
primeinteger, line 36: Good find, looks like this class reverted back to the old value by accident

packages/visual-editor/src/components/Draggable/Placeholder.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }
+  if (isSameZone && destinationIndex > sourceIndex) {
+    return { width, height, bottom: -height, left: -left };
+  }
+
+  if (isEnd) {
+    return { width, height, bottom: -height, left: -left };
+  }
+
+  return { width, height, top, left: -left };
```
primeinteger, line 114: I notice the returned objects are very similar in each branch here, so it could possibly be refactored to have a single return statement and conditional properties in the object like this:

```ts
const isNegativeTop = direction === 'horizontal' && (isSameZone && destinationIndex > sourceIndex || isEnd);
const isNegativeLeft = direction !== 'horizontal' && (isEmptyZone || isSameZone && destinationIndex > sourceIndex || isEnd);

const width = direction === 'horizontal' ? draggableSizes.width : dropzoneSizes.width;
const height = direction === 'horizontal' ? dropzoneSizes.height : draggableSizes.height;
const top = direction === 'horizontal' ? (dropzoneSizes.height - elementSizes.height) / 2 : -height;
const left = direction === 'horizontal' ? -width : (dropzoneSizes.width - elementSizes.width) / 2;

return {
  width,
  height,
  top: isNegativeTop ? -top : top,
  right: direction === 'horizontal' ? -width : undefined,
  bottom: direction !== 'horizontal' ? -height : undefined,
  left: isNegativeLeft ? -left : left,
};
```

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
```diff
@@ -148,41 +79,47 @@ export function Dropzone({
...
+      return isHoveringZone;
+    }
+
+    const draggingParentIds = getZoneParents(draggedSourceId || '');
+
+    if (!draggingParentIds.length) {
+      if (isRootZone) {
+        return true;
+      }
+      return false;
```
primeinteger, line 98: Since this returns either way, you could revise it to `return isRootZone`

packages/visual-editor/src/components/VisualEditorRoot.tsx
```diff
@@ -7,11 +7,27 @@ import { OUTGOING_EVENTS } from '@contentful/experience-builder-core/constants';
...
+const findNearestDropzone = (element: HTMLElement) => {
+  const zoneId = element.getAttribute(CTFL_ZONE_ID);
+
+  if (!element.parentElement) {
+    return '';
+  }
+
+  if (element.tagName === 'BODY') {
+    return '';
+  }
```
primeinteger, line 22: What do you think about `return null` instead of `return ''` on these early returns?  This may be a nit pick but `null` seems more appropriate for representing the absence of a value

packages/visual-editor/src/components/VisualEditorRoot.tsx
```diff
@@ -7,11 +7,27 @@ import { OUTGOING_EVENTS } from '@contentful/experience-builder-core/constants';
 import { useInitializeEditor } from '@/hooks/useInitializeEditor';
 import { useEntityStore } from '@/store/entityStore';
 import { useEditorStore } from '@/store/editor';
+import { useZoneStore } from '@/store/zone';
+import { CTFL_ZONE_ID } from '@/types/constants';
+
+const findNearestDropzone = (element: HTMLElement) => {
```
primeinteger, line 13: Looking at the branch locally, it looks like this function's return type is being inferred as `any` - should the return type be explicitly added?

```suggestion
const findNearestDropzone = (element: HTMLElement): string | null => {
```

packages/visual-editor/src/hooks/useEditorSubscriber.ts
```diff
@@ -281,6 +281,7 @@ export function useEditorSubscriber() {
         }
         case INCOMING_EVENTS.ComponentDragStarted: {
           dragState.updateIsDragStartedOnParent(true);
+          setDraggingOnCanvas(true);
```
primeinteger, line 284: Testing this out locally, I really like this change to trigger the drag sooner (as soon as the card is picked up)

packages/visual-editor/src/components/Dropzone/styles.module.css
```diff
@@ -9,20 +9,8 @@
   transition: outline 0.2s;
 }
 
-.isExpanded {
-  min-height: 80px;
-}
```
primeinteger, line 14: Testing locally, it looks like it can be a little difficult to activate the dropzones inside of containers that are not very tall due to size of elements within.

For example adding a single line text component in a container, the hitboxes above and below the container on the root want to activate, making it harder to drop another component in the same container as the text component:

https://github.com/contentful/experience-builder/assets/8539634/2185a8bf-4d9a-4e14-b7a1-543b88ff45bd

primeinteger (approved):
Awesome work on this!  I really like the solution you came up with to address reparenting and this addresses an impressive number of bugs at the same time 🥵 👏

packages/visual-editor/src/components/Dropzone/styles.module.css
chasepoirier: Nice catch! I've noticed some similar scenarios that I'm going to try and address today.

I'm going to try to add conditional logic to hitboxes to address this instead of expanding elements during a drag which lead to some unintended side effects with reparenting

chasepoirier:
Another enhancement added to this PR; the ability to drop components into the whitespace below the canvas. See video for how it works with a footer outside of the canvas.

https://contentful.atlassian.net/browse/ALT-487

https://github.com/contentful/experience-builder/assets/23040747/bf3a6c04-2456-4f73-9ec5-c7104d050fd9

packages/visual-editor/src/components/Draggable/Placeholder.tsx
chasepoirier: Cleannnnn 🔥

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
chasepoirier: Hah duh, this is funny now that I look at it again 🤣

packages/visual-editor/src/components/VisualEditorRoot.tsx
chasepoirier: Yeah good call here

packages/visual-editor/src/components/Dropzone/styles.module.css
chasepoirier: @primeinteger this will be done with a follow up ticket: https://contentful.atlassian.net/browse/ALT-504

PR #415, opened 2024-02-23
chore: add .npmrc file to gitignore

elylucas:
Git ignores a previously committed .npmrc file, which had a registry reference to install @contentful namespace packages from github. This would cause an issue for users outside of CTFL that don't have access to CTFL github packages and could accidentally install packages from github packages that were meant to be installed from npm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ yarn-debug.log*
 yarn-error.log*
 pnpm-debug.log*
 lerna-debug.log*
+.npmrc
 
 node_modules
 build
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-//npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}
-@contentful:registry=https://npm.pkg.github.com


ryunsong-contentful (approved):

Chaoste (changes requested):
Good call out. I wasn't aware of that when adding the file. I added it as we need it to publish new releases to the github registry. I would expect the CI to fail now in one of the release branch if this file was missing.
We could achieve this by adding the file inside the workflow step like [here](https://github.com/contentful/rich-text/blob/1fbcc367918c03ed6eea7bd9f36a859026ef8cfb/.circleci/config.yml#L19).

elylucas:
> Good call out. I wasn't aware of that when adding the file. Though, I added it as we need the config to publish new releases to the github registry. Probably, the CI would fail now in any of the release branch if this file was missing. We could make this work by adding the file inside the workflow step like [here](https://github.com/contentful/rich-text/blob/1fbcc367918c03ed6eea7bd9f36a859026ef8cfb/.circleci/config.yml#L19).

When you added the publishConfig to the package files should take care of publishing in CI: https://github.com/contentful/experience-builder/pull/354

Chaoste:
> > Good call out. I wasn't aware of that when adding the file. Though, I added it as we need the config to publish new releases to the github registry. Probably, the CI would fail now in any of the release branch if this file was missing. We could make this work by adding the file inside the workflow step like [here](https://github.com/contentful/rich-text/blob/1fbcc367918c03ed6eea7bd9f36a859026ef8cfb/.circleci/config.yml#L19).
> 
> When you added the publishConfig to the package files should take care of publishing in CI: #354

You're right that it already defines the correct registry but we also need to pass in the private github token with write access.

elylucas:
> > > Good call out. I wasn't aware of that when adding the file. Though, I added it as we need the config to publish new releases to the github registry. Probably, the CI would fail now in any of the release branch if this file was missing. We could make this work by adding the file inside the workflow step like [here](https://github.com/contentful/rich-text/blob/1fbcc367918c03ed6eea7bd9f36a859026ef8cfb/.circleci/config.yml#L19).
> > 
> > 
> > When you added the publishConfig to the package files should take care of publishing in CI: #354
> 
> You're right that it already defines the correct registry but we also need to pass in the private github token with write access.

Think you took care of that as well here: https://github.com/contentful/experience-builder/blob/development/.github/workflows/publish.yaml#L37

Im going to push this through and see what happens, as all our builds are failing due to it.

elylucas:
The publish step did fail on this PR, so I updated a new PR to get it working again: https://github.com/contentful/experience-builder/pull/441

Queenkaz (commented):
👏

PR #416, opened 2024-02-24
fix: update Image component to use existing cfHeight and cfWidth [ALT-488]

primeinteger:


primeinteger:
I'm going to switch back to using `cfImageWidth` and `cfImageHeight` after noticing the added complexity in the UI side of this. Having Image fiends separate from `cfWidth` and `cfHeight` will help with having different design fields for Image components as the behavior will be somewhat different than Structure component Height and Widths

PR #428, opened 2024-02-27
chore: bg image work and refactoring of transformers

elylucas:


packages/core/src/utils/transformers/media/transformMedia.ts
```diff
@@ -49,13 +55,15 @@ export const transformMedia = (
...
+        width as string,
         Number(quality),
         format as (typeof SUPPORTED_IMAGE_FORMATS)[number],
-      ).url;
+      );
     } catch (error) {
       console.error('Error transforming image asset', error);
     }
+  } else {
+    return getBoundValue(entityStore, binding, []);
```
primeinteger, line 66: I wonder adding early returns in the `if` and `else if` statements would improve readability.

Basically making the function body more like this psuedocode:
```ts
  if (variableName === 'cfImageAsset') {
    try {
      return getOptimizedImageAsset(...)
    } catch {
      console.error(...)
    }
  } else if (variableName === 'cfBackgroundImageUrl') {
    try {
      return getOptimizedBackgroundImageAsset(...)
    } catch {
      console.error(...)
    }
  }
  return getBoundValue(...)
  ```
  
  What are your thoughts?

File: packages/core/src/utils/transformers/media/getOptimizedImageUrl.ts [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+  quality?: number,
+  format?: string,
+) {
+  const params = new URLSearchParams();
+  if (width) {
+    params.append('w', width.toString());
+  }
+  if (quality && quality > 0 && quality !== 100) {
+    params.append('q', quality.toString());
+  }
```
primeinteger, line 13: Pass in quality = 200 for the top-secret CSI image enhancement feature ;)

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.ts
```diff
@@ -1,50 +1,37 @@
 import { SUPPORTED_IMAGE_FORMATS } from '@/constants';
 import { OptimizedBackgroundImageAsset } from '@/types';
+import { getOptimizedImageUrl } from './getOptimizedImageUrl';
 import { AssetFile } from 'contentful';
 
 interface AssetFileWithRequiredImage extends AssetFile {
   details: Required<AssetFile['details']>;
 }
```
primeinteger, line 8: BTW, I noticed there's an `ImageFile` type in the `contentful` package that includes the types for `details` and `details.image`

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { beforeEach, describe, it, expect } from 'vitest';
+import { getOptimizedBackgroundImageAsset } from './getOptimizedBackgroundImageAsset';
+import { AssetFile } from 'contentful';
+
+let file: AssetFile;
```
primeinteger, line 5: If you switch this to using `ImageFile` instead you can remove the `!` from `file.details.image!` a few times below

primeinteger (approved):
I noticed there's a `ImageFile` type in the contentful package that might be better than using `AssetFile` in some places.  At least where `AssetFileWithRequiredImage` is used would be better to use `ImageFile`.

This all looks great and I like all the unit tests you added and cleaning up the all the commented out code. Good stuff!

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.spec.ts
elylucas: Nice, I wasn't aware of `ImageFile`

packages/core/src/utils/transformers/media/transformMedia.ts
elylucas: Ya usually pretty good change

packages/core/src/utils/transformers/media/getOptimizedImageUrl.ts
elylucas: fine i'll add tests

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.ts
elylucas: lies

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.spec.ts
primeinteger: It turns out `ImageFile` in in a private types packages that's used in the web app.  Nevermind this idea

packages/core/src/utils/transformers/media/getOptimizedBackgroundImageAsset.ts
primeinteger: It turns out `ImageFile` in in a private types packages that's used in the web app.  Nevermind this idea

elylucas:
> I noticed there's a `ImageFile` type in the contentful package that might be better than using `AssetFile` in some places. At least where `AssetFileWithRequiredImage` is used would be better to use `ImageFile`.
> 
> Otherwise this all looks great and I like all the unit tests you added and cleaning up the all the commented out code. Good stuff!

Turns out ImageFile is not available in the `contentful` package, so sticking with what we got for now.

PR #430, opened 2024-02-27
fix: wrap assembly with correct container styles [SPA-1869]

Chaoste:
## Purpose

When an assembly contains more than one top-level container, those should be aligned vertically. Due to the DND-related element that wraps each assembly instance, they are instead positioned horizontally.

## Approach

While it's not the best solution, it works well rendering an assembly instance with `block` instead of `flex` and ensuring that the wrapper itself fills the whole width as the `assembly` node doesn't define an explicit width - it uses `display: contents` instead to forward the children styles (and pretend like it doesn't "exist").

Especially for the case of using an assembly instance inside a column, the height calculation would break as each top-level container/section inside the assembly defines `height: 100%` per default. To avoid these sizing bugs, we need to set the `height` to the actual default of `auto` for top-level nodes.

## Before

Screenshot 2024-02-27 at 16 41 29

## After

Screenshot 2024-02-27 at 16 41 47

## Pattern definition used for testing
It contains two top-level containers, one with width `100%` and one with `width: 50%`:

Screenshot 2024-02-27 at 16 40 39

### DIFF
--- a/packages/visual-editor/src/components/Dropzone/useComponentProps.ts
+++ b/packages/visual-editor/src/components/Dropzone/useComponentProps.ts
@@ -11,6 +11,7 @@ import {
   DESIGN_COMPONENT_NODE_TYPE,
   ASSEMBLY_NODE_TYPE,
   EMPTY_CONTAINER_HEIGHT,
+  ASSEMBLY_BLOCK_NODE_TYPE,
 } from '@contentful/experience-builder-core/constants';
 import type {
   StyleProps,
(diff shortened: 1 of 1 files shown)


chasepoirier (approved):
This looks good to me!

Only question; are we expecting changes to the Assembly styles in preview/delivery mode or only in editor mode?

It would be good to see how this example you have looks in preview mode just to ensure there is minimal difference between editor and preview mode.

Chaoste:
> This looks good to me!
> 
> Only question; are we expecting changes to the Assembly styles in preview/delivery mode or only in editor mode?
> 
> It would be good to see how this example you have looks in preview mode just to ensure there is minimal difference between editor and preview mode.

Thanks @chasepoirier for checking. As you can see in the code change, this will only affect the `visual-editor` as this is the only case where we render the DND wrapper.

Btw this is how it's rendered right now in preview mode. It's not perfectly the same as the column will center everything inside while that's not the default of the assembly definition.
Screenshot 2024-02-27 at 21 02 38

chasepoirier:
> > This looks good to me!
> > Only question; are we expecting changes to the Assembly styles in preview/delivery mode or only in editor mode?
> > It would be good to see how this example you have looks in preview mode just to ensure there is minimal difference between editor and preview mode.
> 
> Thanks @chasepoirier for checking. As you can see in the code change, this will only affect the `visual-editor` as this is the only case where we render the DND wrapper.
> 
> Btw this is how it's rendered right now in preview mode. It's not perfectly the same as the column will center everything inside while that's not the default of the assembly definition. Screenshot 2024-02-27 at 21 02 38

Got it, thanks the response! Food for thought; I think we should eventually look into the feasibility of removing the wrapping DnD element all together and passing the DnD props to the wrapping component of the element itself. That way we no longer have a wrapping div in editor mode

PR #433, opened 2024-02-28
fix: change default values for Image built in styles [ALT-488]

primeinteger:
## Purpose

Changing settings for some of the new Image component's built in styles


packages/core/src/definitions/styles.ts
```diff
@@ -164,26 +165,6 @@ export const builtInStyles: Partial<
     },
     description: 'hyperlink for section or container',
   },
-  cfImageFormat: {
```
elylucas, line 167: how come these went away?

elylucas (approved):
One question to perhaps educate me on the built in styles, but besides that lgtm!

packages/core/src/definitions/styles.ts
primeinteger: All the props in `builtInStyles` get added to all component definitions via `enrichComponentDefinition()` calling `applyBuiltInStyleDefinitions()`

Screenshot 2024-02-28 at 3 50 40 PM

With these props removed from `builtInStyles` we will no longer add these variables to all component definitions:

Screenshot 2024-02-28 at 3 52 37 PM

When the props are in `optionalBuiltInStyles`, they will only be added to components that specifically include them in the component definition array of `builtInStyles`

Screenshot 2024-02-28 at 3 56 11 PM

PR #434, opened 2024-02-28
fix(canvas-interactions): prevent draggables from flying off mouse cursor

chasepoirier:
## Purpose

By implementing the cloning API offered from the hello-pangea/dnd [re-parenting docs](https://github.com/hello-pangea/dnd/blob/main/docs/guides/reparenting.md), we can prevent unwanted css transforms from reaching the draggable component by reparenting the draggable to the body of the document.

## Side Quest
I ended up implementing a reparentChildNode fn in our smart tree update logic. With this enabled, causing a reparent event will instantly commit the changes to the tree instead of waiting for a postMessage from the user_interface with the updated tree. This removes component flickering when reparenting where the component was in it's old location until the update was received.

### Before:

https://github.com/contentful/experience-builder/assets/23040747/40c674b7-b837-4dd8-b3a0-9ad67c285867

### With cloning enabled
(The flickering dropzone indicators will be tackled in another ticket):

https://github.com/contentful/experience-builder/assets/23040747/e6118159-69ad-4fb7-9ff8-c9c7cdfd7d69

## Approach
Use `renderClone` as defined in the hello-pangea/dnd reparenting docs. This clones the draggable and appends it to the body of the document.

### Caveat
Implementing cloning results in the following warning appearing in the console when moving a component that has a nested dropzone.

Screenshot 2024-02-27 at 5 07 44 PM

The reasoning is because when we start a drag with cloning enabled, it removes the draggable from the component tree and appends it to the body instead. This technically changes the amount of `droppableIds` present in the tree which throws this warning. I've thoroughly tested and this has no affect on the actual experience since the droppableIds removed are part of the component being dragged but it does show up for us while developing locally.

### DIFF
--- a/packages/visual-editor/src/components/Dropzone/Dropzone.tsx
+++ b/packages/visual-editor/src/components/Dropzone/Dropzone.tsx
@@ -81,6 +81,7 @@ export function Dropzone({
           zoneId={node.data.id}
           node={node}
           resolveDesignValue={resolveDesignValue}
+          renderDropzone={renderClonedDropzone}
           {...props}
         />
       );
--- a/packages/visual-editor/src/components/Dropzone/Dropzone.tsx
+++ b/packages/visual-editor/src/components/Dropzone/Dropzone.tsx
@@ -18,6 +18,8 @@ import {
   CONTENTFUL_COMPONENTS,
 } from '@contentful/experience-builder-core/constants';
 import { RenderDropzoneFunction } from './Dropzone.types';
+import { EditorBlockClone } from './EditorBlockClone';
+import { DropzoneClone } from './DropzoneClone';
 
 type DropzoneProps = {
   zoneId: string;
@@ -71,6 +73,20 @@ export function Dropzone({
     },
     [resolveDesignValue],
   );
+  // To avoid a circular dependency, we create the recursive rendering function here and trickle it down
+  const renderClonedDropzone: RenderDropzoneFunction = useCallback(
+    (node, props) => {
+      return (
+        <DropzoneClone
+          zoneId={node.data.id}
+          node={node}
+          resolveDesignValue={resolveDesignValue}
+          {...props}
+        />
+      );
+    },
+    [resolveDesignValue],
+  );
 
   if (!resolveDesignValue) {
     return null;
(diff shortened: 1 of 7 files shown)


Chaoste:
If we're fine with ignoring the warning message, is there a way we can hide it via a monkey patch? If it shows in dev everytime, someones drags & drops something on the canvas, this could spam our logs quite a lot while we depend on those for observing our protocol messages.

packages/visual-editor/src/components/Dropzone/Dropzone.tsx
```diff
@@ -71,6 +73,21 @@ export function Dropzone({
     },
     [resolveDesignValue],
   );
+  // To avoid a circular dependency, we create the recursive rendering function here and trickle it down
```
Chaoste, line 76:
```suggestion
```
it's not a recursive rendering in this case, so we can drop the copied message in here.

Chaoste (commented):
The code changes and the recording look good to me. I didn't functionally test though and would leave that for someone else before approving ;)

Small comment: If I understood correctly, the "clone" elements are the ones that are moved around? I remember from other scenarios and libraries that this is called usually a "ghost" element. Maybe I misunderstood put that's at least something I would personally recognize faster as "clone" is kinda reserved word in many contexts.

chasepoirier:
> > I've thoroughly tested and this has no affect on the actual experience since the droppableIds removed are part of the component being dragged but it does show up for us while developing locally.
> 
> If we're fine with ignoring the warning message, how about we hide it via a [monkey patch](https://codepen.io/blackjacques/pen/XEOyme)? If it shows in dev every time, someone drags & drops something on the canvas, this could spam our logs quite a lot while we depend on those for observing our protocol messages.

That could be a good idea if I can't find a solution to resolve the warning. I thought of one more idea to try last night so going to see if that could work before merging. If not, we'll probably want to suppress the log with a monkey patch type solution

chasepoirier:
@Chaoste we chatted during stand up on this issue with the console.warn and decided we're okay leaving it in for local builds for now.

After GA we'll revisit the topic with this issue: https://contentful.atlassian.net/browse/ALT-529

PR #439, opened 2024-02-28
fix: Media type and Image component backwards compat fixes

elylucas:
This PR contains 2 fixes for backwards compatibility with how Media type use to work and the Image component

Small tweak to `transformMedia` to pass the proper path into it (allows Media types that aren't `cfImageAsset` or `cfImageBackgroundUrl` to function as they did prior).

In CompositionBlock, we were looping over the values on the node to build the props. This led to a runtime error if the component had a prop bound to it, but the prop was subsequently removed from the component definition. I switched this to loop over the component definition variables instead, which is how this logic was already done in the useComponentProps hook in visual-editor.


packages/components/src/components/Image/Image.tsx
```diff
@@ -9,11 +8,7 @@ export interface ImageProps extends React.ImgHTMLAttributes<HTMLImageElement> {
 
 export const Image: React.FC<ImageProps> = ({ className = '', src, cfImageAsset, ...props }) => {
   if (!cfImageAsset && !src) {
-    return (
-      <div className="cf-no-image-wrapper">
-        <img src={placeholderImage} className={'cf-image ' + className} {...props} />
-      </div>
-    );
+    return <img src={placeholderImage} className={'cf-image cf-no-image' + className} {...props} />;
```
primeinteger, line 11: Should there be an extra space after the `cf-no-image` class?

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -83,38 +83,41 @@ export const CompositionBlock = ({
...
-          acc[variableName] = value;
-          break;
-        }
-        case 'UnboundValue': {
-          const uuid = variable.key;
-          acc[variableName] = entityStore.unboundValues[uuid]?.value;
-          break;
+    return Object.entries(componentRegistration.definition.variables).reduce(
+      (acc, [variableName, variableDefinition]) => {
+        const variable = node.variables[variableName];
```
primeinteger, line 88: > In CompositionBlock, we were looping over the values on the node to build the props. This led to a runtime error if the component had a prop bound to it, but the prop was subsequently removed from the component definition. I switched this to loop over the component definition variables instead, which is how this logic was already done in the useComponentProps hook in visual-editor.

Makes sense to me 👍

primeinteger (approved):

packages/components/src/components/Image/Image.tsx
elylucas: good catch

PR #440, opened 2024-02-28
feat: adds tooltip defn for heading, text, richtext, button, and image [ALT-462

ryunsong-contentful:
## Purpose
Update component defn to allow us and external users to define their own tooltips by adding 
```
  tooltip?: {
    imageUrl?: string;
    description: string;
  };
```
to the component defn.

Also adds tooltip defns for the following Basic components - heading, text, richtext, button, and image

### DIFF
--- a/packages/components/src/components/Button/index.ts
+++ b/packages/components/src/components/Button/index.ts
@@ -20,6 +20,9 @@ export const ButtonComponentDefinition: ComponentDefinition = {
     'cfLineHeight',
     'cfBorder',
   ],
+  tooltip: {
+    description: 'A button that can be used to trigger an action or navigate to a different page.',
+  },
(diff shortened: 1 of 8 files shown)


packages/components/src/components/Button/index.ts
```diff
@@ -20,6 +20,9 @@ export const ButtonComponentDefinition: ComponentDefinition = {
     'cfLineHeight',
     'cfBorder',
   ],
+  tooltip: {
+    description: 'A button that can be used to trigger an action or navigate to a different page',
+  },
```
lilbitner, line 25: super nit: seems like we add a period to all of the other descriptions, should we add here?

lilbitner (approved):

PR #441, opened 2024-02-29
chore: create .npmrc file for publish step

elylucas:
This creates a .npmrc file right before the publish step so the job can auth with GHA

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -48,14 +48,14 @@ jobs:
             packages/*/build
           key: build-cache-${{ github.run_id }}-${{ github.run_attempt }}
       - name: Authenticate with private GH package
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yaml
```diff
@@ -47,6 +47,10 @@ jobs:
             packages/*/dist
             packages/*/build
           key: build-cache-${{ github.run_id }}-${{ github.run_attempt }}
+      - name: Authenticate with private GH package
+        run: echo -e "//npm.pkg.github.com/:_authToken=${{ steps.vault.outputs.GITHUB_PACKAGES_WRITE_TOKEN }}\n@contentful:registry=https://npm.pkg.github.com" > ./.npmrc
```
ryunsong-contentful, line 51: Would it be better to be more explicit about the file path and use `~/.npmrc` over a relative path

ryunsong-contentful (approved):
LGTM, just a small comment

.github/workflows/publish.yaml
elylucas: This puts it in the current directory. Originally it was the home directory like you mentioned, but I think I changed to to the local directory trying to get it to work. Either place should work.

PR #443, opened 2024-02-29
fix(canvas-interactions) dynamic hitboxes while dragging ALT-504

chasepoirier:
## Purpose
Canvas interactions were occasionally inaccurate or ended up in a "flickering" state where dropzones where stuck flickering back and forth. The PR enables the usage of "hitboxes" that are fixed on top of the canvas using `position: fixed`. This ensures that as components move around the canvas while dragging, that the intended dropzone target is not misinterpreted. 

Tickets:
- https://contentful.atlassian.net/browse/ALT-504
- https://contentful.atlassian.net/browse/ALT-455 (small cleanup on columns empty state styling)

This PR also includes work from @primeinteger on https://contentful.atlassian.net/browse/ALT-308

### DIFF
--- a/packages/core/src/utils/components.ts
+++ b/packages/core/src/utils/components.ts
@@ -1,17 +1,19 @@
 import { CONTENTFUL_COMPONENTS } from '../constants';
 
+const structureComponents = new Set([
+  CONTENTFUL_COMPONENTS.section.id,
+  CONTENTFUL_COMPONENTS.columns.id,
+  CONTENTFUL_COMPONENTS.container.id,
+  CONTENTFUL_COMPONENTS.singleColumn.id,
+]);
+
 export const isContentfulStructureComponent = (componentId?: string) =>
-  [
-    CONTENTFUL_COMPONENTS.section.id,
-    CONTENTFUL_COMPONENTS.columns.id,
-    CONTENTFUL_COMPONENTS.container.id,
-    CONTENTFUL_COMPONENTS.singleColumn.id,
-  ].includes(componentId ?? '');
+  structureComponents.has(componentId ?? '');
 
 export const isEmptyStructureWithRelativeHeight = (
   children: number,
   componentId?: string,
-  height?: string | number
+  height?: string | number,
 ) => {
   return (
     children === 0 &&
--- a/packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
+++ b/packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
@@ -1,0 +1,206 @@
+import React, { CSSProperties, useCallback, useEffect, useMemo, useState } from 'react';
+import styles from './styles.module.css';
+import { useTreeStore } from '@/store/tree';
+import { getItemDepthFromTree } from '@/utils/getItem';
+import { CTFL_ZONE_ID, DRAGGABLE_WIDTH, ROOT_ID } from '@/types/constants';
+import { useZoneStore } from '@/store/zone';
+import { useDraggedItemStore } from '@/store/draggedItem';
+import { createPortal } from 'react-dom';
+
+interface Props {
+  parentZoneId: string;
+  zoneId: string;
+  enableRootHitboxes: boolean;
+}
+
+const INITIAL_OFFSET = 10;
+const OFFSET_INCREMENT = 8;
+const HITBOX_HEIGHT = 20;
+const HITBOX_WIDTH = 80;
+const MIN_HEIGHT = 45;
+const MIN_DEPTH = 20;
+enum HitboxDirection {
+  TOP,
+  LEFT,
+  RIGHT,
+  BOTTOM,
+  SELF_VERTICAL,
+  SELF_HORIZONTAL,
+}
+
+const calcOffsetDepth = (depth: number) => {
+  return INITIAL_OFFSET - OFFSET_INCREMENT * depth;
+};
+
+const Hitboxes: React.FC<Props> = ({ zoneId, parentZoneId, enableRootHitboxes }) => {
+  const tree = useTreeStore((state) => state.tree);
+  const isDraggingOnCanvas = useDraggedItemStore((state) => state.isDraggingOnCanvas);
+  const zoneDepth = useMemo(
+    () => getItemDepthFromTree({ id: parentZoneId }, tree),
+    [tree, parentZoneId],
+  );
+  const [fetchDomRect, setFetchDomRect] = useState(Date.now());
+
+  useEffect(() => {
+    /**
+     * A bit hacky but we need to wait a very small amount
+     * of time to fetch the dom getBoundingClientRect once a
+     * drag starts because we need pre-drag styles like padding
+     * applied before we calculate positions of hitboxes
+     */
+    setTimeout(() => {
+      setFetchDomRect(Date.now());
+    }, 50);
+  }, [isDraggingOnCanvas]);
+
+  const hitboxContainer = useMemo(() => {
+    return document.querySelector('[data-ctfl-hitboxes]');
+  }, []);
+
+  const domRect = useMemo(() => {
+    return document.querySelector(`[${CTFL_ZONE_ID}="${zoneId}"]`)?.getBoundingClientRect();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [zoneId, fetchDomRect]);
+
+  const zones = useZoneStore((state) => state.zones);
+
+  const direction = zones[parentZoneId]?.direction || 'vertical';
+  const isVertical = direction === 'vertical';
+
+  const isRoot = parentZoneId === ROOT_ID;
+
+  const getStyles = useCallback(
+    (direction: HitboxDirection): CSSProperties => {
+      if (!domRect) {
+        return {
+          display: 'none',
+        };
+      }
+
+      const { width, height, top, left, bottom, right } = domRect;
+
+      const isDeepZone = zoneDepth > 5;
+      switch (direction) {
+        case HitboxDirection.TOP:
+          return {
+            width,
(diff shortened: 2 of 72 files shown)


packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+          className={styles.hitbox}
+          style={getStyles(HitboxDirection.SELF_VERTICAL)}
+        />
+      ) : (
+        <div
+          data-ctfl-zone-id={zoneId}
+          className={styles.hitbox}
+          style={getStyles(HitboxDirection.SELF_HORIZONTAL)}
+        />
+      )}
```
primeinteger, line 163: Some repetition could be avoided with a conditional value in the style prop
```suggestion
      <div
        data-ctfl-zone-id={zoneId}
        className={styles.hitbox}
        style={getStyles(isVertical ? HitboxDirection.SELF_VERTICAL : HitboxDirection.SELF_HORIZONTAL)}
      />
```

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+            className={styles.hitbox}
+            style={getStyles(HitboxDirection.LEFT)}
+          />
+          <div
+            data-ctfl-zone-id={parentZoneId}
+            className={styles.hitbox}
+            style={getStyles(HitboxDirection.RIGHT)}
+          />
+        </>
+      )}
```
primeinteger, line 196: Some repetition could be eliminated here as well since the only difference is the style prop value

```ts
      ) : (
        <>
          <div
            data-ctfl-zone-id={parentZoneId}
            className={styles.hitbox}
            style={getStyles(isVertical ? HitboxDirection.TOP : HitboxDirection.LEFT)}
          />
          <div
            data-ctfl-zone-id={parentZoneId}
            className={styles.hitbox}
            style={getStyles(isVertical ? HitboxDirection.BOTTOM : HitboxDirection.RIGHT)}
          />
        </>
      )}
```

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+            width: HITBOX_WIDTH,
+            height: height - HITBOX_HEIGHT,
+            left: right - calcOffsetDepth(zoneDepth) - HITBOX_WIDTH / 2,
+            top: top + HITBOX_HEIGHT / 2,
+            zIndex: 100 + zoneDepth,
+          };
+        case HitboxDirection.SELF_VERTICAL: {
+          if (height > 60 && !isDeepZone) {
+            return { display: 'none' };
+          }
```
primeinteger, line 119: Why `60`?

I see the `SELF_HORIZONTAL` case uses `width > DRAGGABLE_WIDTH`. Would it make sense to have this condition use `height > DRAGGABLE_HEIGHT * 2` instead?

primeinteger (approved):
Looks awesome! Great enhancement to the visual editor drop zones and hitbox technology 👏

packages/core/src/utils/components.ts
```diff
@@ -1,17 +1,19 @@
 import { CONTENTFUL_COMPONENTS } from '../constants';
 
+const structureComponents = new Set([
+  CONTENTFUL_COMPONENTS.section.id,
+  CONTENTFUL_COMPONENTS.columns.id,
+  CONTENTFUL_COMPONENTS.container.id,
+  CONTENTFUL_COMPONENTS.singleColumn.id,
+]);
+
 export const isContentfulStructureComponent = (componentId?: string) =>
```
ryunsong-contentful, line 10: What is the point of making this input param of the function optional? I feel like we should just require a componentId in this function.

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+const Hitboxes: React.FC<Props> = ({ zoneId, parentZoneId, enableRootHitboxes }) => {
+  const tree = useTreeStore((state) => state.tree);
+  const isDraggingOnCanvas = useDraggedItemStore((state) => state.isDraggingOnCanvas);
+  const zoneDepth = useMemo(
+    () => getItemDepthFromTree({ id: parentZoneId }, tree),
+    [tree, parentZoneId],
+  );
+  const [fetchDomRect, setFetchDomRect] = useState(Date.now());
+
+  useEffect(() => {
```
ryunsong-contentful, line 44: You could try useLayoutEffect - I forget the details but I think this is a case where it's worth looking into. Basically the useLayoutEffect hook is used to paint the css and make style related dom edits.

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+  }, [zoneId, fetchDomRect]);
+
+  const zones = useZoneStore((state) => state.zones);
+
+  const direction = zones[parentZoneId]?.direction || 'vertical';
+  const isVertical = direction === 'vertical';
+
+  const isRoot = parentZoneId === ROOT_ID;
+
+  const getStyles = useCallback(
```
ryunsong-contentful, line 72: Maybe we can move this code to a different file - either a styles file or some utils.ts file?

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
```diff
@@ -0,0 +1,206 @@
...
+    (direction: HitboxDirection): CSSProperties => {
+      if (!domRect) {
+        return {
+          display: 'none',
+        };
+      }
+
+      const { width, height, top, left, bottom, right } = domRect;
+
+      const isDeepZone = zoneDepth > 5;
```
ryunsong-contentful, line 82: nit: Assign 5 to some variable called DEEP_ZONE?

packages/visual-editor/src/components/Dropzone/useComponentProps.ts
```diff
@@ -204,3 +218,15 @@ export const useComponentProps = ({
 
   return { componentProps, wrapperProps };
 };
+
+const addExtraDropzonePadding = (padding: string) =>
+  padding
+    .split(' ')
+    .map((value) => {
+      if (value.endsWith('px')) {
```
ryunsong-contentful, line 226: Could do some future proofing using these potential values - unless we know it'll always be px `export const SpacingUnitRegexp = /px|cm|mm|in|pt|pc|em|ex|ch|rem|vw|vh|vmin|vmax|%$/;`

packages/visual-editor/src/utils/getItem.ts
```diff
@@ -31,6 +31,26 @@ function getItemFromTree(
   return undefined;
 }
 
+function findDepthById(
+  node: CompositionComponentNode,
+  id: string,
+  currentDepth: number = 1,
+): number {
+  if (node.data.id === id) {
```
ryunsong-contentful, line 39: finally, an interview DFS algo problem in react

packages/visual-editor/src/utils/getItem.ts
```diff
@@ -68,3 +88,17 @@ export const getItem = (
     children: tree.root.children,
   });
 };
+
+export const getItemDepthFromTree = (selector: ItemSelector, tree: CompositionTree): number => {
```
ryunsong-contentful, line 92: is it possible to even generalize this further, calling it `getItemByNode` and then instead of ROOT_ID you pass in `node.id` and children would be `node.children`?

packages/core/src/utils/components.ts
primeinteger: Basically there are several places in the UI and SDK where is function is called but the componentId might be undefined, so we simplify the implementation by allowing the param to be undefined as well.

Screenshot 2024-02-29 at 2 24 55 PM

packages/visual-editor/src/components/Dropzone/useComponentProps.ts
primeinteger: This method came from my PR originally. I agree it would be nice to support other units of measurement here, but it also seems uncommon for padding to be implemented using some of these options.

This method is used for detecting when there is `< 4px` padding on components so that we can attempt to expand the padding on sides of components that have less padding and avoid decreasing the padding on sides that have more than 4px.  

To support other units, we'd need to convert the value from other units to pixels to check if the converted pixel value `< 4px` or something that would help determine when `4px` is less than the value 🤔

primeinteger: TLDR: We could add a `convertToPx` method but probably as a follow-up feature.

ryunsong-contentful:
Store this PR in your brag doc, this ones for the ages. For the complexity of problems it solves, it's very clean

ryunsong-contentful (approved):

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
chasepoirier: You're officially the king of ternary optimizations 😎 👑

packages/visual-editor/src/components/Dropzone/useComponentProps.ts
chasepoirier: Ticket for this: https://contentful.atlassian.net/browse/ALT-543

packages/visual-editor/src/components/Dropzone/Hitboxes.tsx
chasepoirier: Ticket for this: https://contentful.atlassian.net/browse/ALT-542

PR #446, opened 2024-03-01
fix: use merge instead of rebase when syncing branches [ci skip]

SofiaMargariti:
## Purpose

Avoid rebase conflicts when syncing branches in CI.

## Approach

Sometimes rebase fails when trying to sync `development` branch with `next`. E.g. in [this build](https://github.com/contentful/experience-builder/actions/runs/8110520652/job/22168113172#step:8:286). Last time that this happened was because:
* merging from a feature branch into `development` commits were not squashed (PR of ~70 commits)
* merging from `development` into `next` the 70 commits were squashed into one
* when later CI tried to rebase `development` over `next` there were conflicts between the individual commits in development and the squashed commit in next.

Using `merge` instead of `rebase` does not produce any conflicts. Only caveat is that it will create a new merge commit on next but that shouldn't cause any problems on followup PRs.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -70,13 +70,13 @@ jobs:
 
           if [ ${{ github.ref_name }} = main ]; then
             git checkout next
-            git rebase main
+            git merge main
             git push origin
             git checkout development
-            git rebase next
+            git merge next
             git push origin
           elif [ ${{ github.ref_name }} = next ]; then
             git checkout development
-            git rebase next
+            git merge next
             git push origin
           fi


elylucas:
This is a situation that shouldn't happen if both next and development are clean and in-sync with each other. I'd think that a MC would happen on a git merge in this case as well.

We can't squash merge into next, all the commit history would be lost and the semantic release wouldn't know how to do the release properly. We should be squash merging from our feature branches into development, and the merge commit into next. If that happens, then next will be inline with development and there shouldn't be any issues when it comes time to rebase.

It would be a real unfortunate side effect to have merge commits in the history from releasing due to lack of following process when merging. I realize accidents will happen, but I'd rather see those cleaned up manually.

SofiaMargariti:
> This is a situation that shouldn't happen if both next and development are clean and in-sync with each other. I'd think that a MC would happen on a git merge in this case as well.

In this case `git merge` wouldn't generate a conflict based on my test. I think that is because during rebase you replay each commit separately so if there is a conflict in commit A that is later fixed in commit B `git rebase` will still fail. `git merge` doesn't have this problem because it compares the heads of each branch I guess.

> We can't squash merge into next, all the commit history would be lost and the semantic release wouldn't know how to do the release properly. We should be squash merging from our feature branches into development, and the merge commit into next. If that happens, then next will be inline with development and there shouldn't be any issues when it comes time to rebase.

I don't think the whole history got lost, just the commit of the last merged PR, but I agree we shouldn't squash merge into `next`. Unfortunately I don't see a way to enforce that. What we can enforce is linear history for `development` (so either squash or rebase) via [this setting](https://github.com/contentful/experience-builder/settings/branch_protection_rules/47333757) 
Screenshot 2024-03-04 at 16 47 00
Let me know if you see any issues with having that enabled.

> It would be a real unfortunate side effect to have merge commits in the history from releasing due to lack of following process when merging. I realize accidents will happen, but I'd rather see those cleaned up manually.

Ok, I'll close this PR then and we can manually push if we ran into the same problem again. Should we disallow force-pushing to development or next in this case?

PR #459, opened 2024-03-04
fix(canvas-interactions) resize draggables before dragging ALT-537

chasepoirier:


chasepoirier:
Closing as this will be worked on in the future, post GA https://contentful.atlassian.net/browse/ALT-551

PR #462, opened 2024-03-05
chore: Merge media deep binding with media asset changes[]

elylucas:


packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -27,23 +30,150 @@ export class EntityStoreBase {
     this.addEntity(entity);
   }
 
+  public getEntryOrAsset(
```
elylucas, line 33: The bulk of my changes are here in this method. I refactored the code from `CompositionBlock` and `useComponentProps` to fetch their bound assets from a central place in the `transformBoundContent` method in core. Therefore, the deep entity fetching logic had to be moved someplace central as well.

I put it here in this method. Hopefully it covers all the rest of the cases as well.

packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -27,23 +30,150 @@ export class EntityStoreBase {
...
+        if (undefined === fieldValue) {
+          return {
+            resolvedFieldset,
+            isFullyResolved: false,
+            reason: `Cannot resolve field Link<${entityToResolveFieldsFrom.sys.type}>(sys.id=${entityToResolveFieldsFrom.sys.id}).fields[${field}] as field value is not defined`,
+          };
+        } else if (isLink(fieldValue)) {
+          const entity = this.getEntityFromLink(fieldValue);
+          if (entity === undefined) {
+            return {
```
elylucas, line 72: I switched this from throwing an error to returning a reason, as in editor mode, not all the entites were loaded yet when this ran. Contrary to what the previous error said, I found that if we just wait then they will all be loaded eventually and this will return fully resolved.

packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -27,6 +30,120 @@ export class EntityStoreBase {
...
+      if (!isFullyResolved) {
+        reason &&
+          console.debug(
+            `[exp-builder.sdk::EntityStoreBased::getValueDeep()] Deep path wasn't resolved till leaf node, falling back to undefined, because: ${reason}`,
+          );
+        return undefined;
+      }
+      const [leafEntity] = resolvedFieldset[resolvedFieldset.length - 1];
+      return leafEntity;
+    } else {
```
primeinteger, line 119: I think removing this `else` statement could simplify the flow and slightly reduce indentation. Since the `if` condition above early returns, the `else` clause can be avoided.

primeinteger (approved):
Looking good to me!  Just commenting on a couple small coding style things

packages/core/src/entity/EntityStoreBase.ts
elylucas: The one scenario where I've found where it comes up:

When you are referencing a deeply linked entry's asset and you update that asset in another tab, the live reload in the editor will run into this situation. With my code changes, it would error here, but now it will just throw a warning and eventually the linked asset will be loaded.

packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -27,6 +30,120 @@ export class EntityStoreBase {
...
+      const { resolvedFieldset, isFullyResolved, reason } = resolveFieldset(
+        unresolvedFieldset,
+        headEntity,
+      );
+      if (!isFullyResolved) {
+        reason &&
+          console.debug(
+            `[exp-builder.sdk::EntityStoreBased::getValueDeep()] Deep path wasn't resolved till leaf node, falling back to undefined, because: ${reason}`,
+          );
+        return undefined;
```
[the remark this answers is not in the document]
elylucas: I took care of this in the PR feedback commit.

packages/core/src/entity/EntityStoreBase.ts
elylucas: Normally I'd agree, but there was quite a bit of logic in both statements and I think it becomes confusing in that case. So to combat that, I refactored out the code in the top if block into a new `getDeepEntry` method on the class and was able to do this.

PR #464, opened 2024-03-05
fix: update component tooltips to use component term instead of elements [ALT-544]

ryunsong-contentful:
## Purpose
Replace the term Elements with Components
Clean up the consistency between tooltip descriptions


primeinteger (approved):
😎

PR #465, opened 2024-03-05
fix(canvas-interactions): disable dropzones that are assemblies [ALT-552]

chasepoirier:
## Purpose
Assemblies (Patterns) should disable any Dropzones that are nested within itself since they are read-only.

Ticket: https://contentful.atlassian.net/browse/ALT-552

## Approach
Disable Dropzones that are of the assembly node type. Also add a better comment to outline the "Rules of Dropzones"


ryunsong-contentful (approved):
love the comments

PR #467, opened 2024-03-06
feat: added toJSON function in entity store [SPA-1884]

anwaar931:
## Purpose

Serialize entry and asset references with JSON.stringify()

## Approach

Introduced toJSON function in EntityStore class


Spring3 (approved):
Lovely. Thanks a lot!

PR #468, opened 2024-03-06
fix(canvas-interactions) prevent centering of draggables while reparenting [ALT-557]

chasepoirier:
## Purpose
When reparenting a component on the canvas, starting a drag should not cause the component to move to be centered on the mouse. Instead it should take it's pre-drag coordinates and use those to properly position itself at the start of a drag

Also add an outline to the dragging component to better represent the full size of the draggable when there may be invisible space to the user.

### Additional Bug fixes from a side quest
- Root hitboxes were being rendered even when reparenting non-root components, they should not be rendered in that scenario
- Hitboxes weren't scrolling with the page
- Moving a component off canvas was causing the draggable to jump to the wrong coordinates when scrolled on the page

https://github.com/contentful/experience-builder/assets/23040747/e19294fd-95d0-424f-bfaf-6b04c446b95f


packages/visual-editor/src/hooks/useDraggablePosition.ts
```diff
@@ -0,0 +1,38 @@
+import { useDraggedItemStore } from '@/store/draggedItem';
+import { DraggablePosition } from '@/types/constants';
+import { MutableRefObject, useEffect } from 'react';
+
+interface Params {
+  draggableRef: MutableRefObject<HTMLElement | null | undefined>;
```
primeinteger, line 6: I think this type can be changed to `MutableRefObject<HTMLElement | null>` as there doesn't seem to be a case where the value would be `undefined`.

The `useRef` in `<DraggableComponent>` and `<DraggableComponentList>` would need to be changed to also use `<HTMLElement | null>` to match

packages/visual-editor/src/hooks/useDraggablePosition.ts
```diff
@@ -0,0 +1,38 @@
...
+  position: DraggablePosition;
+}
+
+export default function useDraggablePosition({ draggableId, draggableRef, position }: Params) {
+  const isDraggingOnCanvas = useDraggedItemStore((state) => state.isDraggingOnCanvas);
+  const draggingId = useDraggedItemStore((state) => state.onBeforeCaptureId);
+  const preDragDomRect = useDraggedItemStore((state) => state.domRect);
+
+  useEffect(() => {
+    const el: HTMLElement | undefined | null = draggableRef?.current;
```
primeinteger, line 17: Here's another place I think that `undefined` could safely be removed

primeinteger (approved):
Bringing the heat with this one! Great work🔥

packages/visual-editor/src/hooks/useDraggablePosition.ts
chasepoirier: Good catch!

PR #469, opened 2024-03-06
refactor: Remove deprecated apis in Core constants/types, and Sdk hooks

ryunsong-contentful:
## Purpose
Before we launch to GA and release a Beta version of Experiences sdk, we will want to delete deprecated apis


packages/components/README.md
```diff
@@ -47,7 +47,7 @@ Import the `useExperienceBuilderComponents` hook from the `@contentful/experienc
 import { useExperienceBuilderComponents } from '@contentful/experiences-components-react';
 ```
 
-After the call to `useExperienceBuilder` (where you obtain the `defineComponents` method), pass in `defineComponents` to the `userExperienceBuilder` hook:
+After the call to `fetchBySlug` or '`fetchById` (where you obtain the `defineComponents` method), pass in `defineComponents` to the `userExperienceBuilder` hook:
```
primeinteger, line 50: 👀 extra `'` character before \`fetchById\`

packages/experience-builder-sdk/src/ExperienceRoot.tsx
```diff
@@ -35,8 +25,7 @@ export const ExperienceRoot = ({
   });
 
   if (isEditorMode) {
-    const entityStore =
-      experience && !isDeprecatedExperience(experience) ? experience.entityStore : undefined;
+    const entityStore = experience ? experience.entityStore : undefined;
```
primeinteger, line 28: Maybe this line can be simplified even more now
```suggestion
    const entityStore = experience?.entityStore;
```

packages/create-experience-builder/templates/react-vite-ts/src/App.tsx
```diff
@@ -28,14 +30,23 @@ function App() {
...
+        mode,
+      });
+
+      setExperience(experience);
+    };
 
-  // Register optional default components
-  useExperienceBuilderComponents(defineComponents);
+    fetchData();
+  }, [client, mode]);
```
primeinteger, line 49: Should this `create-experience-builder` app use the `useFetchBySlug` hook instead of calling `fetchBySlug` directly? 🤔

packages/components/README.md
elylucas: These instructions should go away altogether as the components are registered out of the box now

packages/create-experience-builder/templates/react-vite-ts/src/App.tsx
elylucas: fo sho, see the wiki set up instructions on how: https://github.com/contentful/experience-builder/wiki#fetch-and-display-the-experience

elylucas: Actually, i have a branch where this is already updated, heres the code you should be able to use as it for `app.tsx`:

```
import { createClient } from 'contentful';
import { useFetchBySlug, ExperienceRoot } from '@contentful/experience-builder';
import './App.css';

// Import the styles for the default components
import '@contentful/experience-builder-components/styles.css';

const experienceTypeId = import.meta.env.VITE_CTFL_EXPERIENCE_TYPE_ID; //Content type id for the experience
const localeCode = 'en-US'; //Locale code for the experience (could be dynamic)

const client = createClient({
  // your space id
  space: import.meta.env.VITE_CTFL_SPACE_ID,
  // your environment id
  environment: import.meta.env.VITE_CTFL_ENV_ID,
  // Supported values: 'preview.contentful.com' or 'cdn.contentful.com',
  host: import.meta.env.VITE_CTFL_API_HOST,
  // needs to be access token if host = 'cdn.contentful.com' and preview token if 'preview.contentful.com'
  accessToken: import.meta.env.VITE_CTFL_ACCESS_TOKEN,
});

function App() {
  const { experience, isLoading, error } = useFetchBySlug({
    client,
    slug: 'homePage', //Could be fetched from the url,
    experienceTypeId,
    localeCode,
  });

  // handle loading and error states
  if (isLoading) return <div>Loading...</div>;
  if (error) return <div>Error: {error.message}</div>;

  return <ExperienceRoot experience={experience} locale={localeCode} />;
}

export default App;

```

packages/create-experience-builder/templates/react-vite-ts/src/App.tsx
```diff
@@ -1,44 +1,38 @@
+import React from 'react';
 import { createClient } from 'contentful';
-import { useExperienceBuilder, ExperienceRoot } from '@contentful/experiences-sdk-react';
-import { useExperienceBuilderComponents } from '@contentful/experiences-components-react';
+import { useFetchBySlug, ExperienceRoot } from '@contentful/experience-builder';
```
primeinteger, line 3: This import and `styles.css` import below should use the new package names

primeinteger (approved):
I think this is good to go after those last two imports are updated! Great work 👏

PR #470, opened 2024-03-06
fix: changes for Image Object Fit options [ALT-511]

primeinteger:
## Purpose

I decided to remove the `SUPPORTED_IMAGE_OBJECT_FIT` constant and use an object for the display names and values of each image object fit option in the UI:
* https://github.com/contentful/user_interface/blob/185408186abbfca07ac4dcb1326ed9faab035fb9/src/javascripts/features/experience-builder/components/StyleSectionComponents/ImageOptions/ImageFitSection.tsx#L14-L18

This makes the implementation a little more consistent with other style props such as Image Alignment options which follow this implementation:
* https://github.com/contentful/user_interface/blob/185408186abbfca07ac4dcb1326ed9faab035fb9/src/javascripts/features/experience-builder/components/StyleSectionComponents/ImageOptions/ImageAlignmentInput.tsx#L26-L36

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -176,10 +176,3 @@ export enum PostMessageMethods {
 }
 
 export const SUPPORTED_IMAGE_FORMATS = ['jpg', 'png', 'webp', 'gif', 'avif'] as const;
-export const SUPPORTED_IMAGE_OBJECT_FIT = [
-  'fill',
-  'contain',
-  'cover',
-  'none',
-  'scale-down',
-] as const;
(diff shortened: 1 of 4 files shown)


elylucas (approved):

packages/core/src/constants.ts
```diff
@@ -176,10 +176,3 @@ export enum PostMessageMethods {
 }
 
 export const SUPPORTED_IMAGE_FORMATS = ['jpg', 'png', 'webp', 'gif', 'avif'] as const;
-export const SUPPORTED_IMAGE_OBJECT_FIT = [
```
elylucas, line 179: this is probably okay overall. The one gotcha I could see is if we wanted to support another option that was removed, that would require and SDK update. Did you work with travis on removing some of these?

PR #475, opened 2024-03-08
fix: update Image component placeholder styling [ALT-528]

primeinteger:
## Purpose

Changing the Image component placeholder (worked with Travis on these design changes)

Screenshot 2024-03-08 at 11 02 28 AM

Video showing how the Image component placeholder adjusts to different sizes
* The icon in the middle is an SVG so that it won't stretch as the size of the placeholder is changed

https://github.com/contentful/experience-builder/assets/8539634/3745e2fe-d720-4e85-b7d7-974ba5bac7fc

## Approach

The placeholder image uses `src="data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAA"` which is a 1px transparent image:

Screenshot 2024-03-08 at 11 12 34 AM

I took this approach to use an actual `` tag (which requires a `src`) so the placeholder image will behave exactly like a real image does when changing the size/dimensions on the placeholder (before an image is bound, or when editing a pattern).  

When I tried making the placeholder image a `<div>` instead, changing the Image size controls behaves differently for the fill/auto relative sizes.

### DIFF
--- a/packages/components/src/components/variables.css
+++ b/packages/components/src/components/variables.css
@@ -8,7 +8,6 @@
   --cf-color-black: #000;
   --cf-color-gray100: #f7f9fa;
   --cf-color-gray400: #aec1cc;
-  --cf-color-gray100-rgb: 247, 249, 250;
   --cf-color-gray400-rgb: 174, 193, 204;
 
   /* spacing */
--- a/packages/components/src/components/variables.css
+++ b/packages/components/src/components/variables.css
@@ -7,15 +7,9 @@
   --cf-color-white: #fff;
   --cf-color-black: #000;
   --cf-color-gray100: #f7f9fa;
-  --cf-color-gray200: #e7ebee;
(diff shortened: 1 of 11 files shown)


ryunsong-contentful (approved):

packages/components/src/components/variables.css
```diff
@@ -6,6 +6,10 @@
   /* color */
   --cf-color-white: #fff;
   --cf-color-black: #000;
+  --cf-color-gray100: #f7f9fa;
+  --cf-color-gray400: #aec1cc;
+  --cf-color-gray100-rgb: 247, 249, 250;
```
ryunsong-contentful, line 11: Why is the rgb vs hex version needed?

primeinteger: Removed the extra color var 👍

elylucas (approved):
looks great!

PR #476, opened 2024-03-08
fix: hover outlines and default border style [ALT-323]

primeinteger:
## Fix hover outlines on sections with backgrounds

Ticket: https://contentful.atlassian.net/browse/ALT-323

Changes visual editor hover outlines to apply on `:before` psuedo-elements so that the outlines will always render on top.

* This makes it so the hover outlines will show on structure components that have a background image or background color.

This video compares hover outlines in Test App Prod  vs locally to show that the hover outlines works on the yellow background image section and green background color section (in the second half of the video):

https://github.com/contentful/experience-builder/assets/8539634/3ad38711-fa7e-421e-aa68-cbf829665b97

## Fix 1px transparent border causing gaps between sections

Changes the default border style to 0px so that we won't have a 1px transparent border on structure components by default

* This was causing the small gap between buttons that are stacked in the canvas within separate sections/containers.
* The UI already handles setting the border initial value to 1px when a border styles are enabled on the component: https://github.com/contentful/user_interface/blob/master/src/javascripts/features/experience-builder/components/SectionStyleInput.tsx#L62

This video compares the border behavior in Test App Prod vs locally to show that there is no longer a small gap between buttons in a stack of buttons in separate sections, and when adding nested containers with background colors the child section fills the parent section completely without the parent background bleeding through due to the 1px transparent borders (in the second half of the video):

https://github.com/contentful/experience-builder/assets/8539634/3ea651f4-cf0d-4f7e-908e-0a00bbe7a353

## Fix background color when dragging into drop zones with a background image

Changes drop zone styling to apply the background color as an overlay, to work better with sections that are styled with a background color or background image:

https://github.com/contentful/experience-builder/assets/8539634/c7e966fd-0941-457f-bb8e-02503a6d9212

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -110,7 +110,7 @@ export const builtInStyles: Partial<
     type: 'Text',
     group: 'style',
     description: 'The border of the section',
-    defaultValue: '1px solid rgba(0, 0, 0, 0)',
+    defaultValue: '0px solid rgba(0, 0, 0, 0)',
   },
   cfGap: {
     displayName: 'Gap',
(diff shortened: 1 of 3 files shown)


elylucas (approved):

chasepoirier:
Awesome fix here @primeinteger 🔥

PR #477, opened 2024-03-11
refactor: replace Composition -> Experience in core SDK interfaces

SofiaMargariti:
## Purpose
Stop using outdated terms in the SDK

## Approach
-> Replace use of term `Composition` with `Experience` in
* types.ts:

  * `CompositionComponentNode` -> `ExperienceTreeNode`
  * `CompositionTree` -> `ExperienceTree`
  * `Composition` -> `ExperienceFields`
  * `CompositionEntry` -> `ExperienceEntry`
* constants.ts
  * IncomingEvents.CompositionUpdated -> `IncomingEvents.ExperienceUpdated`

-> ~Mark old types as deprecated~ I've removed the old types completely since most of them were already deprecated and we don't want to carry newly deprecated ones past GA.
-> Replace use of deprecated types across the monorepo. Includes [there types](https://github.com/contentful/experience-builder/blob/77507499031655374717eaba3455a4131eac11b2/packages/core/src/types.ts#L19C1-L33C50) that were previously deprecated.

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -28,7 +28,7 @@ export const OUTGOING_EVENTS = {
 
 export const INCOMING_EVENTS = {
   RequestEditorMode: 'requestEditorMode',
-  CompositionUpdated: 'componentTreeUpdated',
+  ExperienceUpdated: 'componentTreeUpdated',
   ComponentDraggingChanged: 'componentDraggingChanged',
   ComponentDragCanceled: 'componentDragCanceled',
   ComponentDragStarted: 'componentDragStarted',
--- a/packages/components/src/components/Assembly/Assembly.tsx
+++ b/packages/components/src/components/Assembly/Assembly.tsx
@@ -1,5 +1,5 @@
 import {
-  CompositionComponentNode,
+  ExperienceTreeNode,
   ResolveDesignValueType,
   StyleProps,
 } from '@contentful/experiences-core/types';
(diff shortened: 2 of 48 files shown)


packages/core/src/constants.ts
```diff
@@ -28,7 +28,7 @@ export const OUTGOING_EVENTS = {
 
 export const INCOMING_EVENTS = {
   RequestEditorMode: 'requestEditorMode',
-  CompositionUpdated: 'componentTreeUpdated',
```
SofiaMargariti, line 31: [Comment for reviewer] Since the actual value hasn't changed here and this is only used by User Interface we don't need to mark the old name as deprecated here. Instead I am removing it straight away.

Spring3 (approved):
I guess we should also have another round over `assembly` -> `pattern` renaming

dimitrycf:
Maybe this console log is worth renaming? 

image

dimitrycf (approved):
Most of those types are actually not public (as in not exposed by SDK React).

(I could only find one public export `INCOMING_EVENTS` that is  exported by the `experiences-sdk-react`. So seems all the rest of the exports are exports of `core` package which is used by the user_interface and within the monorepo). 

To test this change I have also tried building experience-builder-demo with this branch and it built without issues. (there was some sourcemap warning, but it doesn't seem to be related to renaming of types). 

LGTM

SofiaMargariti:
> I know this ticket is primarily about Types, but I wonder if it's worth also renaming word `composition` in tests?
> 
Yes, I can open another PR for this, but I want to get the public facing changes out of the way first, so I didn't spent too much time on the tests.

PR #479, opened 2024-03-11
fix(components): change default of structure component margins to 0 [ALT-519]

chasepoirier:
## Purpose
Small follow up PR related to https://github.com/contentful/user_interface/pull/20219 and fixes https://contentful.atlassian.net/jira/software/c/projects/ALT/boards/1552


ryunsong-contentful (approved):

ryunsong-contentful:
I tried thinking through the implications of the changing the default and making the component sit at the root level the exception (where you apply margin auto) and I don't think there's any down the road blockers that this affects

PR #482, opened 2024-03-13
fix: consolidate image option variables [ALT-581]

primeinteger:
## Purpose

Ticket: https://contentful.atlassian.net/browse/ALT-581

This changes the image component and background images to condense multiple design style variables down into an object variable with multiple style values.

Now the Image component will leverage two variables (one for binding the image asset named `cfImageAsset` and another for design style options named `cfImageOptions`)

```ts
  cfImageAsset: {
    displayName: 'Image',
    type: 'Media',
    description: 'Image to display',
  },
  cfImageOptions: {
    displayName: 'Image options',
    type: 'ImageOptions',
    group: 'style',
    defaultValue: {
      width: DEFAULT_IMAGE_WIDTH,
      height: '100%',
      objectFit: 'none',
      objectPosition: 'center center',
      quality: '100',
      targetSize: DEFAULT_IMAGE_WIDTH,
    },
  },
```

Now the Background Image on components will leverage two variables (one for binding the image asset named `cfBackgroundImageUrl` and another for design style options named `cfBackgroundImageOptions`)

```ts
  cfBackgroundImageUrl: {
    displayName: 'Background image',
    type: 'Media',
    description: 'Background image for component',
  },
  cfBackgroundImageOptions: {
    displayName: 'Background image options',
    type: 'BackgroundImageOptions',
    group: 'style',
    defaultValue: {
      scaling: 'fill',
      alignment: 'left top',
      quality: '100',
      targetSize: '2000px',
    },
  },
```


packages/core/src/definitions/styles.ts
```diff
@@ -120,42 +127,6 @@ export const builtInStyles: Partial<
     description: 'The spacing between the elements of the section',
     defaultValue: '0px',
   },
-  cfBackgroundImageUrl: {
```
elylucas, line 123: are we breaking anything backwards compat wise removing these?

primeinteger: Yes, I'm pretty sure these are breaking changes and would require all components with background images and all image components to be removed and re-added to the canvas to get the correct variable definitions

primeinteger: I think the background images are backwards compatible since the `cfBackgroundImageUrl` variable is staying in tact (I just moved it to the optional built in styles so it's not inherited on every custom component by default)

The `cfBackgroundImageAlignment` and `cfBackgroundImageScaling` variables have been removed and moved into the `cfBackgroundImageOptions` object, so these settings may revert to default options when the changes are released.

elylucas (approved):

PR #485, opened 2024-03-13
fix(canvas-interactions): proper placeholders based on padding and flex direction [ALT-505]

chasepoirier:
## Purpose
Placeholders were sometimes shown in the wrong location because they were not accounting for padding offsets or the flex alignment of the parent component. This PR adds logic to account for various alignments and padding values to more precisely render dropzone placeholders.

#### Additional Bug Fixes
- Not being able to enable nested dropzones because of new overlays
- Adjust usage of `pointer-events: none` for proper drag and non-drag state behavior

https://github.com/contentful/experience-builder/assets/23040747/471a5a28-b5b5-443f-aa2e-c95d8f4c2d36

### DIFF
--- a/packages/visual-editor/src/components/Draggable/Placeholder.tsx
+++ b/packages/visual-editor/src/components/Draggable/Placeholder.tsx
@@ -15,6 +15,61 @@ interface PlaceholderProps extends PlaceholderParams {
   id: string;
 }
 
+const calcOffsetLeft = (
+  parentElement: Element | null,
+  placeholderWidth: number,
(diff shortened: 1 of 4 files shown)


ryunsong-contentful (approved):
LGTM - one thing I noticed that might be worth talking about with Travis when he gets back or maybe tomorrow in standup is the highlights of the dropzone if it's a parent of other children dropzones. Aka - do we want to apply the same dropzone highlight over the entire dropzones including children dropzones of a parent or should the children dropzones be white

PR #488, opened 2024-03-14
fix(canvas-interactions): do not render hitboxes for single column [ALT-571]

chasepoirier:
## Purpose
The columns component does not allow dragging components in-between columns to add new columns. This is only handled via the column presets option in the design tab.

Because of this, we should not render hitboxes for the SingleColumn since they serve no purpose and only cause issues when many columns are added to the columns component. These hitboxes would overlap each other and end up making it seem like a user could not drop a component inside a column because that dropzone would never get enabled.

### DIFF
--- a/packages/visual-editor/src/types/constants.ts
+++ b/packages/visual-editor/src/types/constants.ts
@@ -12,7 +12,7 @@ export const CTFL_DRAGGABLE_ID = 'data-ctfl-draggable-id';
 export const CTFL_DRAGGING_ELEMENT = 'data-ctfl-dragging-element';
 
 export const HITBOX = {
-  WIDTH: 80,
+  WIDTH: 70,
   HEIGHT: 20,
   INITIAL_OFFSET: 10,
   OFFSET_INCREMENT: 8,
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/types/constants.ts
```diff
@@ -12,7 +12,7 @@ export const CTFL_DRAGGABLE_ID = 'data-ctfl-draggable-id';
 export const CTFL_DRAGGING_ELEMENT = 'data-ctfl-dragging-element';
 
 export const HITBOX = {
-  WIDTH: 80,
+  WIDTH: 70,
```
ryunsong-contentful, line 15: Is this a change we should discuss with the team about async in slack? Or maybe at least an FYI thing - two main modes of consideration I have here

1. This is a fundamental change to every component hence why it should be at least an FYI
2. While it's an impactful change, it's a minor difference numerically and is an easy two way street. So I think anything longer than a quick discussion would result in negative return on value relative to time spent.

ryunsong-contentful (approved):
Nice, tested it and is smooth as butter

packages/visual-editor/src/types/constants.ts
chasepoirier: Yeah good point, I can make a quick note in Slack. It's also just for the invisible hitboxes so no change in terms of the visible UI but does lead to a change in the feel for DnD

PR #490, opened 2024-03-14
fix: make images media type instead of text

ryunsong-contentful:

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -1,7 +1,7 @@
 import { ComponentDefinitionVariable, ContainerStyleVariableName } from '@/types';
 
 export const builtInStyles: Partial<
-  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Boolean'>>
+  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Media' | 'Boolean'>>
 > = {
   cfVerticalAlignment: {
     validations: {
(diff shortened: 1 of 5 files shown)


packages/core/src/definitions/styles.ts
```diff
@@ -452,7 +452,7 @@ export const singleColumnBuiltInStyles: Partial<
 };
 
 export const columnsBuiltInStyles: Partial<
-  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Boolean'>>
+  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Media' | 'Boolean'>>
```
ryunsong-contentful, line 455: Should be reverted

packages/core/src/definitions/styles.ts
```diff
@@ -313,7 +313,7 @@ export const containerBuiltInStyles: Partial<
 };
 
 export const singleColumnBuiltInStyles: Partial<
-  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Boolean'>>
+  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Media' | 'Boolean'>>
```
ryunsong-contentful, line 316: All non image related definitions should not contain media

packages/core/src/definitions/styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ComponentDefinitionVariable, ContainerStyleVariableName } from '@/types';
 
 export const builtInStyles: Partial<
-  Record<ContainerStyleVariableName, ComponentDefinitionVariable<'Text' | 'Boolean'>>
```
elylucas, line 4: Good catch getting all these instances of background images. I didn't think of that.

elylucas (changes requested):
I see the background images handled in here, but the `Image` component itself will also need its src url changed to media here: https://github.com/contentful/experience-builder/blob/development/packages/components/src/components/Image/index.ts#L25

PR #491, opened 2024-03-15
fix: locale switching does update content in canvas [SPA-1949]

anwaar931:
## Purpose
Remove useEffect in VisualEditorRoot as it was resetting entity store reactively. Which in some case ends up having empty canvas.


Spring3 (approved):
✅

dimitrycf:
For the future:

I checked and this useEffect (is run only upon mounting of the `VisualEditorRoot` which from my tests happens only one time at when the iframe loads for the first time.  

So it seems there's no need for that useEffect resetting locale to initial one. 


image

dimitrycf (commented):
✅

PR #492, opened 2024-03-15
fix: extend validators to validate component registration on editor mode [SPA-1899]

SofiaMargariti:
## Purpose

* Extend validations to make sure component `definitionId` doesn't not contain invalid characters and has max limit of 32
* Validate components during component registration - **only on editor mode**

## Approach
* Define a simple validator function for component registrations
* [In Editor Mode] Run the validation when a new component is registered
* throw an error if the validation fails
* Validation doesn't ran on preview/delivery mode to avoid introduction exception in the customer's website.

### DIFF
--- a/packages/validators/src/schemas/componentDefinition.ts
+++ b/packages/validators/src/schemas/componentDefinition.ts
@@ -1,0 +1,32 @@
+import { z } from 'zod';
+
+export const DefinitionPropertyTypeSchema = z.enum([
+  'Text',
+  'RichText',
+  'Number',
+  'Date',
+  'Boolean',
+  'Location',
+  'Media',
+  'Object',
+]);
+
+export const DefinitionPropertyKeySchema = z
+  .string()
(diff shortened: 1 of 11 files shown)


packages/validators/src/schemas/componentDefinition.ts
```diff
@@ -0,0 +1,32 @@
...
+  'Date',
+  'Boolean',
+  'Location',
+  'Media',
+  'Object',
+]);
+
+export const DefinitionPropertyKeySchema = z
+  .string()
+  .regex(/^[a-zA-Z0-9-_]{1,32}$/, { message: 'Attribute needs to match: /^[a-zA-Z0-9-_]{1,32}$/' });
```
Spring3, line 16: Attribute or Property?

Spring3:
Can we add tests to this?

ChidinmaOrajiaku (approved):
Would be nice to add tests to this

PR #493, opened 2024-03-15
fix(validators): add valuesByBreakpoint as to BoundValue, UnboundValue and ComponentValue [ALT-581]

primeinteger:
## Purpose

As part of updating the component definition variables for Image and Background Image, I want to add the `valuesByBreakpoint` property that is used by `DesignValue` to also be optionally allowed on `BoundValue`, `UnboundValue`, and `ComponentValue` types.

This will allow design style properties to be attached to these other types to condense the number of variables down in the component definition.

Screenshot 2024-03-13 at 6 04 17 PM


primeinteger:
Cancelling this change for now

PR #498, opened 2024-03-19
refactor(experience-builder-sdk): synchronous style calculation

cdun:
## Purpose

Styling in Experiences is currently a process that happens solely client-side. The hook used to inject styles, `useStyleTag`, uses a combination of state/effect to create the correct style rules and append them to the document head. Since React runs effects after DOM is updated, this approach has meant that there is always at least one render pass where the output is unstyled.

## Approach

Given that none of the work that `useStyleTag` does is asynchronous, we can refactor it process such that:

- Styles are memoised, rather than stateful; this better reflects the fact that they are a product of props rather than something the hook itself can change over time.
- Styles are appended the DOM using a layout effect. Given that these run before paint, this means that styles are immediately available for newly mounted components.


Spring3 (approved):
Works well! Thanks!

PR #499, opened 2024-03-19
fix: convert protocol agnostic image asset url to https [ALT-639]

primeinteger:
## Purpose

Changes the `getOptimizedImageUrl` method to specify https instead of using protocol agnostic urls that come back from the image API.

This prevents the "Mixed Content" error when running the hosted app locally on http. 

I tested images and background images with the test-app running locally, and with the hosted test-app on vercel and I'm not seeing any unexpected side effects.

### DIFF
--- a/packages/core/src/utils/transformers/media/getOptimizedImageUrl.ts
+++ b/packages/core/src/utils/transformers/media/getOptimizedImageUrl.ts
@@ -4,6 +4,10 @@ export function getOptimizedImageUrl(
   quality?: number,
   format?: string,
 ) {
+  if (url.startsWith('//')) {
+    url = 'https:' + url;
+  }
+
   const params = new URLSearchParams();
   if (width) {
     params.append('w', width.toString());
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/transformers/media/getOptimizedImageUrl.ts
```diff
@@ -4,6 +4,10 @@ export function getOptimizedImageUrl(
   quality?: number,
   format?: string,
 ) {
+  if (url.startsWith('//')) {
```
elylucas, line 7: nice and safe!

elylucas (approved):

elylucas:
I think your code will cover it, but probably good to test a manually bound image url as well

primeinteger:
> probably good to test a manually bound image url as well

Manual bound image urls don't go through this method, so those are served up exactly as the user entered them.  Would we want to alter a url that the user entered manually?

PR #502, opened 2024-03-19
fix(components): change image object fit to default to undefined instead of none []

chasepoirier:
## Purpose
Defaulting the image component to object fit none lead to unexpected behavior so we are opting to make objectFit undefined as an initial value so that it has no styling effect by default.


primeinteger (approved):
🏆

PR #504, opened 2024-03-20
fix(core): expect image quality to be a string with a percentage [ALT-643]

elylucas:
After the refactoring to reduce the number of variables, image quality was being passed down to the Experience with a percentage symbol, which before it was just a string with the number in it. Updated the code to expect the percentage and remove it before converting it to a number.

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -193,7 +193,7 @@ export const optionalBuiltInStyles: VariableDefinitions = {
       width: DEFAULT_IMAGE_WIDTH,
       height: '100%',
       objectPosition: 'center center',
-      quality: '100',
+      quality: '80%',
       targetSize: DEFAULT_IMAGE_WIDTH,
     },
   },
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -192,8 +192,6 @@ export const optionalBuiltInStyles: VariableDefinitions = {
     defaultValue: {
       width: DEFAULT_IMAGE_WIDTH,
       height: '100%',
-      objectPosition: 'center center',
-      quality: '80%',
       targetSize: DEFAULT_IMAGE_WIDTH,
     },
   },
@@ -209,7 +207,6 @@ export const optionalBuiltInStyles: VariableDefinitions = {
     defaultValue: {
       scaling: 'fill',
       alignment: 'left top',
-      quality: '100',
       targetSize: '2000px',
     },
   },
--- a/packages/core/src/utils/transformers/media/transformMedia.ts
+++ b/packages/core/src/utils/transformers/media/transformMedia.ts
@@ -71,7 +71,7 @@ export const transformMedia = (
       value = getOptimizedBackgroundImageAsset(
         asset.fields.file as AssetFile,
         width as string,
-        Number(options.quality),
+        options.quality,
         options.format as (typeof SUPPORTED_IMAGE_FORMATS)[number],
       );
       return value;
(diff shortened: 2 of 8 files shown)


packages/core/src/utils/transformers/media/transformMedia.ts
```diff
@@ -39,7 +39,7 @@ export const transformMedia = (
       value = getOptimizedImageAsset(
         asset.fields.file as AssetFile,
         options.targetSize as string,
-        Number(options.quality),
+        options.quality,
```
primeinteger, line 42: I'm seeing an inconsistency from the UI side.

The `cfImageOptions` variable starts with the quality value as `'100'`
* https://github.com/contentful/experience-builder/blob/b2117a703ae8b5ed603e27b481f4764b6e830a1c/packages/core/src/definitions/styles.ts#L197

In the UI, the `<ImageQualityInput>`updates the quality value using
```ts
onBlur={(newValue) => onChange(newValue)}
```

This makes it so Image components that haven't changed the quality setting will use `'100'` but Image components that do change the quality will have the format included in the value.

Screenshot 2024-03-20 at 6 44 01 AM

----

Do you think we should change the default quality value to `'100%'` on the SDK side so it has the format included in both cases?

Or should update the UI side of this to parse out the format when the quality input is changed?

primeinteger: As a side note, I wanted to switch quality to numeric when merging the image option variables, but the `<NumericUnitInput>` handles the value as a string and so keeping it the same as before was the path of least resistance
* https://github.com/contentful/experience-builder/pull/482/files#diff-1659ec2be3f58ed052d1e69cf3e41c3ec8b712dde447b853d562e472ce762afcL241-L246

elylucas: It works with or without the format now, but ya, being consistent is good. Since the NumericUnitInput contains the format in its change events, its good to have the format included in the string as it goes into it as well. Typically we don't need to worry about converting these values to numbers, but we have to here for some logic with the quality. I'll set the default to 80% (since thats in another ticket) and see if it causes any problems.

packages/core/src/definitions/styles.ts
```diff
@@ -193,7 +193,7 @@ export const optionalBuiltInStyles: VariableDefinitions = {
       width: DEFAULT_IMAGE_WIDTH,
       height: '100%',
       objectPosition: 'center center',
-      quality: '100',
+      quality: '80%',
```
primeinteger, line 196: Does this mean that images will default to 80% quality when the image optimization section in the UI is "collapsed"?  

We might need a UI change to incorporate the 80% so that we still default images to 100% quality, and only set the "initial value" to 80% when the image optimization section is expanded/enabled in the UI.

primeinteger (approved):

packages/core/src/definitions/styles.ts
elylucas: The desired default quality is 80%, is that still an issue in the UI?

primeinteger: I thought we wanted the quality to be 100% by default, and then set quality to 80% as the "initial" value only when image optimization is enabled.

If the default quality should be 80% even when image optimization is disabled this should accomplish that without any change needed to the UI

elylucas: Quality should be not set by default (which having it not set is the same as 100% practically), and 80% when image optimization is enabled.

I do have a PR incoming for the UI as I ran into some other issues, I'll send that over soonish.

primeinteger: Awesome work on these updates! 🎉 

This makes sense to start some options as not set by default.  Your UI changes are looking great, really cleans up the overall implementation quite a bit! 👏

PR #507, opened 2024-03-20
fix: fetch all referenced entities 

ChidinmaOrajiaku:
## Purpose
Previously we only fetched 100 referenced entities. So for experiences with more than 100 entities bound, we only render 100 entities.

This PR fetches all bound entities in chunks of 100. 

If we encounter a response too big error during fetch, we reduce the limit and try fetching again.

https://github.com/contentful/experience-builder/assets/30434146/0b86e386-76ca-4e31-a07c-68ea0e628dc2

### DIFF
--- a/packages/core/src/fetchers/fetchAllEntities.spec.ts
+++ b/packages/core/src/fetchers/fetchAllEntities.spec.ts
@@ -1,0 +1,87 @@
+import { entries } from '../test/__fixtures__/entities';
+import { fetchAllEntities } from './fetchAllEntities';
+
+import { describe, afterEach, it, expect, vi, Mock } from 'vitest';
+import { ContentfulClientApi } from 'contentful';
+
+const mockClient = {
+  getAssets: vi.fn(),
+  withoutLinkResolution: {
+    getEntries: vi.fn(),
+  },
+} as unknown as ContentfulClientApi<undefined>;
+
+describe('fetchAllEntities', () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should fetch all entities', async () => {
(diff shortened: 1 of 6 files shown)


packages/core/src/fetchers/fetchAllEntities.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      locale: 'en-US',
+      limit: 20,
+    };
+
+    await fetchAllEntities(params);
+
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenNthCalledWith(11, {
+      'sys.id[in]': params.ids,
+      locale: 'en-US',
+      skip: 90,
```
Spring3, line 59: is `skip` expected to be `90` here? It looks from the previous request that the fetch started from `skip: 0` and `limit: 20`.

I thought that after an error the next request should have parameters `skip: 0` and `limit: 10`. Unless I misunderstood something

Spring3: Ah, no, sorry, I understand it now.

100 / 10 = 10 per batch. 1 request failed , so 10 + 1 request is the last one to fetch the remaining 10 items. All good :D

Spring3 (approved):
Works magically. Love it! Also big thanks for the tests - really stellar piece of work!

https://github.com/contentful/experience-builder/assets/4171202/b71d298a-b659-48f9-9e7e-106014a6ffcd

packages/core/src/fetchers/fetchAllEntities.spec.ts
ChidinmaOrajiaku: It's 90 coz it failed once, then retried with a limit of `10`, hence skip being 90 at the 11th call

PR #514, opened 2024-03-22
chore: remove links to old docs in readme

Spring3:
Renamed "Experience Builder" to "Studio Expeirences"

Links pointing to the old documentation or wiki pages have been replaced with the link with latest docs


chirantan (approved):
LGTM!

Tested the new links and they work fine. Checked the README for other occurrences of "Experience Builder", found none other.

PR #518, opened 2024-03-24
Prep repo for stable GA release - [ALT-468]

elylucas:
## Purpose

This PR prepares the monorepo for the upcoming stable release.

## Approach

- The versioning of the project was changed to used fixed versioning, so each package will published using the same version number. This will help cut down confusion on which package is being used for debugging scenarios. The downside is that each package will get published wether there was a change in them or not. This shouldn't be too big of an issue, however, as the only package meant to be installed by users is `@contentful/experiences-sdk-react`.
  - The fixed versioning will also give us a changelog that will live at the root of the repo which will contain changes to all of the packages.
- The `next` branch will now do a 'beta' release instead of an 'alpha' release.
- The `main` branch will do a stable release with the latest tag.
- The readme.md was updated with info on how the branching and release process works.
- The create-experience-cli and storybook-addon packages were temporarily marked as private so they won't publish as a part of this main workflow. We'll need to add a new workflow that handles these outside of the fixed versioning scenario. These packages are out of date anyways and releasing a stable v1 of them wouldn't be helpful.
- Some actions in the github action workflows were updated to latest. These were throwing deprecation warnings in the builds because they were using an older NodeJS version. The warnings should now go away.
- Some git tags have been removed from the repo that were either not needed or would cause conflicts with upcoming releases. I suggest also removing these tags locally running the command `git fetch --prune origin "+refs/tags/*:refs/tags/*"`
- The `main` branch was updated to the latest from `next` and should be ready to merge into.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,7 +42,7 @@ jobs:
       - name: List components folder
         run: ls -la packages/components
       - name: Save Build folders        
-        uses: actions/cache/save@v3
+        uses: actions/cache/save@v4
         with:
           path: |
             packages/components/styles.css
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
           node-version: '20.11'
           cache: 'npm'
       - name: Restore Cypress Binary
-        uses: actions/cache/restore@v3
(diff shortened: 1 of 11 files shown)


primeinteger (approved):
👍

Queenkaz (commented):
L

PR #519, opened 2024-03-25
fix: fetch all references with includes

ChidinmaOrajiaku:
## Purpose
Fetch all references with includes

https://github.com/contentful/experience-builder/assets/30434146/1ecc9c3d-d1bc-42b4-84e7-b42ec9f03922

With Deep Binding

https://github.com/contentful/experience-builder/assets/30434146/6995a186-42a1-4f1e-8f7c-1eb8ec82238e

### DIFF
--- a/packages/core/src/fetchers/fetchAllEntities.spec.ts
+++ b/packages/core/src/fetchers/fetchAllEntities.spec.ts
@@ -1,0 +1,87 @@
+import { entries } from '../test/__fixtures__/entities';
+import { fetchAllEntities } from './fetchAllEntities';
+
+import { describe, afterEach, it, expect, vi, Mock } from 'vitest';
+import { ContentfulClientApi } from 'contentful';
+
+const mockClient = {
+  getAssets: vi.fn(),
+  withoutLinkResolution: {
+    getEntries: vi.fn(),
+  },
+} as unknown as ContentfulClientApi<undefined>;
+
+describe('fetchAllEntities', () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should fetch all entities', async () => {
+    (mockClient.withoutLinkResolution.getEntries as Mock).mockResolvedValue({
+      items: [...entries],
+      total: 100,
+    });
+
+    const params = {
+      ids: entries.map((entry) => entry.sys.id),
+      entityType: 'Entry' as 'Entry' | 'Asset',
+      client: mockClient,
+      locale: 'en-US',
+      limit: 20,
+    };
+
+    await fetchAllEntities(params);
+
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(5);
+  });
+
+  it('should reduce limit and refetch all entities if response error is gotten', async () => {
+    (mockClient.withoutLinkResolution.getEntries as Mock)
+      .mockRejectedValueOnce(new Error('Response size too big'))
+      .mockResolvedValue({
+        items: [...entries],
+        total: 100,
+      });
+
+    const params = {
+      ids: entries.map((entry) => entry.sys.id),
+      entityType: 'Entry' as 'Entry' | 'Asset',
+      client: mockClient,
+      locale: 'en-US',
+      limit: 20,
+    };
+
+    await fetchAllEntities(params);
+
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenLastCalledWith({
+      'sys.id[in]': params.ids,
+      locale: 'en-US',
+      skip: 90,
+      limit: 10,
+    });
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(11);
+  });
+
+  it('should never reduce limit to less than MIN_FETCH_LIMIT', async () => {
+    (mockClient.withoutLinkResolution.getEntries as Mock).mockRejectedValue(
+      new Error('Response size too big'),
+    );
+
+    const params = {
+      ids: entries.map((entry) => entry.sys.id),
+      entityType: 'Entry' as 'Entry' | 'Asset',
+      client: mockClient,
+      locale: 'en-US',
+      limit: 20,
+    };
+    await fetchAllEntities(params);
+
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenLastCalledWith({
+      'sys.id[in]': params.ids,
+      locale: 'en-US',
+      skip: 0,
+      limit: 1,
+    });
+    expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(5);
+  });
+});
--- a/packages/core/src/fetchers/fetchAllEntities.spec.ts
+++ b/packages/core/src/fetchers/fetchAllEntities.spec.ts
@@ -20,9 +20,19 @@ describe('fetchAllEntries', () => {
   });
 
   it('should fetch all entities', async () => {
-    (mockClient.withoutLinkResolution.getEntries as Mock).mockResolvedValue({
-      items: testEntries,
-      total: 100,
+    (mockClient.withoutLinkResolution.getEntries as Mock).mockImplementation((query) => {
+      const { limit, skip } = query;
+      const result = {
+        items: [...testEntries.slice(skip, skip + limit)],
+        includes: {
+          Entry: [createEntry('some-entry-id'), createEntry(`some-entry-id-${skip}`)],
+          Asset: [],
+        },
+        skip,
+        limit,
+        total: testEntries.length,
+      };
+      return result;
     });
 
     const params = {
@@ -33,17 +43,25 @@ describe('fetchAllEntries', () => {
       limit: 20,
     };
 
-    await fetchAllEntries(params);
+    const result = await fetchAllEntries(params);
 
     expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(5);
+    expect(result.includes.Entry).toHaveLength(6);
+    expect(result.includes.Asset).toHaveLength(0);
   });
 
   it('should reduce limit and refetch all entities if response error is gotten', async () => {
     (mockClient.withoutLinkResolution.getEntries as Mock)
       .mockRejectedValueOnce(new Error('Response size too big'))
-      .mockResolvedValue({
-        items: testEntries,
-        total: 100,
+      .mockImplementation((query) => {
+        const { limit, skip } = query;
+        const result = {
+          items: [...testEntries.slice(skip, skip + limit)],
+          skip,
+          limit,
+          total: testEntries.length,
+        };
+        return result;
       });
 
     const params = {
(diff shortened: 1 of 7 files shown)


dimitrycf:
Will properly test it after PH, but a quick open question: does it "deduplicate" the `includes{}` ? 

because, I wonder if each page will bring it's own `includes{}` with let's say `asset123` and then in theory if 10 pages bring in `asset123` as part of `includes{}` it will be in final `includes{}` 10 times?

ChidinmaOrajiaku:
> Will properly test it after PH, but a quick open question: does it "deduplicate" the `includes{}` ?
> 
> because, I wonder if each page will bring it's own `includes{}` with let's say `asset123` and then in theory if 10 pages bring in `asset123` as part of `includes{}` it will be in final `includes{}` 10 times?

This is up to the API, but in theory, I guess your thought is correct.

packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -0,0 +1,96 @@
...
+  locale: string;
+  skip?: number;
+  limit?: number;
+  responseItems?: Entry[];
+}) => {
+  try {
+    if (!ids.length || !client) {
+      return {
+        items: [],
+        ...(entityType === 'Entry' && { includes: [] }),
```
Spring3, line 43: Wouldn't it try to `...(false)` if `entityType === 'Asset'`?

ChidinmaOrajiaku: The check will be false so it won't spread the includes object

Spring3: Apparently in JS the case I was worried about is handled. 🤔 I thought that attempting to spread an boolean into an object would result in an error, but looks like it's not the case

Screenshot 2024-03-25 at 12 36 38

dimitrycf, line 43: Btw, `includes` seems to be an object with two keys `Entry` and `Asset` , not an array :) 
(i only noticed it on Friday, as I was assuming it was same shape as `.items[]`  🤷 ) 

image



image

packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export const fetchAllEntities = async ({
+  client,
+  entityType,
+  ids,
+  locale,
+  skip = 0,
+  limit = 100,
+  responseItems = [],
```
dimitrycf, line 29: We're passing recursively `reponseItems[]` as a variable which is aggregator for all pages. However, as we don't pass recursively `includes{}`, I wonder if we're aggregating referenced entries recursively?

dimitrycf (changes requested):
Hey Didi! 👋 

As I was trying to understand the algorithm I was making some notes, and noticed that the algorithm can be simplified and certain operations can be made explicit. 

Please have a look at this proposal for the refactoring and let me know what you think.

https://github.com/contentful/experience-builder/pull/522

packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -1,66 +1,125 @@
...
+
+    const {
+      items,
+      includes,
+      total: responseTotal,
+    } = await client.withoutLinkResolution.getEntries({ ...query });
+
+    responseItems.push(...(items as Entry[]));
+    responseIncludes?.Entry?.push(...(includes?.Entry || []));
+    responseIncludes?.Asset?.push(...(includes?.Asset || []));
```
dimitrycf, line 50: subjective: nit: 
From what I understand if `responseIncludes` is available, it _must_ already have `.Entry` and `.Asset` by contract. So secodn `?` is not necessary? 

eg. 
```
responseIncludes?.Entry.push(...(includes?.Entry || []))
```

?

packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -1,66 +1,125 @@
...
+        client,
+        ids,
+        locale,
+        skip,
+        limit: newLimit,
+        responseItems,
+      });
+    }
+
+    throw error;
```
dimitrycf, line 91: When I am reading this, this recursion can happen at the Level0 and at any other levels L1,L2 etc...
This implies that `size too big` message can happen on fetching page3 but not on page1 or page2. 
Also this recursion implies that the halving of the limit can happen recusively eg. `1000->500->250->125->62` (which I guess was the intent. And it would be great to have a test for such a recursive halving too).

But altogether my concern here is that this recursion is two things:
 - monitoring the limit problem 
 - gathering all entries/assets/items

packages/core/src/fetchers/fetchReferencedEntities.ts
```diff
@@ -56,8 +56,8 @@ export const fetchReferencedEntities = async ({
   }
 
   const [entriesResponse, assetsResponse] = (await Promise.all([
-    fetchAllEntities({ client, entityType: 'Entry', ids: entryIds, locale }),
-    fetchAllEntities({ client, entityType: 'Asset', ids: assetIds, locale }),
+    fetchAllEntries({ client, ids: entryIds, locale }),
+    fetchAllAssets({ client, ids: assetIds, locale }),
   ])) as unknown as [MinimalEntryCollection, AssetCollection];
```
dimitrycf, line 61: It would be great to get rid of this `as unknown` cast. 

Now it can be considered leftover from the time when I put it because I was fetching directly via ` client.withoutLinkResolution.getEntries({ ...query });` and this was the only way I found  to wrangle types. 

Now that we have control over `fetchAllEntries()`  we unlikely need it (and this cast is kinda dangerous)

dimitrycf (changes requested):

packages/core/src/fetchers/fetchAllEntities.ts
ChidinmaOrajiaku: I added tests for the recursive halving 

E.g https://github.com/contentful/experience-builder/pull/519/files#diff-8d5e914f2738a65c8156cb1a7020cb49fb345560da7f63c7a7a8773ea7f93aedR144

https://github.com/contentful/experience-builder/pull/519/files#diff-8d5e914f2738a65c8156cb1a7020cb49fb345560da7f63c7a7a8773ea7f93aedR41

packages/core/src/fetchers/fetchAllEntities.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+describe('fetchAllAssets', () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should fetch all assets', async () => {
+    (mockClient.getAssets as Mock).mockResolvedValue({
+      items: testAssets,
+      total: 100,
+    });
```
dimitrycf, line 103: Those tests can be made more reliable by making the mock of `getAssets/getEntries()` respect `skip` and `limit` parameters. 

Without respecting those parameters, it always returns 100 items and this is not how reality works. 

Feel free to copy/paste from this implementation that I made whilst testing the ticket https://github.com/contentful/experience-builder/blob/235f7c7f0d032f15a42810254fae324fd69a5d6d/packages/core/src/fetchers/fetchAllEntities.spec.ts#L24

dimitrycf (changes requested):

packages/core/src/fetchers/fetchAllEntities.ts
ChidinmaOrajiaku: Also, I don't understand your concern.

If the function is in the process of gathering and it gets the response size too big, it halves the limit else it continues gathering the entities.

dimitrycf (commented):
I have noticed that you intorduced separation between `fetchAllEntries()` and `fetchAllAssets()`. I tried that too on Friday (I tried refactoring, but got stuck with wrangling TypeScript types for an hour or so and eventually realized that rolling back was the easier way for a  Friday evening solution).

However, now that there are two functions, we have this pretty complicated logic of recursive halving duplicated 😱   that makes me scared! 

```
  } catch (error) {
    if (
      error instanceof Error &&
      error.message.includes('size too big') &&
      limit > MIN_FETCH_LIMIT
    ) {
      const newLimit = Math.max(MIN_FETCH_LIMIT, Math.floor(limit / 2));
      return fetchAllAssets({
        client,
        ids,
        locale,
        skip,
        limit: newLimit,
        responseItems,
      });
    }

    throw error;
  }
```

packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -0,0 +1,160 @@
...
+      items,
+      includes,
+      total: responseTotal,
+    } = await client.withoutLinkResolution.getEntries({ ...query });
+
+    responseItems.push(...(items as Entry[]));
+    responseIncludes?.Entry?.push(...(includes?.Entry || []));
+    responseIncludes?.Asset?.push(...(includes?.Asset || []));
+
+    if (skip + limit < responseTotal) {
```
dimitrycf, line 52: Maybe this is highly subjective, and maybe I am missing some obvious logic. But to me `skip + limit` sounds like `take age of a tree + multiply by the average speed of hamster running away from a snake and divide by latitude of Sydney` ... 

Again maybe it's just me, but `alreadyFetchedEntities + lastPageEntityCount > totalEntityCount` sounds more natural and intuitive 🤷

packages/core/src/fetchers/fetchAllEntities.spec.ts
```diff
@@ -33,17 +43,25 @@ describe('fetchAllEntries', () => {
       limit: 20,
     };
 
-    await fetchAllEntries(params);
+    const result = await fetchAllEntries(params);
 
     expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(5);
+    expect(result.includes.Entry).toHaveLength(6);
```
dimitrycf, line 49: It would be great to add a comment as to why we compare with 6, eg.

```
// Five pages will return those Entries
// Page 1: `some-entry-id` , `some-entry-id-0`
// Page 2: `some-entry-id` , `some-entry-id-20`
// Page 3: `some-entry-id` , `some-entry-id-40`
// Page 4: `some-entry-id` , `some-entry-id-60`
// Page 5: `some-entry-id` , `some-entry-id-80`
// As `result.includes.Entry` returns only unique entries, there are only 6 uniques as duplicates of `some-entry-id` are not returned
expect(result.includes.Entry).toHaveLength(6);
```

dimitrycf:
Visual validation 
https://www.loom.com/share/4371eb19d2bc4a97bb34e0f7c551cabf

packages/core/src/fetchers/fetchAllEntities.spec.ts
```diff
@@ -33,17 +43,25 @@ describe('fetchAllEntries', () => {
       limit: 20,
     };
 
-    await fetchAllEntries(params);
+    const result = await fetchAllEntries(params);
 
     expect(mockClient.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(5);
```
dimitrycf, line 48: Maybe we can add also test for total amount of items eg. 

```
const result = await fetchAllEntries(params);
expect(result.items).toHaveLength(100);
```

dimitrycf (changes requested):
Also would be great to add validation of total count to calls to :
 - `fetchAllEntries()` 
 - `fetchAllAssets()` 

something like this:

```
const result = await fetchAllEntries(params);
expect(result.items).toHaveLength(100);
```

dimitrycf (approved):
APPROVING TO UNBLOCK

(see the comments with suggestions)

PR #523, opened 2024-03-25
fix: add border radius to image component and make border radius an optional style

ryunsong-contentful:
## Purpose
Making border radius an optional style and fixing the issue where images don't take on the border radius value

Screenshot 2024-03-25 at 5 05 11 PM


primeinteger (approved):
💯

PR #529, opened 2024-03-27
feat: support style injection during ssr [SPA-1971]

Spring3:
Worth noting that the format of the `cfSsrClassname` variable is `string` while the correct format should be an `object` with `type` and other respective properties according to [ComponentVariableType](https://github.com/contentful/experience-builder/blob/development/packages/validators/src/schemas/v2023_09_28/experience.ts#L79) type

So before merging, we should align it for consistency


Spring3:
Closed in favor of #579

PR #540, opened 2024-04-02
feat: [SPA-1774] add hyperlink value to schema

loweisz:
## Purpose

Add the hyperlink value to the schema for the content-api to be ready to accept these values

### DIFF
--- a/packages/visual-editor/src/components/Dropzone/useComponentProps.ts
+++ b/packages/visual-editor/src/components/Dropzone/useComponentProps.ts
@@ -120,7 +120,7 @@ export const useComponentProps = ({
             ...acc,
             [variableName]: value,
           };
-        } else {
+        } else if (variableMapping.type === 'UnboundValue') {
           const value = getUnboundValues({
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/components/Dropzone/useComponentProps.ts
```diff
@@ -131,6 +131,8 @@ export const useComponentProps = ({
             ...acc,
             [variableName]: value,
           };
+        } else {
+          return { ...acc };
```
SofiaMargariti, line 135: Is this only temporary until we implement proper behavior for handling the hyperlink value?

SofiaMargariti (approved):
The schema looks good, just a small question

packages/visual-editor/src/components/Dropzone/useComponentProps.ts
loweisz: yes exactly, this will be implemented in the next PR

PR #572, opened 2024-04-17
fix: activate original dropzone when moving root-level components [ALT-746]

primeinteger:
## Purpose

This change fixes Issue 2 in [ALT-746](https://contentful.atlassian.net/browse/ALT-746) 🐛

I found that the issue was a result of setting the drag state via calling `setDraggingOnCanvas(true)` in the DNDProvider's `onBeforeCapture` - which is before the dimensions have been captured from the DOM based on the [docs](https://github.com/hello-pangea/dnd/blob/main/docs/guides/responders.md#life-cycle-%EF%B8%8F)

To fix the main issue, I moved this call to set the drag state in DNDProvider's `onBeforeDragStart` responder instead so that it won't happen until the DND capture phase has completed.

After that change, the dropzone where the root-level component was picked up from starts activating as expected when moving the dragged component around and then trying to move it back to its original location.

A new issue started happening -- when picking up a component on the canvas the `useDraggablePosition` hook was not adjusting the dragged component position to match the mouse coordinates offset as expected.  I made a small change in that hook to correct that.

https://github.com/contentful/experience-builder/assets/8539634/bbf5d06e-2086-4ff6-9101-44f171f52c10

Another thing to call out -- the blue Placeholder that renders when a dropzone activates does not render when moving a root-level component back to its original location. I looked into that and believe it's because the Placeholder is [part of the DraggableComponent](https://github.com/contentful/experience-builder/blob/55621c603646d8ac1c3da8083881fca5d4c9da5f/packages/visual-editor/src/components/Draggable/DraggableComponent.tsx#L117) which is picked up when dragging. So basically, the blue dropzone Placeholder is no longer on the canvas when you are dragging the component that it relates to.

I'm not exactly sure how to resolve that so I'll open a separate ticket for it.

[ALT-746]: https://contentful.atlassian.net/browse/ALT-746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/utils/styleUtils/stylesUtils.ts
+++ b/packages/core/src/utils/styleUtils/stylesUtils.ts
@@ -100,7 +100,7 @@ export const buildCfStyles = (
 };
 /**
  * Container/section default behavior:
- * Default height => height: EMPTY_CONTAINER_HEIGHT (120px)
+ * Default height => height: EMPTY_CONTAINER_HEIGHT
  * If a container component has children => height: 'fit-content'
  */
 export const calculateNodeDefaultHeight = ({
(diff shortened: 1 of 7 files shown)


ryunsong-contentful (approved):
Magic

PR #577, opened 2024-04-19
feat(components): add loading attribute to image component

elylucas:
## Purpose

Customers would like to add lazy loading to their images

## Approach

Extends the ImageOptimization options to include a loading flag that can be set to 'lazy', 'eager', or undefined, which is configured via the UI. This option will set its value to the `loading` attribute on the option.

### DIFF
--- a/packages/core/src/utils/transformers/media/getOptimizedImageAsset.spec.ts
+++ b/packages/core/src/utils/transformers/media/getOptimizedImageAsset.spec.ts
@@ -16,25 +16,25 @@ describe('transformImageAsset', () => {
 
   it('when width of image is 250, should return a srcSet with 2 parts', () => {
     file.details.image!.width = 250;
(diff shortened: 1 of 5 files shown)


primeinteger (approved):
Clean! 🔥

packages/core/src/utils/transformers/media/getOptimizedImageAsset.spec.ts
```diff
@@ -16,25 +16,25 @@ describe('transformImageAsset', () => {
 
   it('when width of image is 250, should return a srcSet with 2 parts', () => {
     file.details.image!.width = 250;
-    const result = getOptimizedImageAsset(file, '500px');
+    const result = getOptimizedImageAsset({ file, sizes: '500px' });
```
ryunsong-contentful, line 19: Could turn 500px into a constant or really the entire object ```{ file, sizes: '500px' }```

PR #578, opened 2024-04-23
fix: update dev script for validators package to match rest of packages

SofiaMargariti:
## Purpose

Update dev script for validators package to match rest of packages


primeinteger (approved):
🚀

primeinteger:
This change was implemented in #704 as well - so this PR could probably be closed at this point since it was never merged

PR #580, opened 2024-04-23
fix: render default content values for patterns [SPA-2048]

SofiaMargariti:
## Purpose

When rendering a pattern in the editor, as a standalone pattern, outside of an experience, we need to still render the default (placeholder) values into the canvas to assist designers when adjusting the pattern.

## Approach

This PR makes sure that when we encounter a property with type `ComponentValue` (only available for patterns) we render the default value of the component.

### Before
Screenshot 2024-04-24 at 11 02 58

### After
Screenshot 2024-04-24 at 11 03 59

### DIFF
--- a/packages/visual-editor/vitest.config.ts
+++ b/packages/visual-editor/vitest.config.ts
@@ -6,6 +6,7 @@ export default defineConfig({
   //@ts-expect-error ignore
   plugins: [tsconfigPaths()],
   test: {
+    globals: true,
     coverage: {
       reporter: ['text', 'json', 'html'],
       reportsDirectory: 'reports',
--- a/package-lock.json
+++ b/package-lock.json
@@ -43191,6 +43191,7 @@
         "@rollup/plugin-node-resolve": "^15.2.1",
         "@rollup/plugin-terser": "^0.4.3",
         "@rollup/plugin-typescript": "^11.1.3",
+        "@testing-library/react": "^15.0.4",
         "@types/lodash-es": "^4.17.12",
         "@types/react": "^18.2.15",
         "@types/react-dom": "^18.2.7",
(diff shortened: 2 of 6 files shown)


packages/visual-editor/vitest.config.ts
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
   //@ts-expect-error ignore
   plugins: [tsconfigPaths()],
   test: {
+    globals: true,
```
SofiaMargariti, line 9: [note for reviewer] allows to use `describe`, `it`, etc.. without having to import them everyime from vitest

Spring3:
@SofiaMargariti do we want to fix that for delivery/preview as well?

Screenshot 2024-04-26 at 15 02 52

SofiaMargariti:
> @SofiaMargariti do we want to fix that for delivery/preview as well?
> 
> At the moment it is fixed only for editor mode
> 
> Screenshot 2024-04-26 at 15 02 52

🤔 I can apply the same fix.. although should we support preview/delivery for patterns?

Spring3:
> > @SofiaMargariti do we want to fix that for delivery/preview as well?
> > At the moment it is fixed only for editor mode
> > Screenshot 2024-04-26 at 15 02 52
> 
> 🤔 I can apply the same fix.. although should we support preview/delivery for patterns?

Not certain myself, but also why not? Default values have been provided but we are not rendering them. If I created an experience and used this pattern, it would render a pattern with default values (in both - editor and preview/delivery). So why not do so for the pattern itself 🤔

Spring3 (approved):
Approving cause it works in editor mode, but I think we should do the same for preview/delivery modes

dimitrycf (changes requested):
I wonder if we should also render those values for P+D mode? (preview and delievery).

Here's a video explaining what i mean :)

https://www.loom.com/share/53b5e4520c4d4081aa5ae784e71de39f

dimitrycf (approved):
UNBLOCKING, re my and @Spring3  comments :)

SofiaMargariti:
I've updated the PR for preview/delivery mode as well, thanks everyone.

PR #587, opened 2024-04-26
fix(visual-editor): show canvas hover state in layers tab [ALT-787]

primeinteger:
## Purpose

When hovering over an element in the canvas, that element appears in a hover state in the layers tab.

https://github.com/contentful/user_interface/assets/8539634/57344a64-d883-437b-b044-c187e103a521


chasepoirier (approved):
🚀

PR #590, opened 2024-04-26
feat(validators): add optional displayName to ComponentTreeNode [ALT-780]

primeinteger:
## Purpose

Adds an optional `displayName` field to the `ComponentTreeNode` type.
* We can use this field to allow custom names to be set for each component instance on the canvas or layer.

### Reviewer

For a cleaner diff, use the 'Hide whitespace' option in the diff settings on the File Changes tab.

### DIFF
--- a/packages/validators/test/__fixtures__/v2023_09_28/experience.ts
+++ b/packages/validators/test/__fixtures__/v2023_09_28/experience.ts
@@ -246,7 +246,7 @@ export const experience = {
               },
               {
                 definitionId: 'button',
-                displayName: 'Button 1',
+                // Component node without optional `displayName`.
                 variables: {
                   label: {
                     path: '/5J8rtZN_nOJUx5YGdFuF0/fields/name/~locale',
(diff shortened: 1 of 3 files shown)


packages/validators/test/__fixtures__/v2023_09_28/experience.ts
```diff
@@ -244,6 +246,7 @@ export const experience = {
               },
               {
                 definitionId: 'button',
+                displayName: 'Button 1',
```
DanweDE, line 249: Can we have at least one element without `displayName` in this fixture?

```suggestion
                // Component node without optional `displayName`.
```

DanweDE:
I would say this PR/commit shouldn't be a `fix` but a `feat`.

SofiaMargariti (approved):

SofiaMargariti:
> I would say this PR/commit shouldn't be a `fix` but a `feat`.

Agree on this

PR #596, opened 2024-05-02
feat: resolve css variables in experiences

ryunsong-contentful:
## Purpose
Add css variable support for Experiences for Design Tokens

### Approach
Create an in memory resolved css variables values similar to design tokens. That way the ui can consume the values and store it in its immer store and thus make it accessible throughout the ui webapp

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -228,7 +228,7 @@ export type RecursiveDesignTokenDefinition = {
 };
 
 export type DesignTokensDefinition = {
-  spacing?: Record<string, string>;
+  spacing?: Record<string, string> | Record<string, number>;
   sizing?: Record<string, string>;
   color?: Record<string, string>;
   border?: Record<string, { width: string; style: 'solid' | 'dashed' | 'dotted'; color: string }>;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -228,7 +228,7 @@ export type RecursiveDesignTokenDefinition = {
 };
 
 export type DesignTokensDefinition = {
-  spacing?: Record<string, string> | Record<string, number>;
+  spacing?: Record<string, string>;
   sizing?: Record<string, string>;
   color?: Record<string, string>;
   border?: Record<string, { width: string; style: 'solid' | 'dashed' | 'dotted'; color: string }>;
--- a/packages/experience-builder-sdk/src/core/componentRegistry.ts
+++ b/packages/experience-builder-sdk/src/core/componentRegistry.ts
@@ -3,6 +3,7 @@ import type {
   ComponentRegistration,
   ComponentDefinition,
   ComponentRegistrationOptions,
+  DesignTokensDefinition,
 } from '@contentful/experiences-core/types';
 import {
   OUTGOING_EVENTS,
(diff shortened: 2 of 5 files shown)


chasepoirier (approved):
LGTM 🥇

packages/experience-builder-sdk/src/core/componentRegistry.ts
```diff
@@ -189,6 +192,115 @@ export const runRegisteredComponentValidations = () => {
...
+    lineHeight,
+    letterSpacing,
+    textColor,
+    border,
+  } = designTokensDefinition;
+  const resolvedCssVariables = {} as Record<string, string>;
+
+  // Create an element
+  const element = document.createElement('div');
+  document.body.appendChild(element);
```
primeinteger, line 239: Does this work with SSR? 🤔

packages/experience-builder-sdk/src/core/componentRegistry.ts
```diff
@@ -189,6 +192,115 @@ export const runRegisteredComponentValidations = () => {
...
+    Object.assign(resolvedCssVariables, rawResolvedValues);
+  }
+  if (letterSpacing) {
+    const rawResolvedValues = getAllCssVariableValues(element, letterSpacing, 'letter-spacing');
+    Object.assign(resolvedCssVariables, rawResolvedValues);
+  }
+  if (textColor) {
+    const rawResolvedValues = getAllCssVariableValues(element, textColor, 'color');
+    Object.assign(resolvedCssVariables, rawResolvedValues);
+  }
```
primeinteger, line 278: small nit - this could be refactored to use a loop to reduce repetition. for example:
```ts
const cssProperties = [
  { variable: sizing, property: 'width' },
  { variable: color, property: 'background-color' },
  { variable: borderRadius, property: 'border-radius' },
  { variable: fontSize, property: 'font-size' },
  { variable: lineHeight, property: 'line-height' },
  { variable: letterSpacing, property: 'letter-spacing' },
  { variable: textColor, property: 'color' },
];

cssProperties.forEach(({ variable, property }) => {
  if (variable) {
    const rawResolvedValues = getAllCssVariableValues(element, variable, property);
    Object.assign(resolvedCssVariables, rawResolvedValues);
  }
});
```

packages/core/src/types.ts
```diff
@@ -228,7 +228,7 @@ export type RecursiveDesignTokenDefinition = {
 };
 
 export type DesignTokensDefinition = {
-  spacing?: Record<string, string>;
+  spacing?: Record<string, string> | Record<string, number>;
```
primeinteger, line 231: This would also work here:
```suggestion
  spacing?: Record<string, string | number>;
```

packages/experience-builder-sdk/src/core/componentRegistry.ts
Spring3: I would also discourage from trying to affecting the DOM from within a plain js function.

I think react would expect it to be done in the `useEffect`

ryunsong-contentful: It is compatible with SSR, @Spring3 can you explain what the difference in using the useEffect does? I'm only temporarily creating it to grab the css values and then deleting the element.

packages/core/src/types.ts
ryunsong-contentful: That is what I initially tried but you get this in the build

```
  [!] (plugin typescript) RollupError: @rollup/plugin-typescript TS2345: Argument of type 'Record<string, string | number>' is not assignable to parameter of type 'Record<string, string>'.
```

primeinteger (approved):
I'm still not sure if the `document.createElement` approach will work in SSR-land, but approving either way and we can look at that later if it comes up.

PR #597, opened 2024-05-03
feat(experiences): implement layer renaming [ALT-776]

chasepoirier:
## Purpose
Mini PR to add typings to core for layer renaming

Ticket: https://contentful.atlassian.net/browse/ALT-776

### DIFF
--- a/packages/test-app/src/eb-config.ts
+++ b/packages/test-app/src/eb-config.ts
@@ -8,6 +8,7 @@ defineComponents([
     component: ComponentWithChildren,
     definition: {
       id: 'component-with-children',
+      thumbnailUrl: 'https://static.thenounproject.com/png/2397501-200.png',
       children: true,
       name: 'Children Enabled',
       category: 'Custom Components',
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -151,6 +151,7 @@ export type ExperienceTreeNode = {
       componentId: string;
       nodeLocation: string | null;
     };
+    displayName?: string;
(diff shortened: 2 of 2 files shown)


packages/test-app/src/eb-config.ts
```diff
@@ -8,6 +8,7 @@ defineComponents([
     component: ComponentWithChildren,
     definition: {
       id: 'component-with-children',
+      thumbnailUrl: 'https://static.thenounproject.com/png/2397501-200.png',
```
ryunsong-contentful, line 11: Can we get an internal svg and reference that instead of a static url? This will ensure that deletion of the svg/url will not affect and cause sudden breaking "ui" icons.

primeinteger: Right now we only use svg files for our built-in components.  We could upload this file to a Prod contentful space and reference an "in house" URL.  Since thenounproject.com is in the business of providing icons, I don't think we'd see an issue though, and any issue would only affect our own development test app and not the SDK or product itself so it's probably fine to use this URL in my opinion

primeinteger (approved):

packages/test-app/src/eb-config.ts
chasepoirier: Yeah agreed with @primeinteger on this one. This is just testing our internal custom component with a thumbnail URL in a similar way that a customer would do it so.

It did make me think though, maybe we could work to add support of local files like svgs/assets for the thumbnail in the future 🤔

PR #600, opened 2024-05-07
fix: not generate styles on csr if generated on ssr [SPA-2049]

Spring3:


packages/experience-builder-sdk/src/hooks/useClassName.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @returns the className that was eitner generated on the client side or derived from the value, set on server side
+ */
+export const useClassName = ({
+  node,
+  props,
+}: {
+  node: ComponentTreeNode;
+  props: Record<string, any>;
+}) => {
+  const [className, setClassName] = useState<string>(props.cfSsrClassName ?? '');
```
loweisz, line 26: I don't think you need a useState/useEffect here. This could be a derived `useMemo` or am I missing something here?

loweisz (approved):
Looks good to me, just a small potential suggestion

packages/experience-builder-sdk/src/hooks/useClassName.ts
Spring3: Maybe the bit that you have missed is that a className can be set from the `useLayoutEffect` [below](https://github.com/contentful/experience-builder/pull/600/files#diff-05c7e60264f1b70aebdc0607d81cc8c661cfc18d607fce7048d623c9401ed7b3R55)

I wouldn't convert the entire thing into `useMemo` because `useLayoutEffect` performs DOM manipulation by creating a `style` tag and adding to the head of the page.

PR #602, opened 2024-05-10
chore: add example astrojs with experiences app []

chasepoirier:
## Purpose
Include a sample [Astro](https://astro.build/) project that is configured to use Studio Experiences


package.json
```diff
@@ -46,7 +46,8 @@
     "lint-staged": "^15.0.1",
     "nx": "^18.0.4",
     "nx-cloud": "latest",
-    "prettier": "3.2.5"
+    "prettier": "3.2.5",
+    "prettier-plugin-astro": "^0.13.0"
```
primeinteger, line 50: Just confirming, does the main `package.json` file need to include `prettier-plugin-astro`?

It looks like that change was made to `examples/astrojs/package.json`, but I'm not sure that this needs to be added to the main `package.json` as well.

primeinteger (approved):
I tested the astro example app locally and its working quite well! 🎉

PR #619, opened 2024-05-21
fix: post version script to update sdk version with latest [SPA-2018]

anwaar931:
## Purpose
sdkVersion.ts always have previous published version. 

## Approach
In Github action `Version and publish` we run `lerna publish` which will also run `lerna version` as part of its process before publishing. 

`lerna version` uses [npm-version](https://docs.npmjs.com/cli/v9/commands/npm-version) which have a lifecycle `postversion` which have access to latest updated version. This way, we will update the latest version in `sdkVersion.ts` before publishing. We need to test it, hoping it will work.


SofiaMargariti:
Thanks @anwaar931, I am not really sure how to test, I think one step is to add `fix/npm-version-out-of-sync` in the `allowBranch` [here](https://github.com/contentful/experience-builder/blob/development/lerna.json#L7). But I don't know if that would produce a `dev`, `beta` or production version, I think we'd need a `dev` here. Maybe @elylucas you know where that is configured?

PR #623, opened 2024-05-23
Draft: SDK Version Update

elylucas:


anwaar931:
Hey @elylucas, I checked in the colorfulful-coin-eb project, `node_modules/experiences-sdk-react/package.json` have the latest version. But in sdkVersion.ts -> `SDK_VERSION` is not updated. Instead of injecting the version using `process.env.npm_package_version`. How about we do this

`src/sdkVersion.ts`
```
import { version } from '../package.json'

export const SDK_VERSION = version;
```

This will get us the latest version from package.json which lerna has updated itself before publishing. e.g commit like [this](https://github.com/contentful/experience-builder/commit/fb5018b8e21cba21386cd2d1e2d089b43e05c6d6)

What are your thoughts?

elylucas:
> Sorry we tried this approach but it didn't work either.
> 
> Hey @elylucas, I checked in the colorfulful-coin-eb project, `node_modules/experiences-sdk-react/package.json` having the latest version. But in sdkVersion.ts -> `SDK_VERSION` is not updated. Instead of injecting the version using `process.env.npm_package_version`. How about we do this
> 
> `src/sdkVersion.ts`
> 
> ```
> import { version } from '../package.json'
> 
> export const SDK_VERSION = version;
> ```
> 
> This will get us the latest version from package.json which lerna has updated itself before publishing. e.g commit like [this](https://github.com/contentful/experience-builder/commit/fb5018b8e21cba21386cd2d1e2d089b43e05c6d6)

Thanks Anwar. I've updated the PR with a solution, and tested it well through a pipeline and it appears to work well. I used the `postversion` lifecycle script you mentioned and that provided a bit cleaner solution than breaking a part the lerna version and publish scripts.

packages/experience-builder-sdk/bin/searchAndReplaceSdkVersion.cjs
```diff
@@ -0,0 +1,35 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires, no-undef
+const fs = require('fs');
+// eslint-disable-next-line @typescript-eslint/no-var-requires, no-undef
+const { glob } = require('glob');
+
+// The function `searchAndReplaceSdkVersion` is used to find instances of the SDK version in the built dist folder
+// and replace them with the current version. This is typically used during the build process to ensure
+// that the correct version number is reflected in the final build.
+searchAndReplaceSdkVersion;
```
anwaar931, line 9: is this needs to be declared here?

anwaar931 (approved):

packages/experience-builder-sdk/bin/searchAndReplaceSdkVersion.cjs
elylucas: no, not sure what that was about :)

ethan-ozelius-contentful (approved):

PR #625, opened 2024-05-24
RETIRED: Fix/fallback missing components [SPA-2033]

dimitrycf:
## Purpose

> Relevant [user_interface PR](https://github.com/contentful/user_interface/pull/21393)

Instead of showing fatal error upon encountering missing component, just render placholder. Which would still allow the Editor to see the component in the canvas in EDITOR mode, as well as act upon canvas to solve the situation.

## Approach

The approach taken here is to introduce new component `<SelectableComponent>` which is supposed to be a younger and poorer brother of `<DraggableComponent>` with all of the DnD functionality stripped out. 

The assumption behind this choice was that - enhancing `<DraggableComponent>` with additional functionality will make it even more complicated. It already has so many variables/props which it depends on, that it's not that easy to follow what is happening there.

### DIFF
--- a/packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
+++ b/packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
@@ -1,0 +1,78 @@
+import React, { useRef } from 'react';
+import { CSSProperties, ReactNode, SyntheticEvent } from 'react';
+import classNames from 'classnames';
+import styles from './styles.module.css';
+import { Rect } from '@components/Draggable/canvasToolsUtils';
+import Placeholder, { PlaceholderParams } from './Placeholder';
+import { useDraggedItemStore } from '@/store/draggedItem';
+import Tooltip from './Tooltip';
+
+interface SelectableComponentProps {
+  placeholder: PlaceholderParams;
+  wrapperProps: Record<string, string | undefined>;
+  tooltipLabel: string;
+  children: ReactNode;
+  id: string;
+  isAssemblyBlock?: boolean;
+  isSelected?: boolean;
+  onClick?: (e: SyntheticEvent) => void;
+  onMouseDown?: (e: SyntheticEvent) => void;
+  onMouseUp?: (e: SyntheticEvent) => void;
+  onMouseOver?: (e: SyntheticEvent) => void;
+  onMouseOut?: (e: SyntheticEvent) => void;
+  coordinates: Rect | null;
+  isContainer: boolean;
+  blockId?: string;
+  style?: CSSProperties;
+}
+
+export const SelectableComponent: React.FC<SelectableComponentProps> = ({
+  tooltipLabel,
+  children,
+  id,
+  isAssemblyBlock = false,
+  isSelected = false,
+  onClick = () => null,
+  onMouseOver = () => null,
+  coordinates,
+  style,
+  wrapperProps,
+  isContainer,
+  blockId,
+  placeholder,
+  ...rest
+}) => {
+  const ref = useRef<HTMLElement | null>(null);
+  const isHoveredComponent = useDraggedItemStore((state) => state.hoveredComponentId === id);
+
+  return (
+    <div
+      data-ctfl-draggable-id={id}
+      data-test-id={`draggable-${blockId ?? 'node'}`}
+      ref={(refNode) => {
+        ref.current = refNode;
+      }}
+      {...wrapperProps}
+      {...rest}
+      className={classNames(styles.DraggableComponent, wrapperProps.className, {
+        [styles.isAssemblyBlock]: isAssemblyBlock,
+        [styles.isSelected]: isSelected,
+        [styles.isHoveringComponent]: isHoveredComponent,
+      })}
+      style={{
+        ...style,
+      }}
+      onMouseOver={onMouseOver}
+      onClick={onClick}>
+      <Tooltip
+        id={id}
+        coordinates={coordinates}
+        isAssemblyBlock={isAssemblyBlock}
+        isContainer={isContainer}
+        label={tooltipLabel}
+      />
+      <Placeholder {...placeholder} id={id} />
+      {children}
+    </div>
+  );
+};
--- a/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
+++ b/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
@@ -89,7 +89,16 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
   };
 
   if (isComponentMissing || !definition || !elementToRender) {
-    return <div style={{ border: '1px solid gold' }}>Missing component</div>;
+    return (
+      <div
+        style={{
+          border: '1px solid red',
+          width: '100%',
+          height: '100%',
+        }}>
+        Missing component
+      </div>
+    );
   }
 
   if (isSingleColumn) {
--- a/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
+++ b/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
@@ -41,12 +41,13 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
 }) => {
   const setSelectedNodeId = useEditorStore((state) => state.setSelectedNodeId);
   const selectedNodeId = useEditorStore((state) => state.selectedNodeId);
-  const { node, componentId, wrapperProps, definition, elementToRender } = useComponent({
-    node: rawNode,
-    resolveDesignValue,
-    renderDropzone,
-    userIsDragging,
-  });
+  const { isComponentMissing, node, componentId, wrapperProps, definition, elementToRender } =
+    useComponent({
+      node: rawNode,
+      resolveDesignValue,
+      renderDropzone,
+      userIsDragging,
+    });
 
   const coordinates = useSelectedInstanceCoordinates({ node });
 
@@ -86,6 +87,10 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
       nodeId: componentId,
     });
   };
+
+  if (isComponentMissing || !definition || !elementToRender) {
+    return <div style={{ border: '1px solid gold' }}>Missing component</div>;
+  }
 
   if (isSingleColumn) {
     return (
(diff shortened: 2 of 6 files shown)


packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -53,9 +54,10 @@ export const useComponent = ({
     }
 
     if (!registration) {
-      throw Error(
-        `Component registration not found for component with id: "${node.data.blockId}". The component might of been removed. To proceed, remove the component manually from the layers tab.`,
+      console.warn(
+        `;;Component registration not found for component with id: "${node.data.blockId}". The component might of been removed. To proceed, remove the component manually from the layers tab.`,
```
elylucas, line 58: are the double semis (';;') here a typo or needed?

packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -67,30 +69,37 @@ export const useComponent = ({
     areEntitiesFetched,
     resolveDesignValue,
     renderDropzone,
-    definition: componentRegistration.definition,
+    definition: componentRegistration?.definition,
     userIsDragging,
   });
 
+  const createElementToRender = (componentRegistration: ComponentRegistration) => {
```
elylucas, line 76: We definitely have some changes to this same part of the code incoming, so there might be some MCs to work through. Should be fine though.

packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -67,30 +69,37 @@ export const useComponent = ({
...
+    : null;
 
   return {
+    isComponentMissing: !componentRegistration,
     node,
     componentId,
     elementToRender,
     wrapperProps,
-    definition: componentRegistration.definition,
+    definition: componentRegistration ? componentRegistration.definition : null,
```
elylucas, line 103: possibly could be shortened to `definition: componentRegistration?.definition,`

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
```diff
@@ -87,6 +95,36 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
...
 
+  if (isComponentMissing || !definition || !elementToRender) {
+    const renderMissingComponentPlacehoder = () => (
+      <div
+        style={{
+          border: '1px solid red',
+          width: '100%',
+          height: '100%',
+        }}>
+        Missing component &lt;{node.data.blockId}&gt;
```
elylucas, line 106: Putting the angle brackets around the name of the component might make the devs think this is how it is referenced in React, which would be incorrect. I wonder if we should just show the name here minus the brackets

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
```diff
@@ -87,6 +95,36 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
     });
   };
 
+  if (isComponentMissing || !definition || !elementToRender) {
+    const renderMissingComponentPlacehoder = () => (
```
elylucas, line 99: I think this div and related styles could live in the SelectedComponentPlaceholder itself, and just the pass the children into it as a prop.

packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -67,30 +69,37 @@ export const useComponent = ({
...
+          React.createElement(componentRegistration.component, { ...dragProps, ...componentProps })
+      : node.type === ASSEMBLY_NODE_TYPE
+        ? // Assembly.tsx requires renderDropzone and editorMode as well
+          () => React.createElement(componentRegistration.component, componentProps)
+        : () =>
+            React.createElement(
+              ImportedComponentErrorBoundary,
+              null,
+              React.createElement(componentRegistration.component, otherComponentProps),
+            );
```
elylucas, line 88: As an alternative to the method you already have, I think the desired functionality could also be done here by adding another elseif check if the definition is missing, and if so, render the SelectableComponent instead of the componentRegistration.component.

I think if you did so, the SelectableComponent wouldn't need any of the complexity it currently has to support being a draggable, as it would get wrapped in the DraggableComponent just like any other component.

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
elylucas, line 1: I think this file could be name to be a bit more descriptive of what it is doing. Something like "MissingComponentWarning.tsx" or such, assuming displaying missing component information is the only thing this component does.

elylucas (approved):
Testing and it works great. I left a few comments about the approach and some small code style tweaks. I'll go ahead and approve as it works as intended and I don't see anything wrong with the approach. Let me know if you want to discuss the alternative approach I mentioned any more.

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+      onMouseOver={onMouseOver}
+      onClick={onClick}>
+      <Tooltip
+        id={id}
+        coordinates={coordinates}
+        isAssemblyBlock={isAssemblyBlock}
+        isContainer={isContainer}
+        label={tooltipLabel}
+      />
+      <Placeholder {...placeholder} id={id} />
```
Spring3, line 74: personal preference I guess: would it make sense to pick it out of this component and pass the tooltip, placeholder together with {children}

Eg instead of 
```
<SelectableComponent>
   <button>
</SelectableComponent>
```

have it as the following:

```
<SelectableComponent>
  <Tooltip />
  <Placeholder />
  <button />
</SelectableComponent>
```

A bit more to write, but it helps remove a couple of objects from the props of `SelectableComponent` and simplify the logic as a result

packages/visual-editor/src/components/Dropzone/EditorBlockClone.tsx
```diff
@@ -53,6 +53,10 @@ export const EditorBlockClone: React.FC<EditorBlockCloneProps> = ({
   const isAssemblyBlock = node.type === ASSEMBLY_BLOCK_NODE_TYPE;
   const isSingleColumn = node.data.blockId === CONTENTFUL_COMPONENTS.singleColumn.id;
 
+  if (isComponentMissing || !elementToRender) {
```
Spring3, line 56: Isn't `isComponentMissing` and `!elementToRender` sort of the same thing aliased twice? Maybe we could just use one of them?

packages/visual-editor/src/components/Dropzone/useComponent.ts
Spring3: Can I please ask to break this one liner into a more human readable form? As much as one liners are fun, I'd prefer having ifs and proper intermediate steps named well with variables to help understand it better.

Also by writing this code this way you are promoting the rest of the devs in future (who might need to edit it) to follow this style.

dimitrycf:
## ⚠️ RETIRED in favor of https://github.com/contentful/experience-builder/pull/658

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
dimitrycf: ✅ FIXED (in new PR)

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
dimitrycf: ✅  this component retired

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }}
+      {...wrapperProps}
+      {...rest}
+      className={classNames(styles.DraggableComponent, wrapperProps.className, {
+        [styles.isAssemblyBlock]: isAssemblyBlock,
+        [styles.isSelected]: isSelected,
+        [styles.isHoveringComponent]: isHoveredComponent,
+      })}
+      style={{
+        ...style,
```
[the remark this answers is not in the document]
dimitrycf: Hmm... I think Chase can answer this question https://github.com/contentful/experience-builder/commit/2a663ebd7aa355948a332b07235aee3080a8c8dd#diff-a1f5809c00fe34301983a8b59d43e5de9d72ad1c5cc7108ef10108e0a8db0df2R74

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <div
+      data-ctfl-draggable-id={id}
+      data-test-id={`draggable-${blockId ?? 'node'}`}
+      ref={(refNode) => {
+        ref.current = refNode;
+      }}
+      {...wrapperProps}
+      {...rest}
```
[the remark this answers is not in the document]
dimitrycf: Same here https://github.com/contentful/experience-builder/commit/2a663ebd7aa355948a332b07235aee3080a8c8dd#diff-a1f5809c00fe34301983a8b59d43e5de9d72ad1c5cc7108ef10108e0a8db0df2R74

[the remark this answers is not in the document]
dimitrycf: And `<SelectableComponent>` is retired now anyways.

packages/visual-editor/src/components/Draggable/SelectableComponent.tsx
dimitrycf: `<SelectableComponent>` is now retired

packages/visual-editor/src/components/Dropzone/EditorBlockClone.tsx
dimitrycf: ✅  FIXED! Thanks for noticing!

dimitrycf:
@Spring3 I think there was some off-by-one glitch by github and it highlighted some LOC which I didn't touch.

But I see your point, this onliner confused me too (create element within an element within an element ... INCEPTION 😮 ) 

Fixed in new PR 
85794d08f30207553a4c43e1d2082f910a657d63

PR #634, opened 2024-05-30
Allow package versions to be manually published as pre-releases for a selected branch

jsdalton:
## Purpose

* Contributors from time to time want to publish specific builds to the public repo so they can be installed elsewhere (eg in user interface PRs where dependent changes are made)

## Approach

* Use existing workflow dispatch on main branch and add a toggle that allows a prerelease to be built for the selected branch
* Artifact uses same naming convention and lerna option as existing prerelease for `development` branch, except the preid is `prerelease` instead of `dev`. The dist-tag assigned is `prerelease`

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -80,8 +80,7 @@ jobs:
           git config user.email "${{ github.actor}}@users.noreply.github.com"
 
           if [ ${{ inputs.publish-prerelease }} ]; then
-            # temporarily echoing to verify that publish command is correct and correctly rendered
-            echo npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid prerelease-$(date +%Y%m%dT%H%M)-$(git rev-parse HEAD | cut -c1-7) --no-changelog --no-push --yes --no-private --allow-branch ${{ github.ref_name }}
+            npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid prerelease-$(date +%Y%m%dT%H%M)-$(git rev-parse HEAD | cut -c1-7) --no-changelog --no-push --yes --no-private --dist-tag prerelease --allow-branch ${{ github.ref_name }}
           elif [ ${{ github.ref_name }} = development ]; then
             npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid dev-$(date +%Y%m%dT%H%M)-$(git rev-parse HEAD | cut -c1-7) --dist-tag dev --no-changelog --no-push --yes --no-private 
           elif [ ${{ github.ref_name }} = next ]; then
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -2,6 +2,12 @@ name: Publish packages
 
 on:
   workflow_call:
+    inputs:
+      publish-prelease:
+        description: 'Publish prerelease version of the current ref'
+        default: false
+        required: false
+        type: boolean
     secrets:
       VAULT_URL:
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -73,14 +79,18 @@ jobs:
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor}}@users.noreply.github.com"
 
-          if [ ${{ github.ref_name }} = development ]; then
-            npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid dev-$(date +%Y%m%dT%H%M)-$(git rev-parse HEAD  | cut -c1-7) --dist-tag dev --no-changelog --no-push --yes --no-private 
+          if [ ${{ inputs.publish-prerelease }} ]; then
+            # temporarily echoing to verify that publish command is correct and correctly rendered
+            echo npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid prerelease-$(date +%Y%m%dT%H%M)-$(git rev-parse HEAD | cut -c1-7) --no-changelog --no-push --yes --no-private --allow-branch ${{ github.ref_name }}
```
elylucas, line 84: I think we can add this in for real and test it by running this workflow against this branch (feature/publish-manual-workflow).

elylucas: Looks like the --dist-tag option got left off. We'll want that there as either 'dev' or 'prerelease' so it won't accidentally get published as 'latest'.

elylucas: Not particularrily against it, but I was wondering if changing the preid to 'prerelease' from 'dev' was intentional. If so, we'd probably want the dist tag (mentioned above) to be 'prerelease' as well. We'd also need to communicate/document the differences between a prerelease and dev builds.

elylucas (changes requested):
Overall looks great! Pretty clean and eloquent. I did notice something left off that I think needs to be added back then, but besides that, only a few discussion comments, but no real changes from those unless you think so.

.github/workflows/publish.yaml
jsdalton: Great points there. In order:

* Nice! I didn't appreciate that I could just choose this branch and it would run the workflow _from_ this branch (not just _against_ this branch). I will test a bit more and if I have confidence will remove the safety check.
* This was intentional but I'm glad you caught this since I overlooked it defaults to `latest`. I propose that we do not want to use `dev` here because then it would overwrite the current `dev` tag (which should act as a pointer to the latest package published in the development branch). I would propose `prerelease` for now -- though this would not be a stable tag as it would always point to whatever  As an alternative we could go a step further and create a dist tag such as `pr-3059` but I didn't want to pollute the public tags list with a bunch of garbage.
* To your last point, `prerelease` was intentional since `dev` is for packages that are already merged to development (and sort of in the pipeline for eventual release). I wonder what a clearer word would be. "prerelease" is close, "experimental" doesn't quite get it either. i actually think "development" is best but we're squatting on that for work tied to the development branch. Agree about clarifying with a comment but open to naming / label suggestions

elylucas: Cool, thats reasonable.

jsdalton: Did a "test" worfklow run against the previous branch: https://github.com/contentful/experience-builder/actions/runs/9322741240 . Job ran successfully: https://github.com/contentful/experience-builder/actions/runs/9322741240/job/25664624309 . Command it issues looks legit:

`npx lerna publish --conventional-commits --conventional-prerelease --exact --force-publish --preid prerelease-20240531T1839-91b899d --no-changelog --no-push --yes --no-private --allow-branch feature/publish-manual-workflow`

(this is before my recent change adding the missing `dist-tag`

elylucas (approved):
🚀

PR #642, opened 2024-06-03
feat: allow for optional composite design token values in experiences

ryunsong-contentful:
## Purpose
To make composite values optional when defining design tokens like border where it consists of a width, color, and style property.

## Approach
When creating the in memory design token structure, make sure the values are all set during construction. Only composite design tokens like border and eventually text need to be checked because simple key-value DTs are enforced by typescript.
Use `1px` as a fallback default for width
Use `black` as a fallback default for color
Use `solid` as a fallback default for style

## Testing
```
Used this for my DTs where I had a width, color, and style missing
// register design tokens
defineDesignTokens({
  border: {
    Azure: { style: 'solid', color: 'azure' },
    Card: { width: '1px', style: 'solid' },
    Carousel: {},
    Hero: { width: '2px', color: '#ffaabb' },
  },
});
```
which then produced this in the console log on the ui side.
Screenshot 2024-06-03 at 2 59 42 PM

### DIFF
--- a/packages/core/src/registries/designTokenRegistry.ts
+++ b/packages/core/src/registries/designTokenRegistry.ts
@@ -3,13 +3,33 @@ import { builtInStyles, optionalBuiltInStyles } from '../definitions/styles';
 
 export let designTokensRegistry: DesignTokensDefinition = {};
 
+// This function is used to ensure that the composite values are valid since composite values are optional.
+// Therefore only border and in the future text related design tokens are/will be checked in this funciton.
+// Ensuring values for simple key-value design tokens are not neccessary since they are required via typescript.
+const ensureValidCompositeValues = (designTokenDefinition: DesignTokensDefinition) => {
+  // TODO: add validation logic when text related design tokens are added
+
+  // Border validation
+  if (designTokenDefinition.border) {
+    for (const borderKey in designTokenDefinition.border) {
+      const borderValue = designTokenDefinition.border[borderKey];
+      designTokenDefinition.border[borderKey] = {
+        width: borderValue.width || '1px',
+        style: borderValue.style || 'solid',
+        color: borderValue.color || '#000000',
+      };
+    }
+  }
+  return designTokenDefinition;
+};
+
 /**
  * Register design tokens styling
  * @param designTokenDefinition - {[key:string]: Record<string, string>}
  * @returns void
  */
 export const defineDesignTokens = (designTokenDefinition: DesignTokensDefinition) => {
-  Object.assign(designTokensRegistry, designTokenDefinition);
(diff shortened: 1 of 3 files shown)


packages/core/src/registries/designTokenRegistry.ts
```diff
@@ -3,13 +3,47 @@ import { builtInStyles, optionalBuiltInStyles } from '../definitions/styles';
...
+          style: 'solid',
+        };
+      }
+      if (!borderValue.color) {
+        borderValue = {
+          ...borderValue,
+          color: '#000000',
+        };
+      }
+      designTokenDefinition.border[borderKey] = borderValue;
```
primeinteger, line 34: If you want to 🏌️‍♂️ this could be condensed a little bit ⛳ 

```suggestion
      designTokenDefinition.border[borderKey] = {
        ...borderValue,
        width: borderValue.width || '1px',
        style: borderValue.style || 'solid',
        color: borderValue.color || '#000000',
      };
```

packages/core/src/registries/designTokenRegistry.ts
```diff
@@ -3,13 +3,47 @@ import { builtInStyles, optionalBuiltInStyles } from '../definitions/styles';
 
 export let designTokensRegistry: DesignTokensDefinition = {};
 
+// This function is used to ensure that the composite values are valid since now those composite values are optional.
+// Therefore only border and in the future text related design tokens are/will be checked in this funciton.
+// Ensuring values for simplie key-value design tokens are not neccessary since they are required via typescript.
```
primeinteger, line 8: "simple" instead of "simplie"?  I'm having a hard time reading/understanding this comment 🤔

packages/core/src/registries/designTokenRegistry.ts
elylucas: I think I agree with @primeinteger, though I have no idea on the meaning of those emojies :)

elylucas (approved):
Overall looks good to me. I agreed with Adrian on making that one bit of code more concise (if it works), but I'll approve and let you decide on what to do with it.

packages/core/src/registries/designTokenRegistry.ts
primeinteger: https://en.wikipedia.org/wiki/Code_golf

Achieving the same result with less characters - although less characters is not always better, of course. Ideally you find the right balance between simplicity and readability since "code golf" can be taken too far if you're basically writing minified code with single-letter variables and other things like that make the code harder to read

ryunsong-contentful: Yeah but in this case, what you suggested is way better

PR #643, opened 2024-06-03
fix: width calc fix [ALT-797]

sam-is-content:
## Purpose

Before:
https://www.loom.com/share/cc5603ca631c4f2fb82acaa9c556ff85?sid=e7ebc44d-748f-4e14-8556-af37a1b483ca

After:
https://www.loom.com/share/185fb628d70648a79faf337d5d246cb8?sid=14bf5415-30a8-4875-9810-e65092750d0c

Current Bug:
For structure components, with a fixed width using a unit of measure (px, %, em, rem), if you set a margin that includes a unit of measure on both the left and right side, the width of the section subtracts the margins.  (see video if this write-up doesnt make sense)

We should fix this so it behaves as other components do for width.


primeinteger (commented):
I found where the `calc()` was originally added:
* https://contentful.atlassian.net/browse/SPA-1523
* https://github.com/contentful/experience-builder/pull/140

After reading the story description above, it sounded like the update was to make margin act like padding when a container has 100% width.

primeinteger (approved):
I'm a fan of this change. The calc() logic seemed odd to me.

SofiaMargariti:
@ChidinmaOrajiaku do you remember why we added this calculation here? Maybe it's not necessary anymore but I remember we added it for some reason, so double checking

PR #644, opened 2024-06-05
test: extend core package unit tests [SPA-2030]

SofiaMargariti:
## Purpose

Extend unit test coverage in core package.


packages/core/src/utils/pathSchema.spec.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  isDeepPath,
+  lastPathNamedSegmentEq,
+  parseDataSourcePathIntoFieldset,
+  parseDataSourcePathWithL1DeepBindings,
```
Spring3, line 5: I wish we could use this function from our web app
https://github.com/contentful/user_interface/blob/master/src/javascripts/features/experience-builder/core/schema/utils.ts#L38-L55

It tackles both - deep and shallow bindings

Maybe as a future improvement

Spring3 (approved):
(Great job! Qululululluulu)
  v
🦃

elylucas (approved):
Good stuff, thanks Sofia!

packages/core/src/utils/pathSchema.spec.ts
SofiaMargariti: Good suggestion! Maybe we could move this from the web app here and then export it and use it in both places.

PR #645, opened 2024-06-05
Add NextJS example apps

elylucas:
## Purpose

Adds a couple of NextJS apps that can be used for reference in setting up Experiences with both pages and app router.


examples/next-app-router/.env.local.tmpl
primeinteger, line 1: Should the `NEXT_PUBLIC_CTFL_DOMAIN` env var be included here as well, or did you intentionally leave that one out?

examples/next-app-router/src/app/globals.css
```diff
@@ -0,0 +1,20 @@
+@tailwind base;
```
primeinteger, line 1: I noticed the visual editor's component hover labels are shorter when running this app compared to the Test App.

It looks like the tailwind base styling is doing that. When I commented out this line the label matches the size in the Test App.  I'm not sure it's a big deal but wanted to mention it.

Screenshot 2024-06-05 at 12 15 44 PM

examples/next-app-router/src/app/page.tsx
primeinteger, line 1: Should this file be deleted?  Doesn't look like it's being used anywhere from what I can tell

examples/next-pages-router/public/next.svg
primeinteger, line 1: Delete? Doesn't look like it's being used

examples/next-pages-router/public/vercel.svg
primeinteger, line 1: Delete?

examples/next-pages-router/tailwind.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    './src/app/**/*.{js,ts,jsx,tsx,mdx}',
+  ],
+  theme: {
+    extend: {
+      backgroundImage: {
+        'gradient-radial': 'radial-gradient(var(--tw-gradient-stops))',
+        'gradient-conic': 'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
+      },
+    },
+  },
```
primeinteger, line 16: This config might be safe to remove from both example apps, since we're not using the built-in welcome page

examples/next-app-router/src/app/[locale]/[slug]/page.tsx
Comment (primeinteger) [inline examples/next-app-router/src/app/[locale]/[slug]/page.tsx:1]: When running the `next-pages-router` there's lots of `Warning: useLayoutEffect does nothing on the server` when loading the page in editor and delivery mode

There's also an error trying to find the favicon when viewing experiences in delivery mode or accessing the app URLs directly.  I'm not sure that these are a big deal just wanted to mention it

Screenshot 2024-06-05 at 12 31 47 PM

primeinteger (approved):
These new example apps are working great!  I posted a few comments below but those things are all small potatoes and you can decide to take action or ignore them. Great work!

Spring3:
@anwaar931 FYI I tagged you here as well to check Ely's implementation.

anwaar931:
> @anwaar931 FYI I tagged you here as well to check Ely's implementation.

Hey @Spring3, I looked at the implementation. I was able to fix react hydration, inside `detachExperienceStyles`. We need to update these examples once detachExperienceStyles is migrated to core packages.

examples/next-app-router/src/app/[locale]/[slug]/page.tsx
Comment (elylucas) [reply examples/next-app-router/src/app/[locale]/[slug]/page.tsx:1]: I'll file a new ticket with a way to handle this

PR #647, opened 2024-06-06
[ALT-911] fix: only have default background color for components that need to have them

ryunsong-contentful:
## Purpose
Only have default background color for components that need to have them which at this point is the `Button` and `Divider` component

https://www.loom.com/share/d44ced7d7abf4350898ec459c4448e86?sid=c7105e44-928c-4616-99f3-68f27579c313

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -75,7 +75,6 @@ export const builtInStyles: VariableDefinitions = {
     type: 'Text',
     group: 'style',
     description: 'The background color of the section',
-    defaultValue: 'rgba(255, 255, 255, 0)',
   },
   cfWidth: {
     displayName: 'Width',
@@ -427,7 +426,6 @@ export const singleColumnBuiltInStyles: VariableDefinitions = {
     type: 'Text',
     group: 'style',
     description: 'The background color of the column',
-    defaultValue: 'rgba(255, 255, 255, 0)',
   },
   cfFlexDirection: {
     displayName: 'Direction',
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Heading/index.ts
```diff
@@ -59,6 +59,12 @@ export const HeadingComponentDefinition: ComponentDefinition = {
       description: 'The text bold of the heading.',
       defaultValue: true,
     },
+    cfBackgroundColor: {
+      displayName: 'Background color',
+      type: 'Text',
+      group: 'style',
+      description: 'The background color of the heading',
+    },
```
primeinteger, line 67: I don't think this addition is needed.  This variable is inherited from `builtInStyles` const exported from packages/core/src/definitions/styles.ts, so we would only need to define this variable here if we wanted to set a different `defaultValue`.

packages/components/src/components/RichText/index.ts
```diff
@@ -40,6 +40,12 @@ export const RichTextComponentDefinition: ComponentDefinition = {
       description: 'The text alignment of the heading.',
       defaultValue: 'center',
     },
+    cfBackgroundColor: {
+      displayName: 'Background color',
+      type: 'Text',
+      group: 'style',
+      description: 'The background color of the rich text',
+    },
```
primeinteger, line 48: I don't think this addition is needed. Same reason

packages/components/src/components/Text/index.ts
```diff
@@ -40,6 +40,12 @@ export const TextComponentDefinition: ComponentDefinition = {
       description: 'The height of the button.',
       defaultValue: 'fit-content',
     },
+    cfBackgroundColor: {
+      displayName: 'Background color',
+      type: 'Text',
+      group: 'style',
+      description: 'The background color of the text',
+    },
```
primeinteger, line 48: I don't think this addition is needed. Same reason

primeinteger (approved):
I would remove the updates to Heading, RichText and Text, but otherwise looks good 🚀

PR #648, opened 2024-06-07
chore: update breakpoint display icon property []

ChidinmaOrajiaku:
## Purpose
Update breakpoint display icon property

### DIFF
--- a/packages/validators/src/schemas/v2023_09_28/experience.ts
+++ b/packages/validators/src/schemas/v2023_09_28/experience.ts
@@ -84,7 +84,7 @@ export const BreakpointSchema = z
     query: z.string().regex(/^\*$|^<[0-9*]+px$/),
     previewSize: z.string(),
     displayName: z.string(),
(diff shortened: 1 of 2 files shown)


packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -84,7 +84,7 @@ export const BreakpointSchema = z
     query: z.string().regex(/^\*$|^<[0-9*]+px$/),
     previewSize: z.string(),
     displayName: z.string(),
-    displayIconUrl: z.string().optional(),
+    displayIcon: z.string().optional(),
```
SofiaMargariti, line 87: I think you can define this as an enum. This will reflect in the types and the validator and will give an indication to the developer about what are the valid options.
```suggestion
    displayIconUrl: z.enum(['desktop', 'tablet', 'mobile']).optional(),
```

SofiaMargariti (approved):

Chaoste (approved):

PR #653, opened 2024-06-10
feat: add text design token

lilbitner:

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -230,7 +230,9 @@ export type RecursiveDesignTokenDefinition = {
   [key: string]: string | RecursiveDesignTokenDefinition;
 };
 
+/** not yet implemented, don't use */
 type DesignBorderTokenStyle = 'solid' | 'dashed' | 'dotted';
+/** not yet implemented, don't use */
 type DesignTextTokenEmphasis =
   | 'bold'
   | 'italic'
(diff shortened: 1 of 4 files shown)


ryunsong-contentful (approved):
Looks good overall. Few things

1. `Custom` is probably a more apt name over None has technically, there are custom style being applied, just not any of the DT ones.
2. I think next Monday you should alert the team in the morning if there is any objections in merging in Lillians EB PR.
3. I'm sure you're still tracking this but when I change a value and revert it back (aka sm -> md -> sm), the Style should also recognize that it's now once again a predefined DT style.

packages/core/src/types.ts
```diff
@@ -230,19 +230,44 @@ export type RecursiveDesignTokenDefinition = {
   [key: string]: string | RecursiveDesignTokenDefinition;
 };
 
+/** not yet implemented, don't use */
```
ryunsong-contentful, line 233: I think the `DesignBorderTokenStyle` is good to go as nothing changed, it's just reassigning the values to a type

packages/core/src/types.ts
```diff
@@ -230,19 +230,44 @@ export type RecursiveDesignTokenDefinition = {
...
+type DesignBorderTokenStyle = 'solid' | 'dashed' | 'dotted';
+/** not yet implemented, don't use */
```
ryunsong-contentful, line 235: Maybe we can provide something more formal since this is customer facing - normally I wouldn't comment if it was internal code but since devs outside Contentful will see this, would be worth putting something that would formally present the message like

```
/** We are currently in the process of implementing this feature. Do not use this field in your designTokenDefinition. **/
```

ryunsong-contentful (approved):
Minor comments - I am assuming you will adjust them as needed

PR #654, opened 2024-06-11
chore: move ssr styles from sdk to core [SPA-2109]

anwaar931:
## Purpose
Fixed SSR styles detachment for patterns. Major fixes

- Fetching the pattern entry from `entityStore?.experienceEntryFields?.usedComponents`
- Adding `cfSsrClassName` to pattern node

https://contentful.atlassian.net/browse/SPA-2092


packages/core/src/utils/styleUtils/ssrStylesPattern.spec.ts
```diff
@@ -766,15 +765,7 @@ describe('pattern component', () => {
...
-      usedComponents: [
-        {
-          sys: {
-            type: 'Link',
-            linkType: 'Entry',
-            id: patternNode.definitionId,
-          },
-        },
-      ],
+      usedComponents: [patternEntry],
```
Spring3, line 768: shouldn't it be a `Link`?

packages/experience-builder-sdk/src/index.ts
```diff
@@ -17,8 +17,6 @@ export {
   CF_STYLE_ATTRIBUTES,
 } from '@contentful/experiences-core/constants';
 
-export { detachExperienceStyles } from './utils/ssrStyles';
```
Spring3, line 20: to keep the public API untouched, we should keep the export here as the following:
```
export { detachExperienceStyles } from '@contentful/experiences-core'
```

Otherwise, all our code examples will be affected and all our existing customers, who imported this fn from our react-sdk package would get an error saying that there is no such export

packages/core/src/utils/styleUtils/ssrStylesPattern.spec.ts
anwaar931: It can be an array `Link` or a pattern entry https://github.com/contentful/experience-builder/blob/development/packages/core/src/types.ts#L225

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -102,9 +101,9 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
 
     let currentNode: ComponentTreeNode | undefined = undefined;
 
-    const registeredComponenIds = Array.from(componentRegistry.values()).map(
-      ({ definition }) => definition.id,
-    );
+    // const registeredComponenIds = Array.from(componentRegistry.values()).map(
+    //   ({ definition }) => definition.id,
+    // );
```
Spring3, line 106: I guess we. can remove these comments if they are not needed anymore (there's one more below)

Spring3 (approved):

PR #658, opened 2024-06-13
feat: fallback missing components [SPA-2033]

dimitrycf:
## Purpose

> Relevant [user_interface PR](https://github.com/contentful/user_interface/pull/21393)

Instead of showing fatal error upon encountering missing component, just render placholder. Which would still allow the Editor to see the component in the canvas in EDITOR mode, as well as act upon canvas to solve the situation.

On the screenshot you can see that {missing components} are rendered as placeholders. And canvas stays fully interactive, so that the user can address the issue. 

image

## Approach 1 

> Retired in favor of Approach 2

~The approach taken here is to introduce new component `<SelectableComponent>` which is supposed to be a younger and poorer brother of `<DraggableComponent>` with all of the DnD functionality stripped out.~

~The assumption behind this choice was that - enhancing `<DraggableComponent>` with additional functionality will make it even more complicated. It already has so many variables/props which it depends on, that it's not that easy to follow what is happening there.~

## Approach 2 

Whilst initially it seemed having simplistic `<SelectableComponent>` would make things easier, it turned out that it broke drag-n-drop completely in the canvas. And fixing that would force `<SelectableComponent>` to grow in code and complexity which would beat the purpose of it's existence in the first place. 

With Approach 2, we simply modify `<EditorBlock>` component to be compatible with {missing component} situation. 

Also there's one simplification, where instead of passing entire `componentDefinition` we only pass `componentDefinitio.displayName` and thus decouple from the entire `componentDefinition` and make code a bit simpler.

### DIFF
--- a/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
+++ b/packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
@@ -43,7 +43,7 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
   const { slotId } = parseZoneId(zoneId);
   const setSelectedNodeId = useEditorStore((state) => state.setSelectedNodeId);
   const selectedNodeId = useEditorStore((state) => state.selectedNodeId);
-  const { node, wrapperProps, definition, elementToRender } = useComponent({
+  const { node, wrapperProps, definition, elementToRender, isComponentMissing } = useComponent({
     node: rawNode,
     resolveDesignValue,
     renderDropzone,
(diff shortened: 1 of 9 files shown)


packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
```diff
@@ -94,6 +94,42 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
...
+    const renderMissingComponentPlacehoder = () => (
+      <div
+        style={{
+          border: '1px solid red',
+          width: '100%',
+          height: '100%',
+        }}>
+        Missing component &apos;{node.data.blockId}&apos;
+      </div>
+    );
```
primeinteger, line 107: Since this same component is defined in two files, this could potentially be moved to its own file and imported into both places it's used.

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
```diff
@@ -94,6 +94,42 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
...
+        blockId={node.data.blockId}
+        coordinates={coordinates!}
+        wrapperProps={wrapperProps}
+        onClick={onClick}
+        onMouseOver={onMouseOver}
+        displayName={displayName || definition?.name}>
+        {renderMissingComponentPlacehoder()}
+        {/* {isStructureComponent && userIsDragging && (
+          <Hitboxes parentZoneId={zoneId} zoneId={componentId} isEmptyZone={isEmptyZone} />
+        )} */}
```
primeinteger, line 128: Might be better to delete this commented out code

packages/visual-editor/src/components/Dropzone/useComponent.ts
```diff
@@ -69,31 +71,48 @@ export const useComponent = ({
...
-    : node.type === ASSEMBLY_NODE_TYPE
-      ? // Assembly.tsx requires renderDropzone and editorMode as well
-        () => React.createElement(componentRegistration.component, componentProps)
-      : () =>
-          React.createElement(
-            ImportedComponentErrorBoundary,
-            null,
-            React.createElement(componentRegistration.component, otherComponentProps),
-          );
+  const createElementToRender = (componentRegistration: ComponentRegistration) => {
```
primeinteger, line 81: I like this update, the code is easier to read now. 👏

primeinteger (approved):

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
dimitrycf: ✅  FIXED 65ad695c0ba70b01158c1492491aaa947127b931

packages/visual-editor/src/components/Dropzone/EditorBlock.tsx
dimitrycf: ✅  FIXED 65ad695c0ba70b01158c1492491aaa947127b931

packages/visual-editor/src/components/Dropzone/useComponent.ts
dimitrycf: Or was it? I think I spoke about L97.. but you spoke about L81... :)

dimitrycf: ✅  RESOLVEd

PR #664, opened 2024-06-21
fix: add TS types for messages exchanged between SDK and Web app [SPA-2122]

SofiaMargariti:
## Purpose

Add types for messages exchanged between SDK and Wep app.

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -2,13 +2,12 @@ export const SCROLL_STATES = {
   Start: 'scrollStart',
   IsScrolling: 'isScrolling',
   End: 'scrollEnd',
-};
+} as const;
 
 export const OUTGOING_EVENTS = {
   Connected: 'connected',
   DesignTokens: 'registerDesignTokens',
   RegisteredBreakpoints: 'registeredBreakpoints',
-  HoveredSection: 'hoveredSection',
   MouseMove: 'mouseMove',
   NewHoveredElement: 'newHoveredElement',
   ComponentSelected: 'componentSelected',
@@ -19,14 +18,14 @@ export const OUTGOING_EVENTS = {
   ComponentMoved: 'componentMoved',
   CanvasReload: 'canvasReload',
   UpdateSelectedComponentCoordinates: 'updateSelectedComponentCoordinates',
-  UpdateHoveredComponentCoordinates: 'updateHoveredComponentCoordinates',
   CanvasScroll: 'canvasScrolling',
   CanvasError: 'canvasError',
   ComponentMoveStarted: 'componentMoveStarted',
   ComponentMoveEnded: 'componentMoveEnded',
   OutsideCanvasClick: 'outsideCanvasClick',
   SDKFeatures: 'sdkFeatures',
-};
+  RequestEntities: 'REQUEST_ENTITIES',
+} as const;
 
 export const INCOMING_EVENTS = {
   RequestEditorMode: 'requestEditorMode',
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -41,14 +41,14 @@ export const INCOMING_EVENTS = {
   UpdatedEntity: 'updatedEntity',
   AssembliesAdded: 'assembliesAdded',
   AssembliesRegistered: 'assembliesRegistered',
-  InitEditor: 'initEditor',
   MouseMove: 'mouseMove',
-};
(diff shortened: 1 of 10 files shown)


packages/core/src/types.ts
```diff
@@ -380,3 +379,164 @@ export type BackgroundImageOptions = {
...
+  updateSelectedComponentCoordinates: UpdateSelectedComponentCoordinatesPayload;
+  canvasScrolling: CanvasScrollPayload;
+  componentMoveStarted: ComponentMoveStartedPayload;
+  componentMoveEnded: ComponentMoveEndedPayload;
+  outsideCanvasClick: OutsideCanvasClickPayload;
+  sdkFeatures: SDKFeaturesPayload;
+  REQUEST_ENTITIES: RequestEntitiesPayload;
+};
+
+export type SendMessageParams = <T extends OutgoingEvent>(
```
Chaoste, line 451:
```suggestion
export type SendMessageFunction = <T extends OutgoingEvent>(
```

packages/core/src/constants.ts
```diff
@@ -2,13 +2,12 @@ export const SCROLL_STATES = {
...
   IsScrolling: 'isScrolling',
   End: 'scrollEnd',
-};
+} as const;
 
 export const OUTGOING_EVENTS = {
   Connected: 'connected',
   DesignTokens: 'registerDesignTokens',
   RegisteredBreakpoints: 'registeredBreakpoints',
-  HoveredSection: 'hoveredSection',
```
Chaoste, line 11: I guess these deleted messages were left overs and not needed anymore?

packages/core/src/types.ts
```diff
@@ -380,3 +379,164 @@ export type BackgroundImageOptions = {
...
+  data: OUTGOING_EVENT_PAYLOADS[T],
+) => void;
+
+export type OutgoingMessage = {
+  [K in keyof OUTGOING_EVENT_PAYLOADS]: {
+    source: 'customer-app';
+    eventType: K;
+    payload: OUTGOING_EVENT_PAYLOADS[K];
+  };
+}[keyof OUTGOING_EVENT_PAYLOADS];
```
Chaoste, line 462: Sometimes I just feel like TypeScript was created to confuse people. These kinda hacks look like magic to me 😅

packages/core/src/types.ts
```diff
@@ -380,3 +379,164 @@ export type BackgroundImageOptions = {
...
+type SelectedValueTypes = Filter<ComponentPropertyValue['type'], 'UnboundValue' | 'BoundValue'>;
+
+type RequestEditorModePayload = undefined;
+type ExperienceUpdatedPayload = {
+  tree: ExperienceTree;
+  /** @deprecated in favor of assemblies */
+  designComponents?: ExperienceUsedComponents;
+  assemblies?: ExperienceUsedComponents;
+  locale: string;
+  /** @deprecated maybe? */
```
Chaoste, line 474: Is this something that we should clarify in a dedicated place with the team?

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -42,7 +42,7 @@ export const useDetectEditorMode = ({ isClientSide = false }: UseDetectEditorMod
       //Double check if we are in editor mode by listening to postMessage events
       if (typeof window !== 'undefined' && !window.__EB__?.isEditorMode) {
         window.addEventListener('message', onMessage);
-        sendMessage(OUTGOING_EVENTS.Connected);
+        sendMessage(OUTGOING_EVENTS.Connected, undefined);
```
Chaoste, line 45: Mmmh, not fully happy with this but maybe it's even good to make the payload explicit. I tried improving the types but failed so far. I guess it would only work with overloaded function types - an interesting challenge for the dev-typescript channel in Slack if we wanted this to be optional :)

Chaoste (approved):
Very nice work! I like the consistent namings, casings, etc. - all the things that usually sound non-important to other devs :D

packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
SofiaMargariti: Thanks for the suggestion, I also tried a few different things but couldn't make it work, and didn't want to block this PR longer. But I will followup on dev-typescript, because I also don't like this solution :D

PR #665, opened 2024-06-21
chore: add pact consumer test and publish to provider [SPA-2123]

Chaoste:
## Purpose
We want to leverage PactFlow for our new contract tests between UserInterface and ExperiencesSDK. This PR introduces a first consumer test and integrates the pact integration into our GitHub actions.

## Approach
- Add a unit test that creates a pact for an SDK message received from the UserInterface ("updatedEntity"). The test not only defines the payload shape but also tests the subsequent behavior of the message listener.
- Add a GH workflow step to push the pact to PactFlow.
- Persist generated pact as an artifact in the workflow.
- Add a GH workflow step that records the published pact as released when being on the main branch.
- Deactivate both steps via a repository variable `ENABLE_PACTFLOW` until the procurement for the tool is successful.
- Create a system account in PactFlow with permission to record a release.
- Add required secrets and variables to the repository settings.

As the new SDK version is only available after the version was bumped, we can't tag the uploaded pacts with that version beforehand. Thus, we only rely on the commit hash for the time being.

## Screenshots of the successful run-through
Link to green GH workflow: https://github.com/contentful/experience-builder/actions/runs/9646098357
1. Publish pact for commit hash
Screenshot 2024-06-24 at 15 46 02
2. Record release for the commit hash
Screenshot 2024-06-24 at 15 45 40
3. Screenshot from PactFlow UI showing the pushed pact and the recorded release environment "test".
Screenshot 2024-06-24 at 15 37 50

To enable the integration later, simply set this variable to "true":
Screenshot 2024-06-24 at 15 49 51

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,13 +46,11 @@ jobs:
   # Run after publishing according to docs:
   # https://docs.pact.io/pact_broker/recording_deployments_and_releases#recording-releases
   pact-record-release:
-    # needs: publish
-    needs: check
+    needs: publish
     uses: ./.github/workflows/pact-record-release.yaml
-    # github.ref_name == 'main' &&
     if: >-
       vars.ENABLE_PACTFLOW == 'true' &&
-      github.ref_name == 'feat/SPA-2123-publish-pacts' &&
+      github.ref_name == 'main' &&
       !contains(github.event.head_commit.message, 'skip-release') &&
       github.actor != 'dependabot[bot]' &&
       github.actor != 'contentful-automation[bot]'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,11 +46,12 @@ jobs:
   # Run after publishing according to docs:
   # https://docs.pact.io/pact_broker/recording_deployments_and_releases#recording-releases
   pact-record-release:
-    # needs: [publish]
+    # needs: publish
     needs: check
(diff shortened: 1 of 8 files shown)


.github/workflows/pact-record-release.yaml
```diff
@@ -0,0 +1,31 @@
+name: 'Run Checks: Lint & Tests'
```
SofiaMargariti, line 1: Should this name be changed to: `Record Pact Release` or something similar?

.github/workflows/main.yml
```diff
@@ -41,6 +43,20 @@ jobs:
...
+  # Run after publishing according to docs:
+  # https://docs.pact.io/pact_broker/recording_deployments_and_releases#recording-releases
+  pact-record-release:
+    needs: publish
+    uses: ./.github/workflows/pact-record-release.yaml
+    if: >-
+      vars.ENABLE_PACTFLOW == 'true' &&
+      github.ref_name == 'main' &&
+      !contains(github.event.head_commit.message, 'skip-release') &&
+      github.actor != 'dependabot[bot]' &&
```
SofiaMargariti, line 55: Probably not needed because we're only merging manually on main, but we can keep it as extra precaution.

Chaoste: Yes, I thought it would be fine to have more guardrails in place to be prepared if the repo changed over time :)

SofiaMargariti (approved):

PR #674, opened 2024-06-25
feat(create-contentful-studio-experiences): add more validation to CLI tool [ALT-991]

primeinteger:
## Purpose

- Checks to see if the Experience Type already exists before creating one. ([ALT-994](https://contentful.atlassian.net/browse/ALT-994))
- Checks to see if there is already an Experience Entry for the `homePage` slug before creating one. ([ALT-999](https://contentful.atlassian.net/browse/ALT-999))
- Fixes a bug when choosing not to connect to a contentful space. When this step is skipped the project is bootstrapped and the CLI tool informs the user to update their .env.local file. ([ALT-998](https://contentful.atlassian.net/browse/ALT-998))

## Screenshots

Bootstrapping a project only (now mentions to update the .env.local file at the end)

- Screenshot 2024-06-25 at 4 08 45 PM

When selecting an existing space, there are additional validation checks before creating a content type and entry

- Screenshot 2024-06-25 at 3 51 12 PM

[ALT-994]: https://contentful.atlassian.net/browse/ALT-994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ALT-999]: https://contentful.atlassian.net/browse/ALT-999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ALT-998]: https://contentful.atlassian.net/browse/ALT-998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ryunsong-contentful (approved):
Overall awesome improvement. One inconsequential bug I found is that I can move onto the next step with an invalid directory name and then hitting `ctrl+c`. SO rather than killing the entire process it let me go to the next step with the invalid name. Eventually it fails and nothing gets created but just something for you to note in my testing

Screenshot 2024-06-26 at 3 51 00 PM
Screenshot 2024-06-26 at 3 53 18 PM

primeinteger:
> One inconsequential bug I found is that I can move onto the next step with an invalid directory name and then hitting `ctrl+c`. SO rather than killing the entire process it let me go to the next step with the invalid name. Eventually it fails and nothing gets created but just something for you to note in my testing

@ryunsong-contentful ~ I noticed this to it's kind of annoying when you want to cancel the script. It looks like the library we're using provides a method for handling cancellation more gracefully so I'm going to make a small update to incorporate this in a follow-up PR: https://www.npmjs.com/package/@clack/prompts#cancellation

PR #685, opened 2024-07-02
feat: allow Link and Array type references as variables

elylucas:
## Purpose

This PR allows a dev to set up variables that can bind to lists (arrays) and entry references as such:

```ts
variables: {
  // bind to a string list
  listText: {
    displayName: 'List Text',
    type: 'Array',
  },
  // bind to an entry reference
  entryReference: {
    displayName: 'Entry Reference',
    type: 'Link',
  },
  // bind to a list of entries
  listReference: {
    displayName: 'List Entry Reference',
    type: 'Array',
  },
}
```


packages/core/src/utils/transformers/getArrayValue.ts
```diff
@@ -0,0 +1,34 @@
...
+export function getArrayValue(
+  entryOrAsset: Entry | Asset,
+  path: string,
+  entityStore: EntityStoreBase,
+) {
+  if (entryOrAsset.sys.type === 'Asset') {
+    return entryOrAsset;
+  }
+
+  const arrayValue = get(entryOrAsset, path.split('/').slice(2, -1));
```
primeinteger, line 15: It looks like the type for `arrayValue` is being inferred as `any` here.  It might be nice to add a type argument here so TS can help catch type errors if this code is updated in the future.

```suggestion
  const arrayValue = get<Array<string | UnresolvedLink<'Entry' | 'Asset'>>>(entryOrAsset, path.split('/').slice(2, -1));
```

packages/core/src/utils/transformers/getArrayValue.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!isArray(arrayValue)) {
+    console.warn(`Expected a value to an array, but got: ${JSON.stringify(arrayValue)}`);
```
primeinteger, line 18:
```suggestion
    console.warn(`Expected a value to be an array, but got: ${JSON.stringify(arrayValue)}`);
```

primeinteger (approved):
Should this PR be marked as a `feat` instead? Great work on this one! 🚀

PR #687, opened 2024-07-09
test: pact experiences sdk provider verification [SPA-2128]

SofiaMargariti:
## Purpose

Implement pact verification for messages emitted by the SDK (ExperiencesSDKProvider) and consumed by the Web App (UserInterfaceConsumer).

## Approach

* Define a message provider function for each message/pact
* In the message provider:
    * Spy on `window.postMessage` function in order to intercept messages being sent
    * Implement a simple unit test that will trigger the message we're interested in. Depending on the type of message here we might need to call a function/render a hook/render the whole `VisualEditorRoot` component.
* Some trivial events with no payload have been excluded, e.g. `ComponentMoveStarted`, `ComponentMoveEnded`


.github/workflows/publish.yaml
```diff
@@ -55,8 +57,8 @@ jobs:
             packages/test-apps/react-vite/dist
             packages/test-apps/nextjs/.next
           key: build-cache-${{ github.run_id }}-${{ github.run_attempt }}
-      - name: Authenticate with private GH package
-        run: echo -e "//npm.pkg.github.com/:_authToken=${{ steps.vault.outputs.GITHUB_PACKAGES_WRITE_TOKEN }}\n@contentful:registry=https://npm.pkg.github.com" > ./.npmrc
+      - name: Authenticate with private NPM and GH package
+        run: echo -e "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}\n//npm.pkg.github.com/:_authToken=${{ steps.vault.outputs.GITHUB_PACKAGES_WRITE_TOKEN }}\n@contentful:registry=https://npm.pkg.github.com" > ./.npmrc
```
Chaoste, line 61: I'm wondering a bit why this worked before without explicitly adding the token. Were you trying out something or is there a deeper reasoning to that?

Chaoste (approved):
Looks good!

.github/workflows/publish.yaml
SofiaMargariti: This might be related to rebase, I haven't touched this file. I am gonna sync this and the feature branch with development and hopefully this diff will go away

PR #688, opened 2024-07-10
test: experiences SDK consumer pacts [SPA-2129]

SofiaMargariti:
## Purpose

Implement consumer contract tests for ExperiencesSDKConsumer.

## Approach

* Each interaction includes an example payload and a payload matcher as well as a unique id. 
* Each interaction is run through the `createPostMessageReceiver` function which renders the `useEditorSubscriber` hook and triggers a message with the specified payload.
* Each interaction created a pact in the `pacts/ExperiencesSDKConsumer-UserInterfaceProvider.json` file. (This file is ignored by git, but running the tests locally will generate it and can be inspected)
* For each interaction we run a set of assertions based on what is expected to happen when the specific message is received.
* Some trivial events with no payload have been ignored.

### DIFF
--- a/packages/visual-editor/src/hooks/useEditorSubscriber.spec.ts
+++ b/packages/visual-editor/src/hooks/useEditorSubscriber.spec.ts
@@ -154,13 +154,15 @@ describe('Canvas Subscriber methods', () => {
   const createPostMessageReceiverForSelectedEntities = (_event: IncomingEvent, payload) =>
     asynchronousBodyHandler(async () => {
       const store = new EditorModeEntityStore({ entities: [], locale: 'en-US' });
-
-      const promise = store.fetchEntities({
-        missingEntryIds: [entityIds.ENTRY1, entityIds.ENTRY2],
-        missingAssetIds: [],
-        skipCache: true,
+      let listener;
+      vi.spyOn(window, 'addEventListener').mockImplementationOnce((_event, _listener) => {
+        listener = _listener;
       });
-      window.dispatchEvent(new MessageEvent('message', { data: JSON.stringify(payload) }));
+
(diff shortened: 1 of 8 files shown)


packages/visual-editor/src/hooks/useEditorSubscriber.spec.ts
```diff
@@ -85,19 +146,42 @@ describe('Canvas Subscriber methods', () => {
...
-    await pact
-      .given(`a ${event} message is sent`)
-      .expectsToReceive(`a ${event} message`)
-      .withContent(data)
-      .verify(createPostMessageReceiver(event, payload));
-  });
+  const createPostMessageReceiverForSelectedEntities = (_event: IncomingEvent, payload) =>
+    asynchronousBodyHandler(async () => {
+      const store = new EditorModeEntityStore({ entities: [], locale: 'en-US' });
+      console.log(store);
```
Chaoste, line 157:
```suggestion
```

packages/visual-editor/src/hooks/useEditorSubscriber.spec.ts
```diff
@@ -85,19 +146,42 @@ describe('Canvas Subscriber methods', () => {
...
+      expect(store.entities).toEqual(expect.arrayContaining(entries));
+    });
+
+  it.each(interactions)(
+    'should receive the expected payload for $event event',
+    async ({ id, description, event, payload, payloadMatcher }) => {
+      let messageReceiverConstructor = createPostMessageReceiver;
+      if (event !== INCOMING_EVENTS.RequestedEntities) {
+        return;
+      }
```
Chaoste, line 177: Looks like we're currently testing only one out of all those interactions. Is that intended?

PR #690, opened 2024-07-11
fix: useDetectEditorMode throws an error in the nextjs test app []

primeinteger:
## Purpose

Prevent this error from being thrown when running the nextjs test app:

Screenshot 2024-07-11 at 7 56 01 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33363,9 +33363,9 @@
       }
     },
     "node_modules/postcss": {
-      "version": "8.4.38",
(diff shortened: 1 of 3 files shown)


chasepoirier (commented):
Looks like you got some package-lock.json changes in there. Are they needed?

primeinteger:
> Looks like you got some package-lock.json changes in there. Are they needed?

Yeah `npm install` generated those updates. I could remove those changes but it looks like they will just come back the next time someone runs `npm install` so maybe we should keep 'em. Thoughts?

elylucas (approved):

PR #693, opened 2024-07-11
chore: update nextjs apps to use defineComponents

elylucas:
## Purpose

I noticed the new Next.js apps didn't have a call to defineComponents since they didn't have any custom components in the example. When I went to add it, I saw that in the case of app router, the current set up where the ExperienceRoot was directly in a RSC, the call to defineComponents didn't work as that was made server side. I updated the app router apps to put the `ExperienceRoot` in a client component and called the defineComponents method from there.

### DIFF
--- a/examples/next-app-router/src/eb-config.ts
+++ b/examples/next-app-router/src/eb-config.ts
@@ -1,0 +1,21 @@
+import { defineComponents } from '@contentful/experiences-sdk-react';
+
+defineComponents([
+  // Add your custom components here
+  // example:
+  // {
+  //   component: Button,
+  //   definition: {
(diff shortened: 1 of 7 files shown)


examples/next-app-router/src/eb-config.ts
primeinteger, line 1: Do you think we should rename the `eb-config` files to something without "eb"?

Maybe `studio-config` or `exp-config` 🤔

primeinteger (approved):
Looks great!

examples/next-app-router/src/eb-config.ts
elylucas: ya thats a good idea

PR #695, opened 2024-07-16
feat: support rendering of component values for design properties [SPA-2145]

Chaoste:
## Purpose

With the new feature of exposing design properties for pattern instances, they can now be of type `ComponentValues` inside the tree of a pattern.

## Approach

Similar to the previously implemented content variables, we're replacing those with the value of the componentValue when rendering the respective assembly child node.


loweisz (approved):
lgtm 🚀

PR #698, opened 2024-07-17
fix: fix issue where dragging a structure component could periodically make it go bye bye

elylucas:
## Purpose

Fixes [ALT-1038]

## Approach

Dragging a structural component and dropping it in its own position could have an issue where the SDK thinks it switching the parent of the component to itself, which would remove it from the tree. I put a check in that if the droppable id is the same as the draggable id, to nope out of the operation.

This uncovered what could be a potential race condition in the `isDropZoneEnabled` [check](https://github.com/contentful/experience-builder/blob/d48b81a4de4bbaf7deddf5928646db95335e1968/packages/visual-editor/src/components/DraggableBlock/Dropzone.tsx#L95), in which the hovered element could briefly be set back to the component being dragged, which would result in the drop zone of the dragging component not be appropriately disabled.

[ALT-1038]: https://contentful.atlassian.net/browse/ALT-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/visual-editor/src/components/RootRenderer/DNDProvider.tsx
+++ b/packages/visual-editor/src/components/RootRenderer/DNDProvider.tsx
@@ -69,6 +69,13 @@ export const DNDProvider = ({ children }: Props) => {
     updateItem();
     SimulateDnD.reset();
 
+    // If the component is being dropped onto itself, do nothing
+    // This can happen from an apparent race condition where the hovering zone gets set
+    // to the component after its dropped.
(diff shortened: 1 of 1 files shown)


primeinteger (approved):
Nice clean fix!

PR #704, opened 2024-07-22
fix: validate defaultValue type matches type value of variable

elylucas:
## Purpose

[ALT-1025]

## Approach

Instead of doing any conversion of values that get passed in as wrong types, it was determined it would be better if we validate the type of defaultValue. To do so, I extended the zod validators to do some custom validation on `defaultValue` based on the type of the variable ("Text" should be a `string`, "Boolean" should be a`boolean`, "Number" should be a `number`, etc...)

I also refactored the test files around to get them closer to the source code and out of a central "test" folder. This seems to fit better with how we organize tests elsewhere in the monorepo and makes them more discoverable.

[ALT-1025]: https://contentful.atlassian.net/browse/ALT-1025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/validators/src/schemas/componentDefinition.ts
```diff
@@ -22,14 +22,120 @@ export const ComponentDefinitionSchema = z.object({
...
+    z
+      .object({
+        // TODO - extend with definition of validations and defaultValue
+        displayName: z.string().optional(),
+        type: DefinitionPropertyTypeSchema,
+        description: z.string().optional(),
+        group: z.string().optional(),
+        defaultValue: z.union([z.string(), z.number(), z.undefined(), z.any()]).optional(),
+      })
+      .superRefine((val, ctx) => {
```
elylucas, line 34: the majority of the relevant changes to validation are here in this superRefine method

packages/validators/src/validators/index.ts
elylucas, line 1: I refactored some of the validate functions into their own files to break them up a bit

packages/validators/src/schemas/componentDefinition.ts
```diff
@@ -22,14 +22,120 @@ export const ComponentDefinitionSchema = z.object({
...
-      group: z.string().optional(),
-    }),
+    z
+      .object({
+        // TODO - extend with definition of validations and defaultValue
+        displayName: z.string().optional(),
+        type: DefinitionPropertyTypeSchema,
+        description: z.string().optional(),
+        group: z.string().optional(),
+        defaultValue: z.union([z.string(), z.number(), z.undefined(), z.any()]).optional(),
```
primeinteger, line 32: The `PrimitiveValueSchema` type in `src/schemas/v2023_09_28/experience.ts` might be a good option here.
```suggestion
        defaultValue: PrimitiveValueSchema.optional(),
```

primeinteger (approved):

packages/validators/src/schemas/componentDefinition.ts
elylucas: ohh, thanks. I put that in to get it to work and mean to come back to it.

packages/validators/src/schemas/componentDefinition.ts
Chaoste: @elylucas did you consider using `z.or` instead of custom validators? I'm wondering whether a static validation would be better to handle, e.g. when later creating OpenAPI specs out of this (see the ongoing OpenAPI project that already supports zod and other validation libraries).

packages/validators/src/schemas/componentDefinition.ts
```diff
@@ -22,14 +30,120 @@ export const ComponentDefinitionSchema = z.object({
   id: DefinitionPropertyKeySchema,
   variables: z.record(
```
Chaoste, line 31: I realised last week that `ComponentDefinitionSchema` and `ComponentSettingsSchema` have the same variable definitions but we validate them separately. As a result, they are not correctly updated and this one is missing the correct `validations` and `defaultValue` properties. With your PR, we should update the comment in line 35 as you're adding `defaultValue`.
However, it would be great if we could somehow merge parts of this with `ComponentSettingsSchema` which defines all the same properties - the only difference is that componentSettings allows a bound/ unbound content pointer instead of plain strings for content variables.

packages/validators/src/schemas/componentDefinition.ts
elylucas: I had not, do you have an example of what that might look like in this use case? This is my first foray into zod.

packages/validators/src/schemas/componentDefinition.ts
elylucas: Are these the same variables? Seems `ComponentDefinitionSchema` is how its defined in the users code when they call `defineComponents`, and `ComponentSettingsSchema` is how these values are stored in the entry schema. So, you wouldn't want users to be able to specify the bound/unbound pointers in their `defineComponents` call.

Chaoste: You're right that we don't want bound/ unbound pointers in the registration call. Though, besides that they are exactly the same as we save the registered definition as is in the component Settings and just overwrite the `defaultValue`. For that reason, it would be somehow good to also combine both schema as currently they are not correctly defined for that reason that we have two places trying to define the same (validations missing)

packages/validators/src/schemas/componentDefinition.ts
Chaoste: Mmmmh, I'm also a newbie to zod. My assumption was that this could work the same as with union & intersection types in TS.

I'm just [reading](https://github.com/colinhacks/zod?tab=readme-ov-file#or) that `z.or` is basically a convenience method for union types ^^ (same goes for `and` and `intersection`)

So far, I imagine this should be defined like follows without a custom refinement:

```
z.object({
  // Shared properties
  displayName: z.string().optional(),
  description: z.string().optional(),
  group: z.string().optional(),
  // TODO: add validations as defined in ComponentSettingsSchema
}).and(z.union([
  z.object({
    type: z.literal('Array'),
    defaultValue: z.any().array().optional(),
  }),
  z.object({
    type: z.literal('Boolean'),
    defaultValue: z.boolean().optional(),
  }),
  // ... same for the remaining cases
])
```

Chaoste: I think you're PR is already a great improvement, so feel free to skip this. Maybe, we come back to this static validation approach when the need for it arises. So far, the only direct advantage would be more readable types created ouf of zod - something that would be okay to be addressed later as well

packages/validators/src/schemas/componentDefinition.ts
Chaoste: Thanks, looks very promising :)

Chaoste (approved):

packages/validators/src/schemas/componentDefinition.ts
elylucas: I gave this a try, and while it worked, I was having a hard time getting the error messages to be useful. For some reason passing in `message` wasn't being used. 

I agree on this approach is better for typing and code readability, but going to go with this solution for now to get it done.

PR #705, opened 2024-07-23
fix: incorrect placeholder styling [ALT-1011]

chasepoirier:
## Purpose
When we refactored the SDK to no longer include a wrapping div, placeholder styles needed to be updated to reference sizing from the correct element. Instead of referencing the `.parentElement` it needed to reference the element itself since the wrapping div no longer applies.

Ticket: https://contentful.atlassian.net/browse/ALT-1011


primeinteger (approved):
🎉

PR #706, opened 2024-07-23
fix: slot components should fit their content once they have children

elylucas:
## Purpose

[ALT-1012]

## Approach

Removed the filter based on `slotId` when determining if the node was empty or not. This had a ripple effect of removing passing around slotId up through `useComponentProps`, and `useComponent` as the `slotId` was only used in this calculation.

Also checking with @chasepoirier and @primeinteger to see if this filter had a specific purpose and if removing it would have any other effects.

### DIFF
--- a/packages/visual-editor/src/components/DraggableBlock/EditorBlock.tsx
+++ b/packages/visual-editor/src/components/DraggableBlock/EditorBlock.tsx
@@ -66,7 +66,6 @@ export const EditorBlock: React.FC<EditorBlockProps> = ({
     resolveDesignValue,
     renderDropzone,
     userIsDragging,
-    slotId,
   });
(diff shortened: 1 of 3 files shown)


primeinteger (approved):
Nice fix! 👏

primeinteger (approved):
I tested this change out locally and it's all working well!

At one point during nested slots dev I had implemented slot dropzones in the background/"invisible layers" and they didn't appear in the component tree. I think this filter in `useComponentProps` was likely part of that point of the implementation.

The dropzones were replaced with slot containers as normal components that are shown in the layers tab and can be selected on the canvas. This hook doesn't need to filter the children for nested slots, so you definitely made the right fix 😁 

Screenshot 2024-07-23 at 5 49 51 PM

chasepoirier:
Awesome, LGTM 🚀

PR #707, opened 2024-07-30
fix: ssr styles extraction for editable patterns [SPA-2157]

anwaar931:
## Purpose

With Editable patterns we now have different styles for each pattern instance, which needs to be considered when detaching styles for SSR.

## Approach

Pattern instances can have different styles due to Editable patterns, we have to store styles depending on the pattern instance in an experience rather than the pattern node itself.
We generate a unique identifier for each child component of a pattern and associate this identifier with the corresponding node. The style classes are then stored in the pattern wrapper node, keyed by these unique identifiers.
This approach ensures that the styles for each pattern’s instance are maintained separately, preventing conflicts.

### Things to consider
- Types needs to be adjusted as id doesn't exist currently in `ComponentTreeNode`

### DIFF
--- a/packages/core/src/utils/styleUtils/ssrStyles.ts
+++ b/packages/core/src/utils/styleUtils/ssrStyles.ts
@@ -169,7 +169,7 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
           // and this is where the over-writes for the default values are stored
           // yes, I know, it's a bit confusing
           componentVariablesOverwrites: currentNode.variables,
-          // and this is where the pattern node
+          // pass top-level pattern node to store instance-specific child styles for rendering
           patternWrapper: currentNode,
         });
         continue;
@@ -323,13 +323,13 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
       // making sure that we respect the order of breakpoints from
       // we can achieve "desktop first" or "mobile first" approach to style over-writes
       if (patternWrapper) {
-        // @ts-expect-error TODO: fix the types - add id to ComponentTreeNode
+        // @ts-expect-error -- temporarily add node ID to pick up the className during rendering in CompositionBlock
         currentNode.id = currentNode.id ?? generateRandomId(15);
-        // @ts-expect-error TODO: fix the types
+        // @ts-expect-error -- valueByBreakpoint is not explicitly defined, but it's already defined in the patternWrapper styles
         patternWrapper.variables.cfSsrClassName = {
           ...(patternWrapper.variables.cfSsrClassName ?? {}),
           type: 'DesignValue',
-          // @ts-expect-error TODO: fix the types - add id to ComponentTreeNode
+          // @ts-expect-error -- id is not defined in ComponentTreeNode type
           [currentNode.id]: {
             valuesByBreakpoint: {
               [breakpointIds[0]]: currentNodeClassNames.join(' '),
--- a/packages/core/src/utils/styleUtils/ssrStyles.ts
+++ b/packages/core/src/utils/styleUtils/ssrStyles.ts
@@ -575,8 +575,6 @@ export const indexByBreakpoint = ({
     let resolvedVariableData = variableData;
 
     if (variableData.type === 'ComponentValue') {
-      // console.log('componentVariablesOverwrites: ', componentVariablesOverwrites);
-      // console.log('componentSettings: ', componentSettings);
       const variableDefinition = componentSettings?.variableDefinitions[variableData.key];
       if (variableDefinition.group === 'style' && variableDefinition.defaultValue !== undefined) {
         const overrideVariableData = componentVariablesOverwrites?.[variableData.key];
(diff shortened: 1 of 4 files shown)


Spring3 (approved):
Approving on my side, but I'd like to confirm with the others on our upcoming sync that they are fine with this patch for the time being

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -73,19 +75,31 @@ export const CompositionBlock = ({
...
+              'cfSsrClassName',
+            )
+          : undefined,
+      };
     }
 
+    const propMap: Record<string, PrimitiveValue> = {
+      // @ts-expect-error -- bla
+      cfSsrClassName: node.id
+        ? // @ts-expect-error -- bla
```
Chaoste, line 93: I know it's always the same reason but I would really avoid leaving a comment like "bla" in the code.
]
```suggestion
      // @ts-expect-error -- node receives a random ID in ssrStyles (no type support yet)
      cfSsrClassName: node.id
        ? // @ts-expect-error -- node receives a random ID in ssrStyles (no type support yet)
```

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -184,11 +198,27 @@ export const CompositionBlock = ({
 
   const { component } = componentRegistration;
 
+  const _getPatternChildNodeClassName = (childNodeId: string) => {
+    if (isAssembly) {
+      // @ts-expect-error -- bla
```
Chaoste, line 203:
```suggestion
      // @ts-expect-error -- property cfSsrClassName is a map (id to class name) that is added during rendering in ssrStyles
```

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -166,6 +169,8 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
           // and this is where the over-writes for the default values are stored
           // yes, I know, it's a bit confusing
           componentVariablesOverwrites: currentNode.variables,
+          // and this is where the pattern node
```
Chaoste, line 172: What is this comment supposed to express? Maybe adjust or just remove it :)

```suggestion
          // pass top-level pattern node to store instance-specific child styles for rendering
```

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -318,12 +322,28 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
...
+        patternWrapper.variables.cfSsrClassName = {
+          ...(patternWrapper.variables.cfSsrClassName ?? {}),
+          type: 'DesignValue',
+          // @ts-expect-error TODO: fix the types - add id to ComponentTreeNode
+          [currentNode.id]: {
+            valuesByBreakpoint: {
+              [breakpointIds[0]]: currentNodeClassNames.join(' '),
+            },
+          },
+        };
```
Chaoste, line 338:
```suggestion
        // @ts-expect-error -- temporarily add node ID to pick up the className during rendering in CompositionBlock
        currentNode.id = currentNode.id ?? generateRandomId(15);
        // @ts-expect-error -- intermediately store child classNames as map of idToValuesByBreakpoint
        patternWrapper.variables.cfSsrClassName = {
          ...(patternWrapper.variables.cfSsrClassName ?? {}),
          type: 'DesignValue',
          // @ts-expect-error -- The ID property is not yet officially typed nor supported
          [currentNode.id]: {
            valuesByBreakpoint: {
              [breakpointIds[0]]: currentNodeClassNames.join(' '),
            },
          },
        };
```

packages/experience-builder-sdk/src/core/preview/assemblyUtils.ts
```diff
@@ -71,6 +71,8 @@ export const deserializeAssemblyNode = ({
 
   return {
     definitionId: node.definitionId,
+    // @ts-expect-error TODO: fix this
```
Chaoste, line 74:
```suggestion
    // @ts-expect-error -- required for SSR of custom designs on pattern instances, not yet typed
```

packages/experience-builder-sdk/src/core/preview/assemblyUtils.ts
```diff
@@ -108,6 +110,8 @@ export const resolveAssembly = ({
   const deserializedNode = deserializeAssemblyNode({
     node: {
       definitionId: node.definitionId,
+      // @ts-expect-error TODO: fix this
```
Chaoste, line 113:
```suggestion
      // @ts-expect-error -- required for SSR of custom designs on pattern instances, not yet typed
```

Chaoste (changes requested):
As @Spring3 mentioned yesterday in the sync meeting, it will be hard to understand the rational for these changes at a later point. No matter how we follow up, we should document why the types are conflicting here and why we're doing that. I suggested how we could phrase comments to make it at least a bit more tangible.

anwaar931:
Hey @Chaoste, Thanks for suggestions on comments. I have already noticed to adjust these comments yesterday but was planning to update them today. 
Updated the comments. Would be great if you can take a look before your half day. 
Thanks

PR #713, opened 2024-08-08
feat(core): set background color opacity to 0 by default [ALT-887]

primeinteger:
## Purpose

We're changing the default background color opacity to 0 to work better with the design sidebar accordion in the UI

Ticket: https://contentful.atlassian.net/browse/ALT-887


sam-is-content (approved):

ryunsong-contentful (approved):
So this was the right choice all along. I knew it!

PR #715, opened 2024-08-08
feat: static elements iframe rendering

loweisz:

### DIFF
--- a/packages/experience-builder-sdk/src/ExperienceRoot.tsx
+++ b/packages/experience-builder-sdk/src/ExperienceRoot.tsx
@@ -32,8 +32,6 @@ export const ExperienceRoot = ({
     isEditorMode,
   });
 
-  console.log({ isEditorMode, isReadOnlyMode }, window.__EB__);
-
   if (isEditorMode) {
     return (
       <VisualEditorRoot
--- a/packages/experience-builder-sdk/src/ExperienceRoot.tsx
+++ b/packages/experience-builder-sdk/src/ExperienceRoot.tsx
@@ -22,7 +22,7 @@ export const ExperienceRoot = ({
   experience,
   visualEditorMode = VisualEditorMode.LazyLoad,
 }: ExperienceRootProps) => {
-  const isEditorMode = useDetectEditorMode();
+  const { isEditorMode, isReadOnlyMode } = useDetectEditorMode();
   //If experience is passed in as a JSON string, recreate it to an experience object
   const experienceObject =
     typeof experience === 'string' ? createExperience(experience) : experience;
@@ -32,12 +32,15 @@ export const ExperienceRoot = ({
     isEditorMode,
   });
 
+  console.log({ isEditorMode, isReadOnlyMode }, window.__EB__);
+
   if (isEditorMode) {
     return (
       <VisualEditorRoot
         experience={experienceObject as Experience<EntityStore> | undefined}
         visualEditorMode={visualEditorMode}
         initialLocale={locale}
+        isReadOnlyMode={isReadOnlyMode}
       />
     );
   }
--- a/packages/experience-builder-sdk/src/blocks/editor/VisualEditorRoot.tsx
+++ b/packages/experience-builder-sdk/src/blocks/editor/VisualEditorRoot.tsx
@@ -10,12 +10,15 @@ type VisualEditorRootProps = {
   visualEditorMode: VisualEditorMode;
   experience?: Experience<EntityStore>;
   initialLocale: string;
+
+  isReadOnlyMode?: boolean;
 };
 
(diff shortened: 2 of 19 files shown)


packages/experience-builder-sdk/src/hooks/useDetectEditorMode.tsx
```diff
@@ -35,8 +39,20 @@ export const useDetectEditorMode = ({ isClientSide = false }: UseDetectEditorMod
             window.__EB__ = {};
           }
           window.__EB__.isEditorMode = true;
-          window.removeEventListener('message', onMessage);
+          // window.removeEventListener('message', onMessage);
+        }
+      }
+
```
loweisz, line 45: Introducing a new event sent from the web app, that will set a flag similar to the editor mode to set it in read-only mode. 
We will also set the mode to editor mode, so that we enable the iframe communication and don't fetch staff over the network

packages/experience-builder-sdk/src/ExperienceRoot.tsx
```diff
@@ -38,6 +38,7 @@ export const ExperienceRoot = ({
         experience={experienceObject as Experience<EntityStore> | undefined}
         visualEditorMode={visualEditorMode}
         initialLocale={locale}
+        isReadOnlyMode={isReadOnlyMode}
       />
```
loweisz, line 42: For read only we still render the VisualEditorRoot, but we set it into read only

packages/experience-builder-sdk/src/blocks/editor/VisualEditorRoot.tsx
```diff
@@ -27,7 +30,11 @@ export const VisualEditorRoot: React.FC<VisualEditorRootProps> = ({
   return (
     <ErrorBoundary>
       <Suspense fallback={<div>Loading...</div>}>
-        <VisualEditorLoader experience={experience} visualEditorMode={visualEditorMode} />
+        <VisualEditorLoader
+          experience={experience}
+          isReadOnlyMode={isReadOnlyMode}
+          visualEditorMode={visualEditorMode}
+        />
```
loweisz, line 37: [3] Pass it on to the Visual editor

packages/visual-editor/src/components/VisualEditorRoot.tsx
```diff
@@ -60,5 +72,9 @@ export const VisualEditorRoot = ({ experience }: { experience?: Experience<Entit
 
   if (!initialized) return null;
 
+  if (isReadOnlyMode) {
+    return <ReadOnlyRenderer />;
+  }
+
```
loweisz, line 78: [4] Introducing a new Renderer that will handle elements diffently to the normal editor renderer

packages/visual-editor/src/components/ReadOnlyRenderer.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
loweisz, line 1: [5] This renderer will render the blocks directly, as we don't need the editor things like the dropzone and hover stuff.

packages/visual-editor/src/hooks/useReadOnlyComponent.tsx
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
```
loweisz, line 1: [6] Mostly copy-pasted code from the `useEditorSubscriber`, excluding all the events that we are not interested in, like the drag and drop stuff, hover or select events. 

this will mostly just listen to the component tree update events and will update the tree in here

packages/visual-editor/src/hooks/useReadOnlyComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const { children, innerRef, style, ...rest } = props;
+
+    const element = React.createElement(
+      ImportedComponentErrorBoundary,
+      null,
+      React.createElement(componentRegistration.component, {
+        ...modifiedProps,
+        style: { ...sizeStyles, ...style },
+        // forces the container elements to render without the dropzone
+        editorMode: false,
```
loweisz, line 98: [7] Creating the element that will be rendered for this node. We set it here directly to editor mode as this would prevent elements like the container to not render the dropzone

packages/visual-editor/src/components/ReadOnlyEditorBlock.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useMemo, useRef } from 'react';
```
loweisz, line 1: [8] Mostly the same as the blocks for the delivery mode just getting the information differently as we are already in the editor mode land. 

This might be abstracted and be shared with the delivery mode components ?

PR #716, opened 2024-08-12
fix: don't wrap rich text with an inline element

elylucas:
## Purpose

Adding design variables to a rich text component would not always display them properly.

## Approach

The rich text component was wrapping the outputs of the call to `documentToReactComponents` in a span tag, which is an inline level element. The markup that `documentToReactComponents` creates contains block level elements, and wrapping block level elements in an inline element can break layout flow (which we saw in this bug).

The issue was fixed by replacing the `span` tag with a `div` tag as the wrapper.


ryunsong-contentful (approved):
You'll need to update `components/RichText/RichText.cy.tsx` but otherwise good fix!

primeinteger (approved):

PR #717, opened 2024-08-20
[POC] feat: implement metadata prop for server side 3rd party data [ALT-1078]

chasepoirier:
## Purpose
Customers are unable to use 3rd party data fetched on the server within their custom components. This POC explores creating a lightweight global data store (inspired by vanilla zustand) that customers can use to fetch data from external sources which is then available to be used by components rendered from studio experiences.

Spike Document: https://contentful.atlassian.net/wiki/x/zIDmIAE


Chaoste:

PR #722, opened 2024-08-27
feat: "ReadOnlyMode" post message listener

ethan-ozelius-contentful:
### Context
What is read only mode?  Read only mode is a new mode (currently only preview, production & editor) we are working on to enable experience version history, as well as eventually governance.  In read only mode, a user can view an experience, but will be unable to make any changes or save the experience;  Visually, the user will not be able to drag/drop, and the UX features like selected/hovered components, and tool tips will not render.

[Spike document with more details](https://contentful.atlassian.net/wiki/spaces/PROD/pages/4825579580/Read-only+Editor)

## Purpose
Add a new post message listener in the `experience-builder-sdk` package, to listen for `"RequestReadOnlyMode"` messages from contentful user_interface.

### Launch Darkly Flag & feature flagging
There is an existing LD flag in user_interface to gate sending the `"RequestReadOnlyMode"` post message.

### Next steps
Once the SDK is listening for the `"requestEditorMode"` postMessage, we can create a new renderer specific to read only mode.

## Approach
I leveraged the existing `isEditorMode` functionality, as the communication pattern and outcomes will be very similar.  I decided to abstract `useDetectEditorMode` to `useDetectMode` and handle 2 messages instead of 1.  I'd be open to creating 2 separate hooks, but honestly I think the pros/cons either way would be negligible, but definitely open to ideas.

**Note:** Editor mode and read only mode are **mutually exclusive!**.  Only one mode can be enabled at one time to avoid confusion.


packages/core/src/types.ts
```diff
@@ -545,6 +545,7 @@ export type OutgoingMessage = {
 type Filter<T, U> = T extends U ? T : never;
 type SelectedValueTypes = Filter<ComponentPropertyValue['type'], 'UnboundValue' | 'BoundValue'>;
 
+export type RequestReadOnlyModePayload = undefined;
```
ethan-ozelius-contentful, line 548: Considering the idea of passing a tree _as_ the payload when requesting read-only mode.  I don't see why this should need to be a two step process of requesting read only mode, then requesting the component tree to render, but interested to hear what others think.

For example, in user interface, when opening up the version history modal, send the post message:
```js
const { tree } = useTreeStore();
const clonedTree = cloneDeep(tree);
const backwardsCompatibleClonedTree = createBackwardCompatibleTree(clonedTree);

sendMessage(CANVAS_ACTIONS.RequestReadOnlyMode, {
  tree: backwardsCompatibleClonedTree
  locale
});  
```

Unless anyone has any objections I'm going to implement that pattern now.

packages/experience-builder-sdk/src/hooks/useDetectMode.tsx
```diff
@@ -28,29 +30,45 @@ export const useDetectEditorMode = ({ isClientSide = false }: UseDetectEditorMod
...
         if (typeof window !== 'undefined') {
           //Once we definitely know that we are in editor mode, we set this flag so future postMessage connect calls are not made
           if (!window.__EB__) {
             window.__EB__ = {};
           }
           window.__EB__.isEditorMode = true;
-          window.removeEventListener('message', onMessage);
         }
+      } else if (eventData.eventType === INCOMING_EVENTS.RequestReadOnlyMode) {
+        setIsReadOnlyMode(true);
```
Chaoste, line 42:
```suggestion
        setIsReadOnlyMode(true);
        setIsEditorMode(false);
```

To be safe, it would ensure that the other mode is turned off (same in line 32)

Chaoste (commented):
The approach looks good to me 👍

lilbitner (commented):
This makes sense to me! Appreciate the added tests 💯

packages/experience-builder-sdk/src/hooks/useDetectMode.tsx
ethan-ozelius-contentful: Totally agree, this was something that @jsdalton also called out via Slack.  I've since updated to the code so that only 1 mode can be enabled at one time.

At any given time, mode will be either `MODE.READ_ONLY` `MODE.EDITOR` or `MODE.NONE`.

packages/core/src/types.ts
ethan-ozelius-contentful: DONE!

packages/experience-builder-sdk/src/hooks/useDetectMode.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (mounted) {
+      setMode(inIframe() ? StudioExperienceMode.EDITOR : StudioExperienceMode.NONE);
+      // Double check if we are in editor mode by listening to postMessage events
+      if (typeof window !== 'undefined') {
+        window.addEventListener('message', onMessage);
+        sendMessage(OUTGOING_EVENTS.Connected, undefined);
+        sendMessage(OUTGOING_EVENTS.SDKFeatures, sdkFeatures);
+
+        setTimeout(() => {
+          if (!receivedEditorModeMessage.current) {
```
ethan-ozelius-contentful, line 81: TODO: update this condition to see if either readonly, or editor mode have been requested.   If neither have been requested in 100MS, default to NONE.  this condition is currently checking only for editor mode.

- [ ] do it.

packages/core/src/constants.ts
```diff
@@ -54,6 +55,12 @@ export const VISUAL_EDITOR_EVENTS = {
   Ready: 'cfVisualEditorReady',
 };
 
+export enum StudioExperienceMode {
```
Chaoste, line 58: It's kinda funny that the `mode` is having a [comeback](https://github.com/contentful/experience-builder/pull/341/files#diff-c4805c4c6737dc23dbf80cdc5ee3b59903b10e7a76d2859615442821deb714ddR34) in this new shape. We used to have a mode to separate preview, delivery, and editor mode which we deprecated and removed after a while.
Maybe, we should make it somehow clear (e.g. through a comment) that this mode is only for the editor experience inside the iframe and not about preview/ delivery.

ethan-ozelius-contentful: Good call out, I was not aware of the history of "modes" in studio.  That same question entered my mind yesterday, for what it's worth, I created a poll to see how people referred to "editor/read-only modes".

Screenshot 2024-09-04 at 9 06 14 AM

@jsdalton, had a particularly good point, which is what led me decide on `StudioExperienceMode`.

> the canvas view is what you showed, where you can drag and drop and use the toolbars, but in a way it's a subset of the "experience editor" -- which doesn't currently support many other modes of operation other than canvas mode, but could in the future.e

* takes a breather * 😂

All that being said, I agree with you that maybe we more intentionally name this to be just for editor/read-only mode...  In a way that's already clear since the actual values of the enum express it, but that doesn't mean someone couldn't come along 6 months from now and wonder "why doesn't this include preview/delivery?!?!?".

In my weakly held opinion, I like the name `StudioExperienceMode`, or maybe `StudioCanvasExperienceMode`, and a comment should make it exactly clear (hopefully).

Chaoste: Thanks for checking and sharing that poll. Very interesting & surprising insight :)
As long as we're aware of this potential confusion, I'm happy with going forward as is.

packages/experience-builder-sdk/src/ExperienceRoot.tsx
```diff
@@ -22,17 +23,20 @@ export const ExperienceRoot = ({
   experience,
   visualEditorMode = VisualEditorMode.LazyLoad,
 }: ExperienceRootProps) => {
-  const isEditorMode = useDetectEditorMode();
+  const mode = useDetectMode();
```
ethan-ozelius-contentful, line 26: remove

- [ ] remove

matthew-gordon (approved):
🎉 🌮

PR #723, opened 2024-08-29
feat: add style prop to toggle visibility in component registration [SPA-2371]

ChidinmaOrajiaku:
## Purpose
Introduce a cfVisibility style prop variable to the default style definitions

## Approach

- Add ` cfVisibility` to cf component default style definitions
- For custom components, add ` cfVisibility` default variable on component registration

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -56,7 +56,7 @@ export const builtInStyles: VariableDefinitions = {
     defaultValue: 'center',
     displayName: 'Horizontal alignment',
   },
-  cfCanToggleVisibility: {
+  cfVisibility: {
     displayName: 'Visibility toggle',
     type: 'Boolean',
     group: 'style',
@@ -341,7 +341,7 @@ export const containerBuiltInStyles: VariableDefinitions = {
 };
 
 export const dividerBuiltInStyles: VariableDefinitions = {
-  cfCanToggleVisibility: builtInStyles.cfCanToggleVisibility,
+  cfVisibility: builtInStyles.cfVisibility,
   cfBorderRadius: optionalBuiltInStyles.cfBorderRadius,
   cfMargin: {
     displayName: 'Margin',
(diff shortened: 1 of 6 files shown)


packages/core/src/definitions/styles.ts
```diff
@@ -56,6 +56,13 @@ export const builtInStyles: VariableDefinitions = {
     defaultValue: 'center',
     displayName: 'Horizontal alignment',
   },
+  cfVisibility: {
+    displayName: 'Visibility toggle',
+    type: 'Boolean',
+    group: 'style',
+    defaultValue: false,
+    description: 'To enable visibility toggle in patterns',
```
Spring3, line 64: Is it only going to apply to the patterns or to all registered components? 🤔

ChidinmaOrajiaku: Thanks for pointing that out

packages/core/src/definitions/styles.ts
```diff
@@ -56,6 +56,13 @@ export const builtInStyles: VariableDefinitions = {
     defaultValue: 'center',
     displayName: 'Horizontal alignment',
   },
+  cfVisibility: {
+    displayName: 'Visibility toggle',
+    type: 'Boolean',
+    group: 'style',
+    defaultValue: false,
```
Chaoste, line 63:
```suggestion
    defaultValue: true,
```
I thought nodes should be visible, per default

packages/core/src/definitions/styles.ts
```diff
@@ -56,6 +56,13 @@ export const builtInStyles: VariableDefinitions = {
...
+    description: 'To enable visibility toggle in components',
```
Chaoste, line 64: [optional] (not sure if this is of any value but I propose aligning it with the other descriptions in here)
```suggestion
    description: 'The visibility of the component',
```

Chaoste (changes requested):
It's all minor but I had some thoughts. LMK what you think and if I should give it a second look :)

Spring3 (approved):
Agree with changes requested by @Chaoste . I don't have anything more to add :)

Chaoste (approved):

PR #726, opened 2024-08-30
fix: default value for container margin to be auto left/right [ALT-1166]

chasepoirier:
## Purpose
Second half of fixed margin for the Container component. We needed to fix the default value to be `margin: 0 auto 0 auto`.

Ticket: https://contentful.atlassian.net/browse/ALT-1166
Linked PR: https://github.com/contentful/user_interface/pull/23002


primeinteger (approved):
🚀

PR #730, opened 2024-09-05
fix: type error when breakpoint is undefined

SofiaMargariti:
## Purpose

Make sure that `getValueByBreakpoint` doesn't crash if current index doesn't match any breakpoints.
Fixes https://contentful.sentry.io/issues/5794748829/

## Approach

`getValueByBreakpoint` should be more defensive against a `currentIndex` that doesn't match any breakpoints otherwise the above error will happen and will crash the app.

Normally `currentIndex` should correspond to a breakpoint but since we cannot ensure that the caller has verified this we should handle this case more gracefully. It seems that a customer has already ran into this issue but since I cannot reproduce it on my side it's hard to tell where in the workflow the current index gets a wrong value.

### DIFF
--- a/packages/core/src/utils/breakpoints.ts
+++ b/packages/core/src/utils/breakpoints.ts
@@ -109,7 +109,7 @@ export const getValueForBreakpoint = (
   if (valuesByBreakpoint instanceof Object) {
     // Assume that the values are sorted by media query to apply the cascading CSS logic
     for (let index = activeBreakpointIndex; index >= 0; index--) {
-      const breakpointId = breakpoints[index].id;
+      const breakpointId = breakpoints[index]?.id;
       if (valuesByBreakpoint[breakpointId]) {
         // If the value is defined, we use it and stop the breakpoints cascade
         return eventuallyResolveDesignTokens(valuesByBreakpoint[breakpointId]);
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/breakpoints.ts
```diff
@@ -109,16 +109,18 @@ export const getValueForBreakpoint = (
...
       if (valuesByBreakpoint[breakpointId]) {
         // If the value is defined, we use it and stop the breakpoints cascade
         return eventuallyResolveDesignTokens(valuesByBreakpoint[breakpointId]);
       }
     }
     // If no breakpoint matched, we search and apply the fallback breakpoint
     const fallbackBreakpointIndex = getFallbackBreakpointIndex(breakpoints);
-    const fallbackBreakpointId = breakpoints[fallbackBreakpointIndex].id;
-    return eventuallyResolveDesignTokens(valuesByBreakpoint[fallbackBreakpointId]);
+    const fallbackBreakpointId = breakpoints[fallbackBreakpointIndex]?.id;
```
SofiaMargariti, line 120: This is probably never going to fail as long as there is at least one breakpoint available but doesn't hurt to be a bit more defensive here

Chaoste (approved):
Thanks for looking into this. I checked on this yesterday as well but couldn't find a case based on our SDK logic how this could fail. I can only think of two scenarios:
- As the getFallbackBreakpointId is already defensively written, the array might just be empty.
- The customers calls this function manually. Unfortunately, the Stack trace is not helpful for this in Sentry

PR #737, opened 2024-09-10
feat: introduce rendering cfVisibility in SDK

Chaoste:
## Approach

Add `cfVisibility` to the set of CF-controlled attributes and translate it to CSS using `display: none`.

## TODO

- [x] Test with test app
- [ ] Test with SSR


ChidinmaOrajiaku (approved):

Chaoste:
Merged this together with https://github.com/contentful/experience-builder/pull/739

PR #739, opened 2024-09-10
fix: allow design value to be falsy for certain breakpoint [SPA-2391]

Chaoste:
## Purpose

Setting a design value like `cfTextItalic` to `false` for a certain breakpoint doesn't work as the conditional logic in the SDK would fall back to the next truthy value in the set of breakpoint values.

## Approach

Only cascade through the breakpoint values if the value at hand is `undefined | null | ''`

### DIFF
--- a/packages/core/src/utils/breakpoints.ts
+++ b/packages/core/src/utils/breakpoints.ts
@@ -91,6 +91,10 @@ const builtInStylesWithDesignTokens = [
   'cfMaxWidth',
 ];
 
+const isValidBreakpointValue = (value: PrimitiveValue) => {
+  return value !== undefined && value !== null && value !== '';
+};
+
 export const getValueForBreakpoint = (
   valuesByBreakpoint: ValuesByBreakpoint,
   breakpoints: Breakpoint[],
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/breakpoints.ts
```diff
@@ -110,19 +114,20 @@ export const getValueForBreakpoint = (
...
     const fallbackBreakpointId = breakpoints[fallbackBreakpointIndex]?.id;
-    if (valuesByBreakpoint[fallbackBreakpointId]) {
+    if (isValidBreakpointValue(valuesByBreakpoint[fallbackBreakpointId])) {
       return eventuallyResolveDesignTokens(valuesByBreakpoint[fallbackBreakpointId]);
     }
   } else {
     // Old design properties did not support breakpoints, keep for backward compatibility
     return valuesByBreakpoint;
   }
+  return undefined;
```
SofiaMargariti, line 132: You probably don't need to explicitly return undefined here, it would return `undefined` anyway if it reached this point

SofiaMargariti (approved):

ChidinmaOrajiaku (approved):

PR #742, opened 2024-09-11
fix: assemblies incorrectly rendered on initial drop [ALT-1220]

chasepoirier:
## Purpose
Our SDK logic to immediately add a component to the local tree was incorrectly adding an assembly as a regular block on the initial render. This resulted in broken behavior where editor mode props were not passed properly to an assembly and therefore breaking DND.

This fix adds functionality so that we know if we are dragging an assembly type node during a component add action and if we are, we prevent the auto-creation of a block node.

Ticket: https://contentful.atlassian.net/browse/ALT-1220

Related user_interface PR: https://github.com/contentful/user_interface/pull/23249

### DIFF
--- a/packages/visual-editor/src/hooks/useCanvasInteractions.ts
+++ b/packages/visual-editor/src/hooks/useCanvasInteractions.ts
@@ -20,6 +20,10 @@ export default function useCanvasInteractions() {
       return;
     }
 
+    /**
+     * We only have the draggableId as information about the new component being dropped.
+     * So we need to split it to get the blockId and the isAssembly flag.
+     */
     const [blockId, isAssembly] = draggableId.split(':');
 
     const { nodeId: parentId, slotId } = parseZoneId(destination.droppableId);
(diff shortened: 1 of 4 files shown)


packages/visual-editor/src/hooks/useCanvasInteractions.ts
```diff
@@ -20,32 +20,36 @@ export default function useCanvasInteractions() {
       return;
     }
 
+    const [blockId, isAssembly] = draggableId.split(':');
```
Chaoste, line 23: Please note that `isAssembly` is now a string `'false'` or `'true'` and thus always truthy. This will backfire in line 47.
In general, do you think there is a better approach compared to merging and again splitting this string which is supposed to be an "id".

Maybe, we can save us from all of this with the proposal I wrote in the [other PR](https://github.com/contentful/user_interface/pull/23249/files#r1753179850).

Chaoste (approved):

PR #751, opened 2024-09-16
feat: export deserialize pattern node variables from core package [SPA-2376]

anwaar931:
## Purpose
We will be using this similar piece of code in user interface as well. So better to not repeat the code.

### DIFF
--- a/packages/core/src/utils/patternUtils.ts
+++ b/packages/core/src/utils/patternUtils.ts
@@ -1,0 +1,65 @@
+import type {
+  ComponentPropertyValue,
+  ExperienceDataSource,
+  ExperienceUnboundValues,
+  ExperienceComponentSettings,
+  DesignValue,
+} from '../types';
+
+export const deserializePatternVariables = ({
+  nodeVariables,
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/patternUtils.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  for (const [variableName, variable] of Object.entries(nodeVariables)) {
+    childNodeVariable[variableName] = variable;
+    if (variable.type === 'ComponentValue') {
+      const componentValueKey = variable.key;
+      const instanceProperty = componentInstanceProps[componentValueKey];
+      const variableDefinition = patternVariableDefinitions?.[componentValueKey];
+      const defaultValue = variableDefinition?.defaultValue;
+
+      // For assembly, we look up the value in the assembly instance and
```
Chaoste, line 38: [nitpick]
```suggestion
      // For pattern, we look up the value in the pattern instance and
```

Chaoste (approved):
LGTM

PR #754, opened 2024-09-18
fix: resolve fetched entities for experiences [ALT-1146]

elylucas:
## Purpose

Removes the `withoutLinkResolution` chain modifier so that any entries embedded into the fetched entry are resolved in its response. Fixes bug [ALT-1146].

This change also brings in line how the experiences are fetched in delivery and editor modes to be (a bit) more aligned. For instance, the call to fetch the entities in the UI included referenced entities up to 10 levels deep. The query was modified to add the "includes: 10" parameter so that querying the data in each place is more like the other.

## Approach

[ALT-1146]: https://contentful.atlassian.net/browse/ALT-1146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/experience-builder-sdk/src/hooks/useFetchBySlug.spec.ts
+++ b/packages/experience-builder-sdk/src/hooks/useFetchBySlug.spec.ts
@@ -158,7 +158,6 @@ describe('useFetchBySlug', () => {
     await waitFor(() => {
       expect(result.current.error).toBeUndefined();
       expect(clientMock.getEntries).toHaveBeenCalledTimes(1);
-      expect(clientMock.getEntries).toHaveBeenCalledTimes(1);
     });
   });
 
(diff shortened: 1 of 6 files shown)


packages/experience-builder-sdk/src/hooks/useFetchBySlug.spec.ts
```diff
@@ -159,7 +158,7 @@ describe('useFetchBySlug', () => {
     await waitFor(() => {
       expect(result.current.error).toBeUndefined();
       expect(clientMock.getEntries).toHaveBeenCalledTimes(1);
-      expect(clientMock.withoutLinkResolution.getEntries).toHaveBeenCalledTimes(1);
+      expect(clientMock.getEntries).toHaveBeenCalledTimes(1);
```
primeinteger, line 161: Same as the expect statement above it

primeinteger (approved):
👍

PR #756, opened 2024-09-23
chore: deploy marketing demo app to vercel

elylucas:
## Purpose

Deploys preview and production sites to Vercel

## Approach

Since Vercel is already hooked up to deploy the test app via commits, I added a CI job to deploy the marketing demo via the Vercel CLI. This also makes it so that the deploy only happens if the checks pass in CI.


packages/templates/nextjs-marketing-demo/package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@contentful/experiences-sdk-react": "^1.16.0",
+    "@contentful/experiences-sdk-react": "file:../../experience-builder-sdk",
```
chasepoirier, line 12: Just wondering, do we want to have this template also take in the latest SDK or would it be better to scope it to a specific version for better stability for their demo site?

I know that would be move work for us to maintain but curious what you think

elylucas: I was thinking the template will always use the latest sdk. The CLI work I did in another PR installs the latest sdk when you scaffold the app.

primeinteger (approved):
🚀

PR #759, opened 2024-09-24
[ALT-1246] feat: add custom rating to the marketing template

ryunsong-contentful:
## Purpose
Get the Ant Customer Rating component into the template demo app https://ant.design/components/rate

Screenshot 2024-09-24 at 1 40 50 PM
Screenshot 2024-09-24 at 1 41 00 PM

### DIFF
--- a/packages/templates/nextjs-marketing-demo/src/components/CustomRatingComponentRegistration.tsx
+++ b/packages/templates/nextjs-marketing-demo/src/components/CustomRatingComponentRegistration.tsx
@@ -1,0 +1,48 @@
+import { Rate } from 'antd';
+
+export const CustomRatingComponentRegistration = {
+  component: Rate,
+  definition: {
+    id: 'custom-rating',
+    name: 'Custom Rating',
+    category: 'Custom Components',
+    variables: {
+      disabled: {
+        displayName: 'Read only',
+        type: 'Boolean',
+        defaultValue: true,
+        group: 'style',
+      },
+      value: {
+        displayName: 'Stars',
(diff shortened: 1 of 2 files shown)


packages/templates/nextjs-marketing-demo/src/components/CustomRatingComponentRegistration.tsx
```diff
@@ -0,0 +1,52 @@
+import { Rate } from 'antd';
+type CustomRatingComponentProps = React.ComponentProps<typeof Rate>;
+
+export const CustomRatingComponentRegistration = {
+  component: (props: CustomRatingComponentProps) => {
```
elylucas, line 5: since we don't need to modify the props anyway coming in, I think we can just use the Rate component here instead of wrapping it

packages/templates/nextjs-marketing-demo/src/components/CustomRatingComponentRegistration.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  definition: {
+    id: 'custom-rating',
+    name: 'Custom Rating',
+    category: 'Custom Components',
+    variables: {
+      disabled: {
+        displayName: 'Read only',
+        type: 'Boolean',
+        defaultValue: false,
```
elylucas, line 17: I think we might want to make the Read only true by default since I don't really see anyone interacting with for the demo

elylucas (changes requested):

elylucas (approved):

PR #761, opened 2024-09-25
feat: add nextjs marketing demo app to list of cli templates [ALT-1265]

elylucas:
## Purpose

This PR adds the upcoming nextjs marketing demo app to the list of templates you can use to scaffold a new Studio app with the CLI. Other changes came along with it to support multiple templates.


packages/templates/nextjs-marketing-demo/src/getExperience.ts
```diff
@@ -8,6 +8,8 @@ const environment = process.env.NEXT_PUBLIC_CTFL_ENVIRONMENT!;
 const experienceTypeId = process.env.NEXT_PUBLIC_CTFL_EXPERIENCE_TYPE!;
 const domain = process.env.NEXT_PUBLIC_CTFL_DOMAIN || 'contentful.com';
 
+console.log('accessToken', accessToken, 'prevAccessToken', prevAccessToken);
```
ryunsong-contentful, line 11: Is this intentional? Could be better ways to check if the token exists or not (ie check against undefined and throw errors) if that's the intent of the console log

ryunsong-contentful (commented):
Overall the code changes look good. Is there a way I can test this PR? I imagine it involves pulling the branch but after that, is there something specific that would give us confidence that it definitely works? 

Would it be something along the lines of pulling the branch, running npm i and npm run build and then `npm run dev` for the template nextjs marketing demo app?

packages/templates/nextjs-marketing-demo/src/getExperience.ts
elylucas: nope, forgot to remove it, good catch

elylucas:
>  Is there a way I can test this PR? 

I updated the readme to include instructions for developing locally. Let me know if they don't make sense.

> What about making sure things deploy correctly? Would that be covered by github actions?

The CLI tool itself is deployed as a part of our normal CI pipeline. The templates that are included in that particular build will be the ones that were built along with the tools, as, right now, the templates are embedded into the CLI package.

ryunsong-contentful:
Cool, will take a look at your README and try it out

ryunsong-contentful:
So I ran the CLI tool and works as expected in the happy case. However, in the bad workflow, it hangs as follows in the screenshot. Just spins and spins
Screenshot 2024-09-25 at 3 43 02 PM

elylucas:
> So I ran the CLI tool and works as expected in the happy case. However, in the bad workflow, it hangs as follows in the screenshot. Just spins and spins at `No content preview environments for Studio Experiences currently found, creating one..` Screenshot 2024-09-25 at 3 43 02 PM
> 
> Now we can probably log this as a separate bug or we can fix it in this PR. We can probably discuss the path forward tomorrow in person

This would be an already existing issue. We probably don't do a good job doing validation or handling errors now. We should file it as a separate ticket so it doesn't block this work.

PR #764, opened 2024-09-25
feat(nextjs-marketing-demo): add Header, Footer, and light mode styling [ALT-1270]

primeinteger:
## Purpose

For the Marketing Demo app, I added a Header, Footer, and light mode styling to align with the Figma design.

Screenshot 2024-09-25 at 3 53 10 PM


packages/templates/nextjs-marketing-demo/src/components/layout/Footer.tsx
```diff
@@ -0,0 +1,18 @@
+'use client';
```
elylucas, line 1: Is there a reason this and the header need to be client components?

packages/templates/nextjs-marketing-demo/src/components/layout/Header.tsx
```diff
@@ -0,0 +1,38 @@
...
+const Header: React.FC = () => {
+  const [current, setCurrent] = useState('home-page');
+
+  return (
+    <div className={styles.header}>
+      <div className={styles.logo}>
+        <span>Logo</span>
+      </div>
+      <div className={styles.nav}>
+        <Menu
```
elylucas, line 23: I can see theres some interactivity here, might be good to put Menu into its own component and mark that one as client so the Header itself can remain a RSC.

elylucas (approved):
Added a couple of nit pick comments on the use client components. Probably not too big of a deal so marking as approved.

packages/templates/nextjs-marketing-demo/src/components/layout/Header.tsx
primeinteger: Good call, I'll separate the Menu into its own component.

packages/templates/nextjs-marketing-demo/src/components/layout/Footer.tsx
primeinteger: Definitely not needed in this one. It snuck in when I copy/pasted the Header component to create the Footer and I overlooked this line

PR #775, opened 2024-09-30
fix: default width to 100% when component doesn't have a width variable [ALT-956]

elylucas:
## Purpose

The image optimization code was expecting the component to have a cfWidth variable set, and error if it did not. Since the column component did not have cfWidth, it was erroring when trying to set the background image on a column.

## Approach

When seeing if there is a cfWidth set, if there is not, we default to 100% width instead and let the user set a Target Width on the variable to help with optimization.


primeinteger (approved):
I tested this fix locally and it's working well. I can bind a background image to Single Column component now 🎉

PR #777, opened 2024-10-01
fix: override browser hr border style + set default background color [ALT-1281]

sam-is-content:
## Purpose

Could not see color change on divider component unless you increased the divider height - this fixes that

https://contentful.atlassian.net/browse/ALT-1281

## Approach

Root issue:
the user style sheet (the browser's default styles) has a border for <hr> (the divider tag) so when our divider was very short, a user basically only saw the border and was thus unable to see the color change they applied.

Solution:
Override the user style sheet border with 'border: none' on the hr and then set the default background-color on the divider to be gray instead of transparent so users can see the divider (since previously when users inserted a divider they initially mainly saw the border)

https://www.loom.com/share/e6e5ed8b3f2a4524bf737071da2b60e2?sid=e724b599-7ddb-4c6b-b40f-8f7f74108d0b

### DIFF
--- a/packages/components/src/components/Divider/ContentfulDivider.tsx
+++ b/packages/components/src/components/Divider/ContentfulDivider.tsx
@@ -1,5 +1,12 @@
 import React from 'react';
 import './ContentfulDivider.css';
+import { css, cx } from 'emotion';
+
+const styles = {
+  hr: css({
+    border: 'none',
+  }),
+};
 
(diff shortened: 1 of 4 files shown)


ryunsong-contentful (approved):
Noticed that the default divider height it 2px. Just double check with Travis that 2px is the correct default over 1px before merging
Screenshot 2024-10-01 at 4 28 59 PM

PR #778, opened 2024-10-02
fix: width applied to both wrapping div and img [ALT-1293]

primeinteger:
## Purpose

Fixes a bug where the `width` style is being applied to both the wrapping div and element. 

This was especially noticeable when using percent widths on an Image component because the double-width makes the image half the expected size.

## Approach

For editor mode, I updated `useCompnoentProps` to set the height and width to `100%` when those styles are applied to the wrapping div.

For delivery mode, I added a new `PreviewUnboundImage` component in the SDK.  This component replicates the same behavior as editor mode for the unbound image placeholder element. A wrapping is styled with the image component's height and width, and the placeholder element is styled with `100%` height and width to match the editor mode appearance.

https://github.com/user-attachments/assets/4d51edd4-1cb8-479c-a37c-c4c4ee6653ca

https://github.com/user-attachments/assets/5f7b2440-c6e3-454c-bb75-a19a613c94be

### DIFF
--- a/packages/components/src/components/Image/Image.css
+++ b/packages/components/src/components/Image/Image.css
@@ -2,6 +2,8 @@
 
 .cf-no-image {
   position: relative;
+  height: 100%;
+  width: 100%;
 }
 
 .cf-no-image img {
--- a/packages/components/src/components/Image/Image.css
+++ b/packages/components/src/components/Image/Image.css
@@ -1,18 +1,12 @@
 @import url('../variables.css');
 
-.cf-no-image {
-  position: relative;
-  height: 100%;
-  width: 100%;
-}
-
-.cf-no-image img {
+img.cf-placeholder-image {
   background-color: var(--cf-color-gray100);
   outline-offset: -2px;
   outline: 2px solid rgba(var(--cf-color-gray400-rgb), 0.5);
 }
 
-.cf-no-image svg {
+svg.cf-placeholder-icon {
   position: absolute;
   top: 50%;
   left: 50%;
@@ -23,6 +17,6 @@
   max-width: 100%;
 }
 
-.cf-no-image svg path {
+svg.cf-placeholder-icon path {
   fill: var(--cf-color-gray400);
 }
--- a/packages/core/src/utils/components.ts
+++ b/packages/core/src/utils/components.ts
@@ -42,7 +42,7 @@ export const isComponentAllowedOnRoot = ({ type, category, componentId }: Compon
   componentId === CONTENTFUL_COMPONENTS.divider.id;
 
 export const isStructureWithRelativeHeight = (componentId?: string, height?: string | number) => {
-  return isContentfulStructureComponent(componentId) && height?.toString().endsWith('%');
+  return isContentfulStructureComponent(componentId) && !height?.toString().endsWith('px');
 };
 
 const stylesToKeep = ['cfImageAsset'];
--- a/packages/core/src/utils/components.ts
+++ b/packages/core/src/utils/components.ts
@@ -1,12 +1,9 @@
-import { omit } from 'lodash-es';
 import {
   ASSEMBLY_BLOCK_NODE_TYPE,
   ASSEMBLY_DEFAULT_CATEGORY,
   ASSEMBLY_NODE_TYPE,
-  CF_STYLE_ATTRIBUTES,
   CONTENTFUL_COMPONENTS,
 } from '../constants';
-import type { PrimitiveValue } from '../types';
 
 const structureComponentIds = new Set([
   CONTENTFUL_COMPONENTS.section.id,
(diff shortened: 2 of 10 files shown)


packages/core/src/utils/components.ts
```diff
@@ -41,3 +44,11 @@ export const isComponentAllowedOnRoot = ({ type, category, componentId }: Compon
 export const isStructureWithRelativeHeight = (componentId?: string, height?: string | number) => {
   return isContentfulStructureComponent(componentId) && !height?.toString().endsWith('px');
 };
+
+const stylesToKeep = ['cfImageAsset'];
+const stylesToRemove = CF_STYLE_ATTRIBUTES.filter((style) => !stylesToKeep.includes(style));
+const propsToRemove = ['cfHyperlink', 'cfOpenInNewTab', 'cfSsrClassName'];
+
+export const sanitizeNodeProps = (nodeProps: Record<PropertyKey, PrimitiveValue>) => {
```
elylucas, line 52: As I found out when cleaning up the deps for NumericUnitInput, continously adding util functions to a file can cause dep havoc. Any component that needs to import `sanitizeNodeProps` will have to import everything in this file, and if theres some deps in here with dependency violations, any subsequent components could unknowingly inherit those violations.

Therefore, its probably safer to put each util into their own file

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -263,14 +266,17 @@ export const CompositionBlock = ({
...
+    node.variables.cfImageAsset?.type === 'UnboundValue'
+  ) {
+    return <PreviewUnboundImage node={node} nodeProps={nodeProps} component={component} />;
+  }
 
   return React.createElement(
     component,
     {
-      ...omit(nodeProps, stylesToRemove, ['cfHyperlink', 'cfOpenInNewTab', 'cfSsrClassName']),
+      ...sanitizeNodeProps(nodeProps),
```
elylucas, line 279: Will this fix the errors we get about unknown props being spread onto html elements that have popped up again?

packages/core/src/utils/components.ts
elylucas: Also just realized this is the experiences repo, not UI, so probably not that big of a deal here, but could be in the future.

primeinteger: Good point. I'll move this to its own file.

packages/components/src/components/Image/Image.css
```diff
@@ -2,6 +2,8 @@
 
 .cf-no-image {
   position: relative;
+  height: 100%;
```
elylucas, line 5: wouldn't we use the styling that comes from the component bindings for height and width, even if there was no image?

ryunsong-contentful:
Seems to also fix that pattern height bug as well!

Screenshot 2024-10-02 at 2 01 57 PM

packages/visual-editor/src/hooks/useComponentProps.ts
```diff
@@ -218,20 +217,27 @@ export const useComponentProps = ({
 
   const cfStyles = useMemo(() => buildCfStyles(props as StyleProps), [props]);
 
+  const isAssemblyBlock = node.type === 'assemblyBlock';
+  const isSingleColumn = node?.data.blockId === CONTENTFUL_COMPONENTS.columns.id;
+  const isStructureComponent = isContentfulStructureComponent(node?.data.blockId);
+
+  // Move size styles to the wrapping div and override the component styles
+  const overrideStyles: CSSProperties = {};
   const sizeStyles: CSSProperties = {
```
sam-is-content, line 226: wrapperStyles❗ ❗ ❗ ❗ ❗ ❗

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
primeinteger: Probably not, unfortunately. I didn't actually change this code, I just moved it to a util file to make it reusable.  We had this same code duplicated in 2 files and I needed to do the same thing in 3rd file.  But at least now there is a single source of these bits instead of multiple copies.

packages/components/src/components/Image/Image.css
primeinteger: We do apply the Image size when there is no image bound.  The new `<PreviewUnboundImage>` component handles that, and applies the image size styles to a wrapper element.

Your comment made me realize that we won't really need this `.cf-no-image` element any more.  Previously this wrapper was used for centering the svg icon within the placeholder element.  With the new `<PreviewUnboundImage>` component providing a wrapper, this `.cf-no-image` wrapper can go away.  I refactored this a little bit and added some classes so it's easier to find where each element comes from in the codebase.

Screenshot 2024-10-02 at 3 20 00 PM

* `.cf-preview-unbound-image` is the wrapper in `<PreviewUnboundImage>` which receives the image size styles
* `.cf-placeholder-image` and `.cf-placeholder-icon` are the placeholder elements in the `<Image>` component when no image is bound

elylucas (approved):

PR #779, opened 2024-10-03
fix: revised buildCfStyles and add useComponentProps tests [ALT-1293]

primeinteger:
## Purpose

Add tests for the updated height/width styles logic in `useComponentProps` hook (added by #778)

## Approach

While debugging the tests I noticed that `buildCfStyles` adds `text-decoration: none` and `font-style: normal` on every element that it generates styles for.

That's not really necessary since those are css default values, so not including the style gives the same result.  I adjusted `buildCfStyles` so that it won't attach those default styles on every element so it's footprint is slightly smaller and more intentional.


chasepoirier (approved):
Good stuff!

PR #782, opened 2024-10-04
fix: columns uses incorrect default value []

chasepoirier:
## Purpose
Columns is set to margin `0 auto 0 auto` from the user_interface but the default is `0 0 0 0` in the SDK which results in the change indicator showing on initial drop of the columns component.


ryunsong-contentful (approved):
Nice subtle fix

PR #783, opened 2024-10-04
fix: update NextJS apps to import config server side so design tokens are registered [ALT-1314]

elylucas:
## Purpose

Design tokens were not being used for our internal apps on NextJS in SSR.

## Approach

The config to register components and design tokens was only being called in client components, so when the CSS was generated server side, it was unable to look up the design tokens being used.

This PR updates the apps to also include the configuration file on server side pages as well as client side pages.


ryunsong-contentful (approved):
Lol to think all we needed to do it to just define design tokens. Small suggestion, might be worth adding a CSS variable or two if there's one available in the app

PR #784, opened 2024-10-04
fix: apply margin to wrapperStyles when drag wrapper is enabled [ALT-1293]

primeinteger:
## Purpose

Fixes the Divider component after the changes in #778 caused a regression with the Divider's position

## Approach

The Divider positioning issue was due to the Divider component having Margin on by default.  This issue was affecting any component that applies Margins and Relative height/width.

I updated the `useComponentProps` hook to apply the Margin styles to `wrapperStyles` while removing the Margin from the component via `overrideStyles` which positions the component correctly when mixing margins + relative height/width.

I adjusted the conditions in `useComponentProps` so that the `wrapperStyles` and `overrideStyles` will only be used when the hook is called with `requireDragWrapper: true` -- making sure that we aren't accidentally using `wrapperStyles` unless the wrapper is actually being used.

## Video

https://github.com/user-attachments/assets/1c12a928-f5b4-465d-9d45-9e50a53ddff0

### DIFF
--- a/packages/components/src/components/Divider/ContentfulDivider.tsx
+++ b/packages/components/src/components/Divider/ContentfulDivider.tsx
@@ -1,22 +1,14 @@
 import React from 'react';
 import './ContentfulDivider.css';
-import { css, cx } from 'emotion';
-
-const styles = {
-  hr: css({
-    border: 'none',
-  }),
-};
 
 export type ContentfulDividerProps = {
   className?: string;
 };
 
-export const ContentfulDivider = (props: ContentfulDividerProps) => {
-  const { className } = props;
+export const ContentfulDivider = ({ className = '', ...props }: ContentfulDividerProps) => {
   return (
     <div className="cf-divider" {...props}>
-      <hr className={cx(styles.hr, className)} />
+      <hr className={className} />
     </div>
   );
 };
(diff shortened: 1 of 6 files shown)


packages/components/src/components/Divider/ContentfulDivider.tsx
```diff
@@ -1,22 +1,14 @@
...
-  }),
-};
 
 export type ContentfulDividerProps = {
   className?: string;
 };
 
-export const ContentfulDivider = (props: ContentfulDividerProps) => {
-  const { className } = props;
+export const ContentfulDivider = ({ className = '', ...props }: ContentfulDividerProps) => {
```
primeinteger, line 8: This subtle change makes it so `className` will not be spread via `...props` to the `div.cf-divider` element

packages/components/src/components/Divider/ContentfulDivider.tsx
```diff
@@ -1,22 +1,14 @@
 import React from 'react';
 import './ContentfulDivider.css';
-import { css, cx } from 'emotion';
-
-const styles = {
-  hr: css({
-    border: 'none',
-  }),
-};
```
primeinteger, line 9: The components package doesn't have `emotion` installed. I moved this style to the component's css file instead.

ryunsong-contentful (approved):
LGTM - Could double check with Travis that 20px is the appropriate margin default, my money's that he says 24px

PR #785, opened 2024-10-08
fix: dnd between custom components and patterns [ALT-1230]

chasepoirier:
## Purpose
Fixes a bug where we couldn't drop or move a component between patterns or custom components.

Ticket: https://contentful.atlassian.net/browse/ALT-1230

Also fixes a styling bug with the blue placeholder so that it aligns properly in respect to flex direction of the parent element. 

Chore: cleanup husky configs for v10

https://github.com/user-attachments/assets/0be6624a-b764-4292-982e-4802f4bc2e59

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
 npx commitlint --edit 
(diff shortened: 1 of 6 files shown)


primeinteger (approved):
Looks good to me when testing locally!  I can drag-n-drop between any type of component or pattern on the canvas and I'm not seeing any new unexpected behaviors. 🚀

PR #786, opened 2024-10-08
fix: make pattern height take 100% of available space [ALT-1294]

sam-is-content:
## Purpose

Bug where even if user set a pattern to 100% height, the editor did not show it as such

https://contentful.atlassian.net/browse/ALT-1294

## Approach

patterns create both an assembly div and a child div of that for the assemblyBlock. AssemblyBlock has the styles the user selected for the pattern while assembly is mostly a wrapper and only had 100% width on it. If we add 100% height too then the user should see their styles as expected.

Screenshot 2024-10-08 at 10 54 42 AM

### DIFF
--- a/packages/visual-editor/src/components/DraggableBlock/styles.module.css
+++ b/packages/visual-editor/src/components/DraggableBlock/styles.module.css
@@ -28,6 +28,7 @@
 
 .Dropzone.isAssembly {
   width: 100%;
+  height: 100%;
 }
 
 .DraggableComponent:is(.Dropzone).isDragging {


packages/visual-editor/src/components/DraggableBlock/styles.module.css
```diff
@@ -28,6 +28,7 @@
 
 .Dropzone.isAssembly {
   width: 100%;
+  height: 100%;
```
primeinteger, line 31: I can confirm this fixes the specific issue on the marketing demo site where this issue was identified:

The testimonial card on the right now fills the full height of the column, even though there is less content on the right card. 👍 

Screenshot 2024-10-08 at 1 58 24 PM

Screenshot 2024-10-08 at 1 58 38 PM

primeinteger (approved):
Clean 👏

PR #787, opened 2024-10-09
chore: update example next apps to show importing studio config properly [ALT-1315]

elylucas:
## Purpose

Updates Next examples to show how to properly import studio config client and server side.


primeinteger (approved):

ryunsong-contentful (approved):
I like this idea in general of injecting example usages. One thing I might suggest adding is a link to our docs for each section for component registration, DTs, and custom breakpoints in the comment sections. That way the user can look at our example and quickly reference our docs from the code as well.

PR #796, opened 2024-10-15
fix: update display name for text variable in the Heading component []

primeinteger:
## Purpose

This is a small change that has annoyed me for a long time.

The real purpose for making this change is I want to kick off a fresh deployment for the react-vite test app in vercel to see if it will update the SDK for this hosted app running in prod.

### DIFF
--- a/packages/test-apps/react-vite/package.json
+++ b/packages/test-apps/react-vite/package.json
@@ -8,7 +8,6 @@
     "start": "vite",
     "prebuild": "rimraf dist",
     "build": "tsc && vite build",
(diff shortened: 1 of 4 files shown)


packages/test-apps/react-vite/package.json
```diff
@@ -8,7 +8,6 @@
     "start": "vite",
     "prebuild": "rimraf dist",
     "build": "tsc && vite build",
-    "build:vercel": "npm run build --prefix ../core && npm run build --prefix ../components && npm run build",
```
primeinteger, line 11: This doesn't get used anymore. It looks like this was part of the old vercel build config before some stuff moved to github actions

PR #800, opened 2024-10-21
feat: new component registration option for editorWrapperWidth [ALT-1358]

primeinteger:
## Purpose

Adds a new `editorWrapperWidth` option to the component registration that can be used to set a `width` style on the DND wrapping div in editor mode.

This new option can be useful for custom components that do not use the `cfWidth` variable to allow the component width to be changed in studio.

For custom components that have a relative width using css or styling in the component code, the `editorWrapperWidth` option can be used to set the `width` of the DND wrapping div when selecting the component in studio.

If the custom component has a relative width such as `100%`, you can set `editorWrapperWidth: '100%'` in the registration options to make sure the component will have the correct width in editor mode.

## Demo

I created a small test component named Colorful Box, with height/width/background styles baked in with css, to demonstrate that setting `editorWrapperWidth: '50%'` allows the component to render correctly 50% wide and matches between editor and delivery modes.

https://github.com/user-attachments/assets/885a4f99-82a5-4f36-b493-74ea2230d2a2


ryunsong-contentful (approved):
Worked for me using your Colorful box component. As for the long-term solution I feel like this could be a short-term fix to alleviate the immediate customer issue while we figure out a longer term permanent solution. Like what we're doing now is bandaging on top of bandaging as issues come up.

PR #809, opened 2024-10-24
fix: handle leaking dragProps []

Chaoste:
## Purpose

Screenshot 2024-10-24 at 13 34 48

We're getting this warning when rendering dividers in the canvas. This is because, we're not using `withComponentWrapper`. I'm not sure why but that might be something that @ryunsong-contentful could answer as I don't have context on the creation of the divider.

### DIFF
--- a/packages/components/src/components/Button/Button.tsx
+++ b/packages/components/src/components/Button/Button.tsx
@@ -92,6 +92,7 @@ export const Button: React.FC<ButtonProps> = ({
     </button>
   );
 
+  // FIXME: Wrapping <button> with <a> is not ARIA conform. Instead, <a> should look like a button.
   return url ? (
     <a href={url} target={target}>
       {button}
--- a/packages/components/src/components/Divider/ContentfulDivider.tsx
+++ b/packages/components/src/components/Divider/ContentfulDivider.tsx
@@ -3,9 +3,15 @@ import './ContentfulDivider.css';
 
 export type ContentfulDividerProps = {
   className?: string;
+  dragProps?: unknown;
 };
 
-export const ContentfulDivider = ({ className = '', ...props }: ContentfulDividerProps) => {
+export const ContentfulDivider = ({
+  className = '',
+  // We have to exclude this explicitly from rendering as withComponentWrapper is not used
+  dragProps: _,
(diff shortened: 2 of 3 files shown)


SofiaMargariti (approved):

elylucas:
This feels like a bandaid for a larger issue. Wouldn't any component that has wrapComponent disabled be affected by this issue? I wouldn't want this to become a pattern that users would have to implement in their own components

Theres been other instances where our props have been applied to dom elements, we usually have to pull them out earlier in the process. I think looking into where the props are being applied in the visual editor and extracting `dragProps` if the node is a dom element would be a more eloquent solution.

Chaoste:
> This feels like a bandaid for a larger issue. Wouldn't any component that has wrapComponent disabled be affected by this issue? I wouldn't want this to become a pattern that users would have to implement in their own components
> 
> Theres been other instances where our props have been applied to dom elements, we usually have to pull them out earlier in the process. I think looking into where the props are being applied in the visual editor and extracting `dragProps` if the node is a dom element would be a more eloquent solution.

I looked into this and found out that we always exclude the `dragProps` [here](https://github.com/contentful/experience-builder/blob/7d3efcddaf50772ef653deadb31fbafae4e1222e/packages/experience-builder-sdk/src/utils/withComponentWrapper.tsx#L34) and that this is not about the `wrapComponent` flag at all. No matter which value it has, we will exclude such internal props when wrapping the registered component with `enrichComponentDefinition`. We do this for built-in as well as custom ones.

There are only five exceptions in `DEFAULT_COMPONENT_REGISTRATIONS` ([here](https://github.com/contentful/experience-builder/blob/27effa0f2caa6dc0dcb88591aa3833c5e0e1e1be/packages/experience-builder-sdk/src/core/componentRegistry.ts#L87-L88)): `container, section, columns, singleColumn, divider`
Those are not wrapped at all. As this is all built by Altitude, I couldn't understood the rational behind this and also couldn't find a component explaining why that is.
That's why I decided to fix this specific bug right away instead of refactoring this code as I'm missing the required context for that. If you can help me understand those exception, I'd be happy to adjust the PR and leave an explanatory note in the code.

PR #810, opened 2024-10-28
fix: improve performance of styles generation []

Chaoste:
## Purpose

When [this SDK error](https://contentful.sentry.io/issues/6019374985/?alert_rule_id=15048860&alert_type=issue&notification_uuid=5d6c2561-db7c-47ce-a2d7-ef5ebaf293b5&project=8049&referrer=slack) popped up today, I started looking into the code where this error originated from. It seems like, we recreate the styles too often.

## Approach

In `useComponentProps` we re-create the styles map `componentStyles` on every render which in turn will add them to the DOM again and again on every render. I added a `useMemo` to only regenerate the styles map when necessary and there save us from changing the DOM styles on every render.

Additionally, I improved the logic `useEditorModeClassName` to clean up old styles if they are not used anymore. This is only relevant to nodes without IDs which theoretically shouldn't happen but I wanted make the hook save against this case, so I added a `removeStylesTag` logic.

### DIFF
--- a/packages/visual-editor/src/hooks/useEditorModeClassName.ts
+++ b/packages/visual-editor/src/hooks/useEditorModeClassName.ts
@@ -23,22 +23,35 @@ export const useEditorModeClassName = ({
       return;
     }
 
-    const [className, styleRule] = buildStyleTag({ styles, nodeId });
+    const [newClassName, styleRules] = buildStyleTag({ styles, nodeId });
+    addStylesTag(newClassName, styleRules);
 
-    setClassName(className);
-
-    const existingTag = document.querySelector(`[data-cf-styles="${className}"]`);
-
-    if (existingTag) {
-      existingTag.innerHTML = styleRule;
-      return;
+    if (className !== newClassName) {
+      setClassName(newClassName);
+      // Clean up: remove outdated styles from DOM
+      removeStylesTag(className);
     }
-
-    const styleTag = document.createElement('style');
-    styleTag.dataset['cfStyles'] = className;
-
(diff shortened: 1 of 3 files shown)


packages/visual-editor/src/hooks/useEditorModeClassName.ts
```diff
@@ -23,22 +23,35 @@ export const useEditorModeClassName = ({
...
-    if (existingTag) {
-      existingTag.innerHTML = styleRule;
-      return;
-    }
+const removeStylesTag = (className: string) => {
+  const existingTag = document.querySelector(`[data-cf-styles="${className}"]`);
+  if (existingTag) {
+    document.head.removeChild(existingTag);
+  }
+};
```
primeinteger, line 44: I added some local debug logging to this method and never observed it finding an existing tag to remove. I could only see this function getting called with an empty string for the className (from the initial state definition).

Do you know of a scenario that would trigger this function to remove an existing style tag from the dom? Is there a test that could be written to verify the behavior?

primeinteger (approved):
I tested this locally with the react-vite and nextjs test apps and everything looks good to me.  All of the design property changes are rendering in the canvas as expected.

packages/visual-editor/src/hooks/useEditorModeClassName.ts
Chaoste: You're right that this scenario currently is not occurring as every node has an id. I pointed that out in the PR description as well because I only noticed it late but continued to make the function "saver".

PR #812, opened 2024-10-28
poc: refactor sdk to use dnd from user_interface [ALT-1365]

chasepoirier:
## Purpose
Stand up a POC to test moving dnd to the user_interface

Ticket: https://contentful.atlassian.net/browse/ALT-1365


Chaoste:
We're following up on this in a new separate PR as part of SPA-2745

PR #813, opened 2024-10-28
feat: enable recursive pattern rendering [SPA-2434]

Spring3:
## Purpose

Allow to render patterns recursively. Support has been added for the editor mode only.

## Approach

1. **For now only works in editor mode**
2. **Binding, design values over-writes are not functioning properly**


Chaoste (approved):
Changes in this PR are looking good. Will do testing & feedback on that in the other PR

PR #818, opened 2024-10-29
fix: change container/columns default Margin value [ALT-1388]

primeinteger:
## Purpose

Changes the Container and Columns default margin value back to `0 0 0 0` and added a comment to mention that the UI overrides the value to `0 Auto 0 Auto` when these components are added at the root/top-level

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -343,7 +343,8 @@ export const containerBuiltInStyles: VariableDefinitions = {
     type: 'Text',
     group: 'style',
     description: 'The margin of the container',
-    defaultValue: '0 auto 0 auto',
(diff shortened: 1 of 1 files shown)


elylucas (approved):

primeinteger:
> Why are we doing this? And why is the UI not just applying this defaultValue? I feel like there is a good reason but I can't remember

@ryunsong-contentful, When you drop a new Columns or Container component on the canvas, the `ComponentAddedAction` in the UI calls `applyNodeModifications()` and if the component is dropped on the root level it will override the `defaultValue` from the component definition with `0 Auto 0 Auto` to set left/right margin auto.

We need the `defaultValue` to be `0 0 0 0` for these two components here so that when that override doesn't occur on the UI side (such as dropping these two components into a nested structure instead of the root level) it will not add the left/right margin auto.

PR #819, opened 2024-10-30
feat: populate pattern nodes on initialisation [SPA-2462]

ChidinmaOrajiaku:
## Purpose
For Editor mode, populate pattern nodes on initialisation web app side.

## Approach

This PR removes resolving assembly in Editor mode. The editor now sends the deserialised pattern nodes instead.


Spring3 (approved):
👍 works well from what I've tested

Chaoste (approved):
Looks reasonable to me :)
I will do the testing & commenting on the other PR as that is the crucial part.

Just one note: We have [this fixture](https://github.com/contentful/experience-builder/blob/bb5082266588f28efef8372a89c36dfbf695ef4e/packages/visual-editor/test/__fixtures__/assembly.ts#L121) that is now not used anymore and even generates an example node of "the old way". I'd suggest either updating to match reality to avoid misusage for the next developer or just removing it.

PR #824, opened 2024-11-04
feat: resolve entries that are embedded into bound entries [ALT-1368]

elylucas:
## Purpose

We have several instances where if we bind an entry to a component variable, any referenced entities in that entry are not resolved, and therefore not available as a passed in prop to the component.

Instances where this can occur:
An entry bound into a rich text field
An entry bound to another entry that itself has references to another entry
A list of entries/assets bound to an array variable type where some of which could have references as well

Pretty much any case where a bound entity would have a reference to another entry.

## Approach

Initially, this was attempted to be solved by removing the `withoutLinkResolution` call to the fetch entities API. This automatically resolved any referenced entries no matter how they were referenced. However, this caused issues elsewhere, so we had to do this manually in the transform methods for each specific variable type that could have referenced entries (Array, Link, and RichText).

### DIFF
--- a/packages/core/src/fetchers/fetchAllEntities.ts
+++ b/packages/core/src/fetchers/fetchAllEntities.ts
@@ -43,7 +43,7 @@ export const fetchAllEntries = async ({
       };
     }
 
-    const query = { 'sys.id[in]': ids, locale, limit, skip };
+    const query = { 'sys.id[in]': ids, locale, limit, skip, include: 2 as 2 | undefined };
 
     const {
       items,
--- a/packages/core/src/utils/transformers/getArrayValue.ts
+++ b/packages/core/src/utils/transformers/getArrayValue.ts
@@ -26,7 +26,24 @@ export function getArrayValue(
     if (typeof value === 'string') {
       return value;
     } else if (value?.sys?.type === 'Link') {
-      return entityStore.getEntityFromLink(value);
+      const resolvedEntity = entityStore.getEntityFromLink(value);
+      if (!resolvedEntity) {
+        return;
+      }
+      //resolve any embedded links - we currently only support 2 levels deep
+      const fieldKeys = Object.keys(resolvedEntity.fields || {});
+      fieldKeys.forEach((fieldKey) => {
+        const field = resolvedEntity.fields[fieldKey];
+        if (field && field.sys?.type === 'Link') {
+          const entity = entityStore.getEntityFromLink(field);
+          if (entity) {
+            resolvedEntity.fields[fieldKey] = entity;
+          }
+        }
+      });
+
+      return resolvedEntity;
+      // return entityStore.getEntityFromLink(value);
     } else {
       console.warn(`Expected value to be a string or Link, but got: ${JSON.stringify(value)}`);
       return undefined;
--- a/packages/core/src/utils/transformers/getArrayValue.ts
+++ b/packages/core/src/utils/transformers/getArrayValue.ts
@@ -31,9 +31,8 @@ export function getArrayValue(
         return;
       }
       //resolve any embedded links - we currently only support 2 levels deep
-      const fieldKeys = Object.keys(resolvedEntity.fields || {});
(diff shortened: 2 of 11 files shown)


packages/core/src/fetchers/fetchAllEntities.ts
```diff
@@ -43,7 +43,7 @@ export const fetchAllEntries = async ({
       };
     }
 
-    const query = { 'sys.id[in]': ids, locale, limit, skip };
+    const query = { 'sys.id[in]': ids, locale, limit, skip, include: 2 as 2 | undefined };
```
primeinteger, line 46: What do you think about using `2 as const` here instead?
```suggestion
    const query = { 'sys.id[in]': ids, locale, limit, skip, include: 2 as const };
```

packages/core/src/utils/transformers/getArrayValue.ts
```diff
@@ -26,7 +26,24 @@ export function getArrayValue(
...
     } else if (value?.sys?.type === 'Link') {
-      return entityStore.getEntityFromLink(value);
+      const resolvedEntity = entityStore.getEntityFromLink(value);
+      if (!resolvedEntity) {
+        return;
+      }
+      //resolve any embedded links - we currently only support 2 levels deep
+      const fieldKeys = Object.keys(resolvedEntity.fields || {});
+      fieldKeys.forEach((fieldKey) => {
+        const field = resolvedEntity.fields[fieldKey];
```
primeinteger, line 36: Using `Object.entries` might be nice here since the loop is using both the `fieldKey` and `field`

```suggestion
      const fields = resolvedEntity.fields || {};
      Object.entries(fields).forEach(([fieldKey, field]) => {
```

packages/core/src/utils/transformers/getResolvedEntryFromLink.ts
```diff
@@ -20,5 +20,32 @@ export function getResolvedEntryFromLink(
...
   const resolvedEntity = entityStore.getEntityFromLink(value);
+
+  if (!resolvedEntity) {
+    return;
+  }
+
+  //resolve any embedded links - we currently only support 2 levels deep
+  const fieldKeys = Object.keys(resolvedEntity.fields || {});
+  fieldKeys.forEach((fieldKey) => {
+    const field = resolvedEntity.fields[fieldKey];
```
primeinteger, line 31: Using `Object.entries` might be nice here too (same as the comment above)

packages/core/src/utils/transformers/getArrayValue.ts
```diff
@@ -26,7 +26,24 @@ export function getArrayValue(
...
+        if (field && field.sys?.type === 'Link') {
+          const entity = entityStore.getEntityFromLink(field);
+          if (entity) {
+            resolvedEntity.fields[fieldKey] = entity;
+          }
+        }
+      });
+
+      return resolvedEntity;
+      // return entityStore.getEntityFromLink(value);
```
primeinteger, line 46: Should the old line of code be removed?

packages/core/src/utils/transformers/transformBoundContentValue.spec.ts
```diff
@@ -197,5 +233,135 @@ describe('transformBoundContentValue', () => {
...
+          },
+        ],
+        data: {},
+        nodeType: 'document',
+      });
+    });
+  });
+
+  describe('when the variable type is a "Link"', () => {
+    it('it should resolve an entry that has a reference to another entry', () => {
```
primeinteger, line 311: Great updates to the test suite! 🙌

primeinteger (approved):
Great work on this! 🔥

packages/core/src/fetchers/fetchAllEntities.ts
elylucas: Since the 'include' prop on the query param used a numeric literal, this is what I did to make TS happy as it wouldn't just let me pass in a number to it.

PR #825, opened 2024-11-06
fix: drag-n-drop error when moving custom components [ALT-1446]

primeinteger:
## Purpose

Fix error `Invariant failed: Cannot finish a drop animating when no drop is occurring`

This issue first started in version `1.17.2-dev-20241007T1903-dddb811.0` which was released into `1.18.0`

This issue was due to the changes to `useComponent` in https://github.com/contentful/experience-builder/pull/784 which unintentionally made it so the component drag-n-drop wrapper was disabled by default, instead of enabled by default.

https://github.com/user-attachments/assets/ce9f6d2d-0033-4eac-8b2d-dee1a6e57ab2


chasepoirier (approved):
Looks good!

PR #830, opened 2024-11-07
fix: don't add client object to fetchBy dependency arrays to fix infinite rerenders [ALT-1447]

elylucas:
## Purpose

https://contentful.atlassian.net/browse/ALT-1447

## Approach

Removed the client from useEffect dep arrays so it wouldn't cause issues changing. This should be a safe change since the client doesn't typically change during the request.


ryunsong-contentful (approved):
Nice

PR #833, opened 2024-11-08
poc: add support for custom RSC in preview/delivery mode

chasepoirier:
## Purpose
Explore what it would take to fully support customer components that are React Server Components in Studio.

## Approach
This approach extends off of the work to make our [ExperienceRoot a server component](https://contentful.atlassian.net/browse/ALT-1428) so that we can understand where limitations will be and what work needs to be done next for the overall initiative.

Findings are as follows:
- The entityStore needs to be used by both client and server components, but we can only pass plain JS objects between client and server components, not functions or classes. We'll need a way to deserialized/serialize the entityStore so it can be used by both sides.
- When a custom defines components, that component definition will need to be supplied to both the client and server side of our SDK. For example, if a custom defines their custom components in a server component, the componentRegistry will only know about the custom components on the server. We need a way to have the componentRegistry be available on both client and server. This runs into similar caveats above where we can't pass customer components (functions)
- We need to update the build script of our VisualEditor pkg if we want to make at least the VisualEditorRoot a server component. Right now the build script bundles all the code into one file so client code is mixed with server code.
- We'll need a way to pre-render server components from the `PreviewDeliveryRoot.tsx` and `VisualEditorRoot.tsx` so that they can be passed as props to the client side `CompositionBlock.tsx` components. This is where having the component definition option of `.isRSC: true` comes into play so server components are pre-rendered.


Chaoste:

PR #834, opened 2024-11-08
feat: remove 'use client' directive from ExperienceRoot [ALT-1428]

chasepoirier:
## Purpose
Completes the first step to allowing RSCs through Studio by removing the 'use client' directive from the ExperienceRoot component.

Ticket: https://contentful.atlassian.net/browse/ALT-1428

## Approach
Uses the pattern of ServerComponents as props of ClientComponents to allow the rendering of server components nested within client components.

https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#supported-pattern-passing-server-components-to-client-components-as-props


Chaoste:

PR #836, opened 2024-11-12
[ALT-1452] refactor: components now contain the definition to provide to the core package

ryunsong-contentful:
## Purpose
Refactor code - move all the component definitions into their proper directory with their required definitions


Chaoste (approved):
Nice! :)

primeinteger (approved):
I like the consistency with the basic components. Clean work 👏

PR #838, opened 2024-11-19
fix: react on changes of breakpoint when tree is the same [SPA-2487]

Chaoste:
## Purpose
When testing initialisation with custom breakpoints for [this fix PR](https://github.com/contentful/user_interface/pull/24124#issuecomment-2486324029), I realised that the diffing logic will ignore breakpoint changes if the tree didn't change.

## Approach
Even if the tree is empty (and thus an empty tree diff), we accept a change of the breakpoint settings.

### DIFF
--- a/packages/visual-editor/src/store/tree.ts
+++ b/packages/visual-editor/src/store/tree.ts
@@ -41,10 +41,6 @@ export interface TreeStore {
   ) => void;
 }
 
-const isAssemblyNode = (node: ExperienceTreeNode) => {
-  return node.type === ASSEMBLY_NODE_TYPE;
-};
-
 export const useTreeStore = create<TreeStore>((set, get) => ({
   tree: {
     root: {
@@ -109,15 +105,7 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
      * rerendering the entire tree.
      */
     const treeDiff = getTreeDiffs({ ...currentTree.root }, { ...tree.root }, currentTree);
-
-    const currentBreakpoints = (currentTree?.root?.data?.breakpoints ?? []).slice().sort();
-    const newBreakpoints = (tree?.root?.data?.breakpoints ?? []).slice().sort();
-    let didBreakpointsChange = false;
-    if (newBreakpoints.length && newBreakpoints.length !== currentBreakpoints.length) {
-      didBreakpointsChange = newBreakpoints.some(
-        (value, index) => currentBreakpoints[index] !== value,
-      );
-    }
+    const didBreakpointsChange = hasBreakpointDiffs(currentTree, tree);
 
     // The current and updated tree are the same, no tree update required.
     // Special case: Breakpoints changed (e.g. empty experience gets reloaded or breakpoints updated)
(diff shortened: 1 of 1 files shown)


packages/visual-editor/src/store/tree.ts
```diff
@@ -110,8 +110,18 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
      */
     const treeDiff = getTreeDiffs({ ...currentTree.root }, { ...tree.root }, currentTree);
 
+    const currentBreakpoints = (currentTree?.root?.data?.breakpoints ?? []).slice().sort();
```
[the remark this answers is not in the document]
Chaoste: Thanks, that's a good idea. As it's only used in here I added to the other util functions in this file at the bottom. If we would need this again, I'm happy to move it over to `@/utils`. Right now, it's a bit unorganized as we have quite a lot of separate util functions in there.

packages/visual-editor/src/store/tree.ts
```diff
@@ -109,9 +105,11 @@ export const useTreeStore = create<TreeStore>((set, get) => ({
      * rerendering the entire tree.
      */
     const treeDiff = getTreeDiffs({ ...currentTree.root }, { ...tree.root }, currentTree);
+    const didBreakpointsChange = hasBreakpointDiffs(currentTree, tree);
```
Spring3, line 108: Should we instead feed the breakpoints array? Not sure if we need to pass the entire trees 😓

Spring3 (changes requested):
Current equality check doesn't account for 2 scenarios:
* the check only happens if the length of the new and old arrays is not equal, which is not going to always be the case
* Given that a breakpoint is an object, `===` comparison won't work

```
[
    {
        "id": "test-desktop",
        "query": "*",
        "displayName": "All Sizes",
        "displayIcon": "desktop",
        "previewSize": "100%"
    },
    {
        "id": "test-tablet",
        "query": "<982px",
        "displayName": "Tablet",
        "displayIcon": "tablet",
        "previewSize": "820px"
    },
    {
        "id": "test-mobile",
        "query": "<576px",
        "displayName": "Mobile",
        "displayIcon": "mobile",
        "previewSize": "350px"
    }
]
```

Maybe let's just use lodash.isEqual?

packages/visual-editor/src/store/tree.ts
Chaoste: Saving a few lines within a quite large method was driving my code style, even though it's not nice to give the method access to the whole tree. Somehow, I didn't like to write the following:
```
const didBreakpointsChange = hasBreakpointDiffs(
  currentTree?.root?.data?.breakpoints ?? [],
  tree?.root?.data?.breakpoints ?? []
);
```
Do you have something else in mind, especially as we're only using lodash now within the method?

Chaoste:
> Current equality check doesn't account for 2 scenarios:
> [...]
> Maybe let's just use lodash.isEqual?

Thanks for calling that out! I don't know why I committed such a faulty code - probably just copy and paste from SO without thinking about it. As we're already using lodash, it's actually a very nice idea to use that well established and tested logic :)

SofiaMargariti (approved):
Thanks for creating the separate method.

Is there any chance we could add a test for this? Not gonna block as this file doesn't have any tests already, but I'd say it's generally a good practice even for small changes to have tests as in the future the logic can evolve and become more complex.

SofiaMargariti:
> > Thanks for creating the separate method.
> > Is there any chance we could add a test for this? Not gonna block as this file doesn't have any tests already, but I'd say it's generally a good practice even for small changes to have tests as in the future the logic can evolve and become more complex.
> 
> Actually wanted to add that but then figured that we shouldn't test logic from external libraries. Lodash has their own tests for `isEqual` so there is not much left that we could actually test besides that

True, with the simplified logic testing just the `hasBreakpointDiffs` probably makes no sense. Testing the `updateTree` and how it behaves when there are breakpoint diffs is more meaningful. But in any case this is not a blocker from my side.

Spring3 (approved):
👍

PR #840, opened 2024-11-19
fix: small clean up in ExperienceRoot

sam-is-content:


primeinteger (approved):
Thanks for doing that! 👏

PR #843, opened 2024-11-21
feat: add experimental components to define components

sam-is-content:
## Purpose
add an experimental components option to options in define components
https://contentful.atlassian.net/jira/software/c/projects/ALT/boards/1552?selectedIssue=ALT-1413

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -119,7 +119,9 @@ export type ComponentRegistration = {
 
 export type ComponentRegistrationOptions = {
   enabledBuiltInComponents?: string[];
-  experimentalComponents?: Record<string, boolean>;
+  experimentalComponents?: {
+    carousel?: boolean;
+  };
 };
 
 export type Binding = {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -119,6 +119,7 @@ export type ComponentRegistration = {
 
 export type ComponentRegistrationOptions = {
   enabledBuiltInComponents?: string[];
+  experimentalComponents?: Record<string, boolean>;
 };
 
 export type Binding = {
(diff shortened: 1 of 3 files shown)


packages/test-apps/react-vite/src/studio-config.ts
```diff
@@ -6,181 +6,188 @@ import NestedSlots from './components/NestedSlots';
...
+      },
     },
-    options: {
-      wrapComponent: false,
+  ],
+  {
+    experimentalComponents: {
+      carousel: true,
     },
   },
```
primeinteger, line 189: Could you also update the `ComponentDefinitionOptions` type to include this new option?

[packages/core/src/types.ts#L120-L122](https://github.com/contentful/experience-builder/blob/development/packages/core/src/types.ts#L120-L122)

Then TS will know about this new option when it's implemented in the `defineComponents` function.

primeinteger (commented):
Adding this as an option to `defineComponents` works for me!

This new option could also be added to the nextjs test app so that we can use it in SSR land as well.

packages/test-apps/react-vite/src/studio-config.ts
sam-is-content: yeah i tried to before but git didnt seem to pick up on the change 🤔

packages/core/src/types.ts
```diff
@@ -119,6 +119,7 @@ export type ComponentRegistration = {
 
 export type ComponentRegistrationOptions = {
   enabledBuiltInComponents?: string[];
+  experimentalComponents?: Record<string, boolean>;
```
primeinteger, line 122: I wonder if it would be better to define the object explicitly since it only has 1 expected property 🤔 
```suggestion
  experimentalComponents?: {
    carousel?: boolean;
  };
```

primeinteger (approved):
🔥

elylucas (approved):

PR #847, opened 2024-11-23
feat(nextjs-marketing-demo): show extra footer info for debugging

primeinteger:
## Purpose

Exports the `version` from the SDK package.

Updates the marketing demo template so that you can add `debug` to the URL query string to get additional debug info in the footer (displays the version of the Experiences SDK and React).

Screenshot 2024-11-22 at 7 21 51 PM

### DIFF
--- a/packages/templates/nextjs-marketing-demo/src/components/Footer/FooterDebugging.tsx
+++ b/packages/templates/nextjs-marketing-demo/src/components/Footer/FooterDebugging.tsx
@@ -1,0 +1,27 @@
+'use client';
+
+import React from 'react';
+import { useSearchParams } from 'next/navigation';
+import { version as sdkVersion } from '@contentful/experiences-sdk-react';
+import { version as reactVersion } from 'react/package.json';
+import { Divider, Flex, Typography } from 'antd';
+import styles from './styles.module.css';
+
+const FooterDebugging: React.FC = () => {
+  const searchParams = useSearchParams();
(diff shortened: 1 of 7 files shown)


packages/templates/nextjs-marketing-demo/src/components/Footer/FooterDebugging.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useSearchParams } from 'next/navigation';
+import { version as sdkVersion } from '@contentful/experiences-sdk-react';
+import { Divider, Flex, Typography } from 'antd';
+import styles from './styles.module.css';
+
+const FooterDebugging: React.FC = () => {
+  const searchParams = useSearchParams();
+  const isDebugMode = searchParams.has('debug');
+
+  if (!isDebugMode) {
```
elylucas, line 14: wonder if we can check for the dev env var here instead of passing in a query param

packages/templates/nextjs-marketing-demo/src/components/Footer/FooterDebugging.tsx
```diff
@@ -0,0 +1,27 @@
+'use client';
```
elylucas, line 1: i think this could probably be a server component

elylucas (changes requested):
Good idea, requested a couple of small tweaks

PR #850, opened 2024-12-02
feat: update types for ComponentDefinition variables [ALT-690]

primeinteger:
## Purpose

This branch updates the types for `ComponentDefinition.variables` to map out the data type for each built-in design variable.

## Approach

Previously when accessing a component definition variable, the type could be a long list of possibilities, like this:

Screenshot 2024-12-02 at 3 58 09 PM

Developers would then use type casting to make the type more specific and accurate in the UI, like this:
```ts
cfImageOptions as ComponentDefinitionVariable<'Object'> | undefined
```

After these changes, the type casting is no longer needed for the built-in variables because the types will infer correctly:

Screenshot 2024-12-02 at 3 25 45 PM

## Comparison

These changes will allow us to remove the type casting for component definition variables from `<SectionStyles>` and a few other components in the UI, along the lines of:

Screenshot 2024-12-02 at 3 51 43 PM

### DIFF
--- a/packages/core/src/definitions/styles.ts
+++ b/packages/core/src/definitions/styles.ts
@@ -1,15 +1,8 @@
 import { DEFAULT_IMAGE_WIDTH } from '@/constants';
-import { ComponentDefinitionVariable, ContainerStyleVariableName } from '../types';
-
-type VariableDefinitions = Partial<
-  Record<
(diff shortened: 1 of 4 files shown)


ryunsong-contentful (approved):
noice

PR #851, opened 2024-12-03
[ALT-1459] feat: add is in experiences editor mode boolean to component when dropped onto the canvas

ryunsong-contentful:
## Purpose
Pass in isExpEditorMode as a prop to rendered components

https://github.com/user-attachments/assets/7b8739cf-5302-45c7-bf4e-b81f2e7d5228

### DIFF
--- a/packages/visual-editor/src/hooks/useComponent.tsx
+++ b/packages/visual-editor/src/hooks/useComponent.tsx
@@ -86,7 +86,7 @@ export const useComponent = ({
     const { dragProps = {} } = props || {};
 
     const {
-      editorMode: _editorMode,
+      editorMode: isInExpEditorMode,
       renderDropzone: _renderDropzone,
       node: _node,
       ...customComponentProps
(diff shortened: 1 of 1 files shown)


primeinteger (changes requested):
I tried to test this but I don't see this prop arriving to the built-in Button component or the custom Component With Children component in the canvas.

I also noticed your code adds the prop as `isInExpEditorMode` but your demonstration video shows you reading the prop as `editorMode` - but I tried using both props and I don't see it arriving to the component with a value either way.

primeinteger (approved):
Pardon my skill issue, the boolean was just getting stripped out because I was rendering it in the component output as a boolean instead of converting it to a string 🤦 

I tested built-in and custom components and I see the new `isInExpEditorMode` prop coming through as expected, only in editor mode, and it doesn't "pollute" the DOM with any new HTML attributes. Looks good! 👏

PR #859, opened 2024-12-04
feat: add optional id property for component tree nodes

jsdalton:
## Purpose

A few upcoming features (e.g. diffing between the component tree for version history) would benefit from a stable UUID-style identifier for tree nodes that's persisted and (relatively) stable throughout the lifetime of that node.

## Approach

* Add a new `id` property to the validator for the component tree node schema, on the current schema version

For now we decided against creating a new schema version to introduce this property. This can come later, once the implementation is stabilized and we introduce code that ensures that old versions of experiences (ie that use the current schema version) can be seamlessly upgraded to the new version correctly.


Chaoste (approved):
Luv iiiiit

SofiaMargariti (approved):

PR #862, opened 2024-12-06
fix: revert type for validation option values [ALT-690]

primeinteger:
## Purpose

Reverting changes to the value type in `ValidationOption` (the validation rules of component definition variables)

I tried to make this type smarter and infer the right type based on the variable type, but it's actually not a good idea to do this because the validation rules are only expected for certain variable types as simple string or number values in the UI side, so it's best to leave this type as it was before.

Eventually we could expand the support for validations on other variable types and that would be a better time to revise this type.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2507,7 +2507,6 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
-      "extraneous": true,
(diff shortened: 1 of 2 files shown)


ryunsong-contentful (approved):
Gross

elylucas (approved):

Chaoste (approved):

PR #870, opened 2024-12-11
feat: changes required for nested patterns rendering and styles [SPA-2475]

anwaar931:
## Purpose
We have recently introduced nested patterns. In order to render them in preview/delivery mode we need to make some changes to our existing code. 
This PR mostly focuses on populating the used Components array with the nested patterns.


Chaoste (approved):
Tested both CSR and SSR. Works perfectly 👍

PR #872, opened 2024-12-12
fix(visual-editor): prevent hover tooltip for selected component [ALT-1090]

primeinteger:
## Purpose

This change prevents the visual editor's hover tooltip from displaying when hovering on the selected component.


ryunsong-contentful:
Work as intended so approving. However, there are still some outliers that we can make future tickets for. For example maybe we want hover to be on top in cases like Columns vs Column. Although this is one of those issues where the juice might not be worth the squeeze since one's on the UI and other is presented via the SDK, and making the SDK hover label show over the UI selected component label would be a large code change for a small improvement. 

Screenshot 2024-12-12 at 12 54 59 PM

ryunsong-contentful (approved):
Tested manually and works as expected

PR #873, opened 2024-12-12
fix: hyperlinks to assets or entries should resolve in a rich text variable [ALT-1493]

elylucas:
## Purpose

[ALT-1493] - when hyperlink to an asset or entry in a rich text field and binding it to a rich text variable in studio, the entities where not being resolved.

## Approach

[ALT-1493]: https://contentful.atlassian.net/browse/ALT-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/test/__fixtures__/entities.ts
```diff
@@ -350,6 +351,131 @@ export const entryWithEmbeddedAssetInRichText: Entry = {
...
+            {
+              nodeType: 'list-item',
+              data: {},
+              content: [
+                {
+                  nodeType: 'asset-hyperlink',
+                  data: {
+                    target: {
+                      sys: {
+                        id: entityIds.ASSET1,
```
primeinteger, line 440: Should this have a different ID like `ASSET2` so it's not confused with the "embedded-asset-block"?

primeinteger (approved):
Great work on this! 

I tested a custom rich text component and can see with custom node handling we can display hyperlinks to other entries/assets quite easily. We can also add custom node handling and render a customized view for any embedded assets/entries 💙

PR #874, opened 2024-12-16
fix(experience-builder-sdk): SSR classNames not being attached to components rendered in SSR [ZEND-5755]

primeinteger:
## Purpose

Fixed an issue in `CompositionBlock` where `cfSsrClassName` was being set to `undefined` when `getPatternChildNodeClassName` was not defined. Added checks to ensure `getPatternChildNodeClassName` exists before calling it, ensuring SSR class names are applied when rendering components in preview/delivery.

For internal tracking: https://contentful.atlassian.net/browse/ZEND-5755

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
@@ -90,6 +90,7 @@ export const CompositionBlock = ({
       };
     }
 
+    // TODO: remove the deep tertiaries for better clarity and maintainability
     const propMap: Record<string, PrimitiveValue> = {
       cfSsrClassName:
         node.id && getPatternChildNodeClassName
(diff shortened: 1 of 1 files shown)


elylucas (approved):
Approving for now, but left a comment on code clarity.

ryunsong-contentful (approved):

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
```diff
@@ -220,7 +221,11 @@ export const CompositionBlock = ({
       ? node.children.map((childNode: ComponentTreeNode, index) => {
           return (
             <CompositionBlock
-              getPatternChildNodeClassName={_getPatternChildNodeClassName}
+              getPatternChildNodeClassName={
+                isAssembly || getPatternChildNodeClassName
```
elylucas, line 225: same deal here. This feels nonsensical and like its covering up deeper issues. Cool that it works for now, but we should refactor.

PR #882, opened 2024-12-19
fix: render patterns in preview [SPA-2518]

Chaoste:
## Purpose

As a regression of https://github.com/contentful/experience-builder/pull/874, the application would crash everytime when `cfSsrClassName` was not defined, i.e. when not running in SSR.

Screenshot 2024-12-19 at 09 12 27

## Approach
Simplify the ternary conditions and remove some duplicate conditional checks. As `resolveDesignValue` handles undefined values, I adjusted its types and was able to further reduce our conditional logic in `CompositionBlock`.


anwaar931 (approved):
:rocket:

PR #896, opened 2025-01-06
[ALT-1508] feat: add gatsby project to templates and add readmes for templates and test-apps

ryunsong-contentful:
## Purpose
Add gatsby to the examples directory so that it can be used for the Gatsby dev doc guide.

There's two examples, one for SPA client side Gatsby apps and the other for SSG. The SSG has an additional file `gatsby-node.js` to polyfill the path import in order to run the project locally.

## Approach
Started a new gatsby project by doing the following steps 
1. `npm init gatsby`

Integrate with Experiences
1. `npm i @contentful/experiences-sdk-react`
2. Created `examples/gatsby/src/utils/useContentfulClient.js` file to create the Contentful Client to hit the Contentful CDN
3. Created `examples/gatsby/src/pages/experiences-slug-example.js` which uses the Client to fetch the experience object and use the <ExperienceRoot> to render the Experience.


examples/gatsby/gatsby-spa/README.md
```diff
@@ -0,0 +1,76 @@
+<p align="center">
+  <a href="https://www.gatsbyjs.com/?utm_source=starter&utm_medium=readme&utm_campaign=minimal-starter-ts">
+    <img alt="Gatsby" src="https://www.gatsbyjs.com/Gatsby-Monogram.svg" width="60" />
+  </a>
+</p>
+<h1 align="center">
+  Gatsby Minimal TypeScript Starter
+</h1>
```
primeinteger, line 8: In my opinion, we could replace this fancy html header with a simple one like our other example apps
```suggestion
# Studio Experiences with Gatsby single page application example
```

examples/gatsby/gatsby-spa/README.md
```diff
@@ -0,0 +1,76 @@
...
+    - [Guides](https://www.gatsbyjs.com/docs/how-to/?utm_source=starter&utm_medium=readme&utm_campaign=minimal-starter-ts)
+    - [API Reference](https://www.gatsbyjs.com/docs/api-reference/?utm_source=starter&utm_medium=readme&utm_campaign=minimal-starter-ts)
+    - [Plugin Library](https://www.gatsbyjs.com/plugins?utm_source=starter&utm_medium=readme&utm_campaign=minimal-starter-ts)
+    - [Cheat Sheet](https://www.gatsbyjs.com/docs/cheat-sheet/?utm_source=starter&utm_medium=readme&utm_campaign=minimal-starter-ts)
+
+## 🚀 Quick start (Netlify)
+
+Deploy this starter with one click on [Netlify](https://app.netlify.com/signup):
+
+[<img src="https://www.netlify.com/img/deploy/button.svg" alt="Deploy to Netlify" />](https://app.netlify.com/start/deploy?repository=https://github.com/gatsbyjs/gatsby-starter-minimal-ts)
```
primeinteger, line 76: In my opinion, we could delete this.

examples/gatsby/gatsby-spa/src/pages/index.tsx
primeinteger, line 1: In my opinion, we don't need all this css and html content in our example app

primeinteger (commented):
Just a few comments so far before I start testing. Thanks for taking more time to switch it to TS and making it more consistent with other example apps! I only reviewed the SPA app but my opinion-based comments also apply to the SSG app

examples/gatsby/gatsby-spa/src/pages/ExperienceSlugExample.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { ExperienceRoot, useFetchBySlug } from '@contentful/experiences-sdk-react';
+import { useContentfulClient } from '../utils/useContentfulClient';
+
+const slug = ''
+const localeCode = '';
+const experienceTypeId = '';
```
primeinteger, line 7: I find it a little bit odd to hardcode the slug and locale to one value. Is that intentional because of being a single-page application?

We should use the `GATSBY_CTFL_EXPERIENCE_TYPE` environment variable for the `experienceTypeId`

examples/gatsby/gatsby-spa/README.md
```diff
@@ -0,0 +1,76 @@
...
+  GATSBY_CTFL_ENVIRONMENT=
+  GATSBY_CTFL_SPACE=
+  GATSBY_CTFL_ACCESS_TOKEN=
+  GATSBY_CTFL_PREVIEW_ACCESS_TOKEN=
+  GATSBY_CTFL_EXPERIENCE_TYPE=
+```
+
+3. **Create the Experience in the Contentful Experienecs UI**
+
+In the settings tab of Experiences, ensure that your Slug field matches the slug in step 1. Save your changes and publish to see this Gatsby project hooked up with Contentful Experiences.
```
primeinteger, line 35: These steps may need to mention what Content Preview URL to configure. I can't figure it out.  I tried:
* `http://localhost:8000` -> shows the gatsby index page (not studio-enabled)
* `http://localhost:8000/{entry.fields.slug}` -> shows the 404 page (even though I hardcoded the slug variable in ExperienceSlugExample.tsx to match the slug of my experience)
  Screenshot 2025-01-09 at 9 29 18 AM
  Screenshot 2025-01-09 at 9 28 57 AM

examples/gatsby/gatsby-spa/src/pages/404.tsx
```diff
@@ -0,0 +1,49 @@
...
+  fontSize: "1.25rem",
+  borderRadius: 4,
+}
+
+const NotFoundPage: React.FC<PageProps> = () => {
+  return (
+    <main style={pageStyles}>
+      <h1 style={headingStyles}>Page not found</h1>
+      <p style={paragraphStyles}>
+        Sorry 😔, we couldn’t find what you were looking for.
```
primeinteger, line 31: I noticed (as I'm struggling to get the example running) that I'm seeing a 404 page but the content doesn't match this file. Any idea where the 404 content below comes from?

Screenshot 2025-01-09 at 9 28 57 AM

examples/gatsby/gatsby-spa/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "develop": "gatsby develop",
+    "start": "gatsby develop",
+    "build": "gatsby build",
+    "serve": "gatsby serve",
+    "clean": "gatsby clean",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@contentful/experiences-sdk-react": "^1.27.1",
+    "gatsby": "^5.14.1",
```
elylucas, line 20: we should have a dep on the 'contentful' lib here as well as its a peer dep of the sdk.

examples/gatsby/gatsby-spa/src/pages/index.tsx
elylucas: The other example apps redirect the '/' route to '/home-page', and the routing is setup so the segment after the root is the slug that gets passed to fetch the experience.

examples/gatsby/gatsby-spa/src/pages/ExperienceSlugExample.tsx
elylucas: Ya, I think the slug should be dynamic based on the route (see my previous comment). 

For the locale, the other example apps are using built in i18n to detect a localeCode. Does gatsby have anything similar that we could use for providing the locale?

examples/gatsby/gatsby-spa/.env.local.example
```diff
@@ -0,0 +1,6 @@
+# Rename this file to .env.local and fill in the values for the following environment variables
```
elylucas, line 1: I had to name my file .env.development in order to get it to pick up the env vars, and that seems to be what Gatsby suggests when running the 'npm run develop' command. Curious if other found the same, and if so, we should update this filename and instructions to use .env.development instead.

examples/gatsby/gatsby-spa/src/utils/useContentfulClient.ts
```diff
@@ -0,0 +1,18 @@
...
+import { useMemo } from 'react';
+import { createClient } from 'contentful';
+
+export const useContentfulClient = () => {
+  const client = useMemo(() => {
+    const clientConfig = {
+      space: process.env.GATSBY_CTFL_SPACE || '',
+      environment: process.env.GATSBY_CTFL_ENVIRONMENT || '',
+      host: 'cdn.contentful.com',
+      accessToken: process.env.GATSBY_CTFL_ACCESS_TOKEN || '',
```
elylucas, line 11: probably be good to grab a `isPreview` param from the querystring and make the host dynamic based on if they are trying to pull up preview mode or delivery mode here.

examples/gatsby/gatsby-ssg/src/pages/ExperienceSlugExample.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const { nodes } = response.data.allContentfulExperience;
+
+  for (const node of nodes) {
+    const { slug, title, node_locale: localeCode, contentful_id, sys } = node;
+
+    const experienceEntry = await fetchById({
+        client,
+        experienceTypeId: sys.contentType.sys.id, // There was a missing comma in the docs, fix this typo in the docs
```
elylucas, line 39: can remove this comment

examples/gatsby/gatsby-ssg/src/pages/ExperienceSlugExample.tsx
```diff
@@ -0,0 +1,61 @@
+import path from 'path';
+import { fetchById, detachExperienceStyles } from '@contentful/experiences-sdk-react';
+import { useContentfulClient } from '../utils/useContentfulClient';
+
+exports.createPages = async ({ graphql, actions, reporter }: any) => {
```
elylucas, line 5: I get some errors when trying to run `npm run build`:

```
ERROR #11328  API.NODE.VALIDATION

The plugin "gatsby-plugin-page-creator" created a page without a valid default export.
```

This should create a static site with all those experiences prebuilt. 

Ideally, this would create the experiences to follow the same routing structure as the other apps (www.mysite.com/{slug}) and create a page for each experience that is in the space being used.

examples/gatsby/gatsby-spa/src/pages/ExperienceSlugExample.tsx
ryunsong-contentful: It doesn't have anything for i18n like Next does. This probably falls under the nice to have category but ultimately falls under the responsibility of the end user to figure out. Similar to how if they want to integrate with tailwind, that's on them and not on us to show how to do. 

Maybe some exceptions in the future will be when we have full-fledged tutorials and then point to an example that has bells and whistles.

examples/gatsby/gatsby-spa/README.md
ryunsong-contentful: I updated the README to make it work properly

examples/gatsby/gatsby-spa/src/pages/404.tsx
ryunsong-contentful: I updated the README to make it work properly

PR #897, opened 2025-01-07
chore(components): remove type and as field from built-in components [LUMOS-460, LUMOS-461]

chasepoirier:
## Purpose
As part of our content binding redesign, we are cleaning up the naming conventions and fields used for built-in components. This PR is for removing the AS field from the Text component and the Type field from the Heading component.

Tickets:
- https://contentful.atlassian.net/browse/LUMOS-460
- https://contentful.atlassian.net/browse/LUMOS-461

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39727,10 +39727,6 @@
         "node": ">=4"
       }
     },
-    "node_modules/studio-experiences-react-app": {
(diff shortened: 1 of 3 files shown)


chasepoirier:
Closing in favor of #900

PR #898, opened 2025-01-07
fix: don't spread props meant for custom components onto plain html wrapping elements

elylucas:
## Purpose
[ALT-1504]

Additional props were being applied to the html div container when wrapping was turned on. This would cause noisy console log messages (and error screens in React 19), to show when the app is in dev mode. This PR strips out any custom props when wrapping is turned on and ensures they don't get applied to the wrapper, and only passed to the underlying component.

## Approach

[ALT-1504]: https://contentful.atlassian.net/browse/ALT-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/test-apps/react-vite/src/components/ExpEditorMode.tsx
```diff
@@ -0,0 +1,20 @@
...
+const ExpEditorMode: React.FC<ExpEditorModeProps> = (props) => {
+  const { isInExpEditorMode, myValue, className } = props;
+  console.log('eee', { props });
+  return (
+    <section className={className}>
+      <div>isInExpEditorMode {isInExpEditorMode ? 'true' : 'false'}</div>
+      <div>myValue {myValue}</div>
+    </section>
+  );
+};
```
Chaoste, line 18: Do you intentionally left this testing component here or do you want that to get merged as well?

Chaoste (approved):
I was hoping to address this for such a long time. Thanks for taking over! 🚀
I have one question about the testing code but except for that it's looking good to me.

packages/test-apps/react-vite/src/components/ExpEditorMode.tsx
elylucas: no probably not, nice catch

PR #900, opened 2025-01-08
chore: rename and move variables to design tab for text and heading [LUMOS-482]

chasepoirier:
## Purpose
The `as` variable for the Text built-in component and the `type` variable for the Heading built-in component should be moved to the design tab and have the display name changed to `HTML tag`

Ticket: https://contentful.atlassian.net/browse/LUMOS-482

NOTE:
This has been tested and it does not break existing experiences that have content bound to these variables. Even though the variable is moved from the content to the design tab, previously set values (whether a bound entry, asset, or manual text) are persisted. When the value is changed from the design tab, it still works as expected.

Text component
Screenshot 2025-01-08 at 11 05 25 AM

Heading component
Screenshot 2025-01-08 at 11 05 35 AM

### DIFF
--- a/packages/components/src/components/Heading/index.ts
+++ b/packages/components/src/components/Heading/index.ts
@@ -67,9 +67,10 @@ export const HeadingComponentDefinition: ComponentDefinition = {
       defaultValue: 'Heading',
     },
     type: {
-      displayName: 'Type',
+      displayName: 'HTML tag',
       type: 'Text',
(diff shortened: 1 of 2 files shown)


jsdalton (approved):

Chaoste (approved):
I don't really understand why this is backward compatible but tested it and it's looking good. Makes a lot of sense to me :)

PR #908, opened 2025-01-14
fix(experience-builder): show the selection rectangle of nested patterns [SPA-2534]

Chaoste:
## Purpose

When selecting a nested pattern or its children, the SDK logic would treat the topmost parent pattern as the related one.

## Approach

Change the logic to consider the case of a nested pattern by adjusting the node location splitting. so far, we used `split('---')` which matched the [first occurrence](https://stackoverflow.com/questions/34648186/split-at-last-occurrence-of-character-then-join) of in the node location and didn't consider a path consisting of multiple `---`.
By using a regex, we match the last occurrence and thereby select the actual responsible pattern instead of the topmost.
Example regex101: https://regex101.com/r/XerKPS/1


SofiaMargariti (commented):
The code looks good and it works as expected, but there are a lot of if/else and special casing here. I think this code would benefit from some unit tests that cover the difference cases. Would you be able to add some?

SofiaMargariti (approved):
Thank you for adding the tests, it looks good to me!

PR #927, opened 2025-01-15
Validations/binding source type

ethan-ozelius-contentful:
### Context
There are certain component properties where it doesn't make sense to bind certain types of content. 

1. Binding an image to an Experience doesn't make any sense, nor does manually binding an image.
2. Binding an URL behavior to an asset also doesn't make sense.

## Purpose
Add a new optional validation called `bindingSourceType` to the existing validation object, enabling component properties to be optionally configured to restrict what types of content binding
(entry, manual, asserts, experience).

## Approach
Add a new Zod enum/array to describe the schema of this new validation, export that data structure for re-use in the `core` package.

### DIFF
--- a/packages/validators/src/schemas/v2023_09_28/experience.ts
+++ b/packages/validators/src/schemas/v2023_09_28/experience.ts
@@ -132,6 +132,10 @@ const ComponentTreeNodeSchema: z.ZodType<ComponentTreeNode> = BaseComponentTreeN
   children: z.lazy(() => ComponentTreeNodeSchema.array()),
 });
 
+const BindingSourceTypeEnumSchema = z
+  .array(z.enum(['entry', 'asset', 'manual', 'experience']))
+  .nonempty();
+
 export const ComponentVariableSchema = z.object({
   displayName: z.string().optional(),
   type: DefinitionPropertyTypeSchema,
@@ -140,6 +144,7 @@ export const ComponentVariableSchema = z.object({
   defaultValue: PrimitiveValueSchema.or(ComponentPropertyValueSchema).optional(),
   validations: z
     .object({
+      bindingSourceType: BindingSourceTypeEnumSchema.optional(),
       required: z.boolean().optional(),
       format: z.literal('URL').optional(),
       in: z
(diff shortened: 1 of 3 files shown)


packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -140,6 +144,7 @@ export const ComponentVariableSchema = z.object({
   defaultValue: PrimitiveValueSchema.or(ComponentPropertyValueSchema).optional(),
   validations: z
     .object({
+      bindingSourceType: BindingSourceTypeEnumSchema,
```
primeinteger, line 147: Should the `.optional()` be added here instead of in the `BindingSourceTypeEnumSchema` type definition?

That would allow the type to be used in places where we'd want to require a value too.

```suggestion
      bindingSourceType: BindingSourceTypeEnumSchema.optional(),
```

primeinteger: Right now since the `BindingSourceTypeEnumSchema` type has `.optional()` it creates a type like this:
```ts
type BindingSourceTypeEnumSchema = 'entry' | 'asset' | 'manual' | 'experience' | undefined;
```

Any time you use the `BindingSourceTypeEnumSchema` type it will always have the possibility of being undefined since it's built into the type.

I suggested removing the `.optional()` from the `BindingSourceTypeEnumSchema` type so that it produces this instead:
```ts
type BindingSourceTypeEnumSchema = 'entry' | 'asset' | 'manual' | 'experience';
```

I feel like that is a better type and more flexible because you can make it optional when you use it:

```ts
bindingSourceType?: BindingSourceTypeEnumSchema
```
or
```ts
bindingSourceType: BindingSourceTypeEnumSchema | undefined
```

ethan-ozelius-contentful: 100% agree.  paired on this and implemented this change.  Even added a bonus of `nonempty()` to the type to disallow consumers providing an empty array.

Screenshot 2025-01-22 at 3 00 58 PM

PR #930, opened 2025-01-15
[ALT-1508] fix: react-router import in gatsby spa to match the docs

ryunsong-contentful:
## Purpose
Update the react-router imports to match the Gatsby docs https://github.com/contentful/doc-app/pulls


primeinteger (commented):
The gatsby app's package.json should probably install `react-router-dom` as a dependency to match this change as well

ryunsong-contentful:
> The gatsby app's package.json should probably install `react-router-dom` (instead of `react-router`) as a dependency to match this change as well

Sounds good, I did a mini quest and basically the summary is that react-router-dom is a super-set of react-router (basically has everything react-router has and more)

primeinteger (approved):
Looks good! 👏

elylucas (approved):
Tested it out locally and worked like a charm 🚀

PR #931, opened 2025-01-16
fix: prevent missing components error in read only mode [LUMOS-208]

primeinteger:
## Purpose

This change fixes an issue with the Missing Components Error showing unintentionally when viewing experiences/patterns in read only mode.

|Before|After|
|----|----|
|Screenshot 2025-01-15 at 5 04 35 PM|Screenshot 2025-01-15 at 5 04 42 PM|


jsdalton (approved):
Great! 💯 

This is the key piece of logic I was missing last week: https://github.com/contentful/experience-builder/blob/2660b140e1a91b442b353a7ae78beb3d71f5fc91/packages/experience-builder-sdk/src/hooks/useDetectCanvasMode.tsx#L25C1-L32C6

PR #934, opened 2025-01-16
feat: add gatsby-ssg example app [ALT-1526]

elylucas:
## Purpose

Adds and example app to demonstrate how to do ssg with Gatsby


examples/gatsby/gatsby-ssg/gatsby-node.mjs
```diff
@@ -0,0 +1,74 @@
...
+      actions.createPage({
+        path: slug,
+        component: path.resolve(`./src/templates/ExperienceTemplate.tsx`),
+        context: {
+          experienceJson: JSON.stringify(experience),
+          stylesheet,
+          localeCode,
+        },
+      });
+      console.log(`created page for slug ${slug}`);
```
primeinteger, line 69: I noticed if you change an Experience's slug in the settings tab that the canvas will 404 until you restart the gatby app server. It also happens when you create a new Experience, you have to restart the gatby app server so it will "reindex" the page in this createPages process. 

Should this be mentioned in the readme?

examples/gatsby/gatsby-ssg/src/images/icon.png
primeinteger, line 1: I don't see anywhere this icon is being used. Could it be deleted?

examples/gatsby/gatsby-ssg/src/pages/index.tsx
primeinteger, line 1: What do you think about replacing this auto-generated index file with the modified one in the gatsby-spa example?  Ryun made that other index file to be a little more specific to gatsby example in studio.

primeinteger (approved):
I configured the env vars and tested this gatsby ssg example app and it seems to be working as expected. Very nice!

examples/gatsby/gatsby-ssg/gatsby-node.mjs
elylucas: Expected behavior, but probably worth mentioning in case someone isn't quite familar with how it works

examples/gatsby/gatsby-ssg/src/pages/index.tsx
elylucas: I redirected to '/home-page' in the other examples so did so here as well. Forgot to put it in.

PR #937, opened 2025-01-20
fix: let visibility display style take precedence when toggled [SPA-2468]

ChidinmaOrajiaku:
## Purpose

Let visibility display style take precedence when toggled

Following Amr discovery in the video below, some components have inline display which overrides the transform visibility `display: none` 

Bug Video

https://github.com/user-attachments/assets/8ab30fb0-fca9-4c1b-8feb-977944bc0a35

This PR ensures the transform visibility `display: none` takes precedence when the value is false

Fix after PR

https://github.com/user-attachments/assets/67f96e78-cbdf-409d-92e0-aa580c5a0e46


Chaoste (approved):
Nice!

PR #946, opened 2025-01-22
fix(visual-editor): handle circular pattern dependencies in the tree [SPA-2511]

Chaoste:
## Purpose
When accidentally causing a loop in the chain of pattern dependencies, the SDK would crash and the browser tab becomes not responsible.

## Approach
To address this, we want to render a similar error as we do for missing component definitions. To detect a loop during the render cycle, we pass through a list of `wrappingPatternIds` that gets extended when "diving into" a pattern node.

To detect the loop as early as possible (and not pollute the canvas with multiple instances of a single pattern), we have to know the ID of the opened experience ID. When finding the same ID again deeper inside the tree, we stop immediately the rendering cycle and render the error message. To know the global experience ID, we leverage the `blockId` of the root node. So far, it was the same as the root node id. Using this for the entry ID sounds more reasonable to me and saves us from adding another attribute or message to the communication layer.

Note: This handles only the editor case. For preview/ delivery, there will be a separate PR

Screenshot 2025-01-22 at 17 29 17

### DIFF
--- a/packages/visual-editor/src/components/DraggableBlock/Dropzone.tsx
+++ b/packages/visual-editor/src/components/DraggableBlock/Dropzone.tsx
@@ -31,6 +31,7 @@ type DropzoneProps = {
   className?: string;
   WrapperComponent?: ElementType | string;
   dragProps?: DragWrapperProps;
+  wrappingPatternIds?: Set<string>;
 };
 
 export function Dropzone({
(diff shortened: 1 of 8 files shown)


ChidinmaOrajiaku (approved):
Looks good and works well 👍

PR #962, opened 2025-01-27
fix(preview): detect and handle circular dependencies [SPA-2511]

Chaoste:
## Purpose

When there is a circular reference among used patterns, the fetchers and the renderer will fail with a "Maximum call stack size exceeded".

## Approach
To handle this properly, we have to pass down the parent patterns. If a pattern appears that is already part of that set, we bail to avoid an infinite loop.


packages/core/src/fetchers/shared/removeCircularReferences.ts
```diff
@@ -0,0 +1,35 @@
+import { ExperienceEntry, Link } from '@/types';
+
+export const removeCircularReferences = (
```
anwaar931, line 3: Would be nice, if we have a short info about how this function is removing circular references

anwaar931 (approved):
Code looks good to me. 👍

packages/core/src/fetchers/shared/removeCircularReferences.ts
Chaoste: Good call out, I forgot to push that commit ^^

PR #964, opened 2025-01-28
fix(core-fetchers): remove circular self references [SPA-2557]

Chaoste:
## Purpose
When creating a hyperlink that points to the experience entry itself, this breaks SSR applications preview/ delivery mode.

Screenshot 2025-01-28 at 11 11 53

## Approach
Similar to the circularity check for `usedComponents`, circular references are replaced by a plain link.


Spring3 (approved):
👍

PR #971, opened 2025-01-31
chore(pre-binding): update validators for prebinding fields [LUMOS-495]

chasepoirier:
## Purpose
Update the validator's package so that we can begin saving prebinding data to both patterns and experiences.

Ticket: https://contentful.atlassian.net/browse/LUMOS-495

NOTE: this is just to get us started, an additional ticket exists to finalize the field names in the validators package once they have been set. https://contentful.atlassian.net/browse/LUMOS-523

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2507,6 +2507,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -16,6 +16,7 @@ import type {
   ExperienceComponentTree,
   ComponentDefinitionPropertyType,
   BindingSourceTypeEnum,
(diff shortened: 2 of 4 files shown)


ryunsong-contentful (approved):
LGTM, only comment I have is that maybe we could publish a new schema version based on today's date? Could be something that helps with any reverting if anything comes to that or might provide us with a clean separation of what's unreleased for prebinding vs what is. Although maybe it's more of a hassle to move everything to the new schema if we went this route. Just a thought and how to setup the initial work

chasepoirier:
> LGTM, only comment I have is that maybe we could publish a new schema version based on today's date? Could be something that helps with any reverting if anything comes to that or might provide us with a clean separation of what's unreleased for prebinding vs what is. Although maybe it's more of a hassle to move everything to the new schema if we went this route. Just a thought and how to setup the initial work

Yeah publishing a new version is pretty complex, hence why we haven't done it yet. A new schema is really just meant for breaking changes and since these are just additions that's why we can get away with keeping the same one. If you check the git blame, you'll notice this file has been updated many times with additions since that date.

PR #978, opened 2025-02-04
feat: apply the built in styles via media queries in preview mode [SPA-2573]

Spring3:
## Purpose

The JS logic of style breakpoint detection + style injection wasn't reliably working on the initial render and always applied the styles for a default breakpoint 

In this PR I am changing this to always pass the Ctfl provided built-ins as media queries, while custom components are going to be wired up with the JS approach. It is not ideal, as the custom props might still be stuck with the wrong values, but that feature is experimental as a whole so, it will be investigated separately

## Approach

I've revamped the prop mapping and broke it into steps to improve the readability and clarity of the code.

All props are broken into 4 categories `ssrProps` `styleProps` `customDesignProps` and `contentProps`

`ssrProps` optionally contain the preset classname from the ssr style extraction
`styleProps` are ctfl provided built ins, that are going to be compressed into a media query style
`customDesignProps` - potentially not CSS convertible props, specific to the registered components. We know nothing of them, so the best we can do is let the JS breakpoint detection wire it up correctly
`contentProps` are Unbound/Bound/Component Value type of props.


DanweDE:
> It is not ideal, as the custom props might still be stuck with the wrong values, but that feature is experimental as a whole so, it will be investigated separately

Could you elaborate please?

Do you mean even when CSR kicks in, we might render the wrong breakpoint values there? This would be bad.

Or do you mean just on the server, we're picking one breakpoint and render that, which might be the wrong one depending on the user's device? If that's the issue, I'd say that's fine for now and expected, it's a more fundamental issue to solve.

Spring3:
> Do you mean even when CSR kicks in, we might render the wrong breakpoint values there? This would be bad.

That's the way it behaves now afaik. It is driven by the JS logic and we might need to research some way to trigger that check early enough for it to adjust for the initial render

What I'm trying to say is that there might a way to fix it. Just not something I was trying outside of this ticket.

I am going to test this tomorrow with more scenarios to see how it performs and will provide a better summary

packages/experience-builder-sdk/src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,179 @@
...
+            'justify-content': 'safe center',
+            'flex-direction': 'column',
+            'flex-wrap': 'nowrap',
+            'font-style': 'normal',
+            'text-decoration': 'none',
+            'box-sizing': 'border-box'
+          }
+        */
+    // I create a hash of the object above because that would ensure hash stability
+    const styleHash = md5(`${node.id}-${JSON.stringify(stylesForBreakpointWithoutUndefined)}`);
```
Chaoste, line 95: [optional, note for later]
Given the insights from [#993](https://github.com/contentful/experience-builder/pull/993), we could use here the breakpoint ID as well. This allows us to still reuse class names across nodes that are completely similar instead of creating one class for each node and each breakpoint for the whole tree.
Just a micro-optimization so we can definitely go ahead and improve this later.

```suggestion
    const styleHash = md5(`${breakpoint.id}-${JSON.stringify(stylesForBreakpointWithoutUndefined)}`);
```

packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.spec.tsx
```diff
@@ -383,31 +383,9 @@ describe('CompositionBlock', () => {
       expect(screen.getByText('New year eve')).toBeInTheDocument();
     });
 
-    it('renders nested patterns', () => {
+    it.skip('renders nested patterns', () => {
```
Chaoste, line 386: We should re-enable this test

Chaoste (commented):
The code changes themself look good to me. I'm not a fan of polluting the code files with all this fixture data but that's something for another day :)
In general, this is a great refactoring/ new feature - thanks for all the effort so far!

I will thoroughly test this tomorrow morning before approving. Besides that, I think we should at least bring back the test for nested patterns and adjust the fixture as called out in my first comment.

packages/experience-builder-sdk/src/hooks/useMediaQuery.ts
SofiaMargariti: Initially Daniel was using the breakpoint, but it wasn't working and then he switched to nodeId. We'd need to test both cases to confirm if that's still the case and to understand why this happens.

packages/experience-builder-sdk/src/utils/parseComponentProps.ts
```diff
@@ -0,0 +1,207 @@
...
+  resolveClassNamesFromBuiltInStyles,
+  resolveCustomDesignValue,
+  resolveBoundValue,
+  resolveHyperlinkValue,
+  resolveUnboundValue,
+}: {
+  mainBreakpoint: Breakpoint;
+  node: ComponentTreeNode;
+  componentDefinition: ComponentDefinition;
+  resolveClassNamesFromBuiltInStyles: (
```
Chaoste, line 43: Passing in this callback seems unnecessary as we just need the `node` and `breakpoints` for this.

Chaoste: Will be tackled in https://github.com/contentful/experience-builder/pull/999

packages/experience-builder-sdk/src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // and prefix the className to make sure the value can be processed
+    const className = `cfstyles-${styleHash}`;
+
+    // I save the generated hashes into an array to later save it in the tree node
+    // as cfSsrClassName prop
+    // making sure to avoid the duplicates in case styles for > 1 breakpoints are the same
+    if (!currentNodeClassNames.includes(className)) {
```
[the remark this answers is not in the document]
SofiaMargariti: Maybe we can clean this up in the followup PR.

SofiaMargariti (approved):
Tests look good to me and I tested this PR in a few different scenarios so I am happy to approve.

Tested with:
* Built-in and custom design values
* On different breakpoints
* Within experiences, patterns and nested patterns
* Tested the same experience with both CSR and SSR

There is a small comment about an unnecessary variable but we can clean that up later.

packages/experience-builder-sdk/src/hooks/useMediaQuery.ts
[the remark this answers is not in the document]
Chaoste: Already done in this follow-up PR: https://github.com/contentful/experience-builder/pull/999

PR #994, opened 2025-02-17
chore(circleci): remove nextjs deployment to Vercel []

Chaoste:
The job for deploying the nextjs marketing & react-vite demo started failing today on all branches and is not available any more on Vercel.
To unblock the pipeline, drop the job from the config.


SofiaMargariti (approved):
Looks like you need to comment out the next job as well, project `prj_HoAvIbgvZ3gYJDLCAaNsHIpBvI0k` is also missing.

Chaoste:
Seems to have been resolved on its own. The projects are back online and the pipeline green

PR #995, opened 2025-02-17
fix: update active breakpoint correctly

SofiaMargariti:
## Purpose

Makes sure that active breakpoint is updated correctly when breakpoints as eventually updated.

## Approach

**Fix 1**
The root of the problem was that initial state is only used on the first render (see also this [SO question](https://stackoverflow.com/questions/58818727/react-usestate-not-setting-initial-value)), but in the first render we'll always pass in `breakpoints: []` which leads to `mediaQueryMatches` being empty. On the next render when breakpoints have been fetched this line doesn't actually update the `mediaQueryMatches`:
```
 const [mediaQueryMatches, setMediaQueryMatches] =
    useState<Record<string, boolean>>(initialMediaQueryMatches);
```
We need to instead update the `mediaQueryMatches` inside the useEffect [here](https://github.com/contentful/experience-builder/pull/995/files#diff-c71237ef291dccb20a3d2be94092a02b56c8169fc3f7214e9855fac0a5f3d504R26) (`breakpoints` is a dependency for the useEffect)

There are two almost identical useBreakpoints hook one in `visual-editor` and one in the `experience-builder-sdk`. It was the `experience-builder-sdk` that had the issue. `visual-editor` hook already had the above fix.

**Fix 2**
Fallback and active breakpoints were also moved into the resolveDesignValue callback to make sure they always have the updated values, as they also depend on the `mediaQueryMatches`. After fix 1 this fix might not be necessary but I'd suggest we keep it as this version has been tested by the customer.

A screenshot from the customer's experience with this fix and some extra logging inside the `resolveDesignValue` callback:
Screenshot 2025-02-18 at 11 29 47

I don't have a screenshot from before but essentially **activeBreakpointIndex was always 0** and **mediaQueryMatches where empty**.

### DIFF
--- a/packages/core/src/utils/breakpoints.spec.ts
+++ b/packages/core/src/utils/breakpoints.spec.ts
@@ -28,6 +28,7 @@ describe('getValueForBreakpoint for css values', () => {
         valuesByBreakpoint,
         breakpoints,
         desktopIndex,
+        desktopIndex,
         variableName,
       );
       expect(value).toEqual(desktopValue);
@@ -40,6 +41,7 @@ describe('getValueForBreakpoint for css values', () => {
         valuesByBreakpoint,
         breakpoints,
         tabletIndex,
+        desktopIndex,
         variableName,
       );
       expect(value).toEqual(tabletValue);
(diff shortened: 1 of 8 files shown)


Spring3 (approved):
nice! it was easy to overlook

dimitrycf (approved):
Looks great! 
tested also manually  several times. Here's just one video https://www.loom.com/share/b6a033db59f34e629f37de7d55512630

Jest tests are top notch! :chefs-kiss:  👍

PR #999, opened 2025-02-19
refactor: css generation

Chaoste:
## Purpose

Improve the readability of critical code, reduce duplicated code, remove not needed code.

## Approach

- Remove debugging comments of actual CSS code
- Generalise repeated code for creating CSS code, e.g. filtering for undefined rules & checking for structures with relative height

## TODO
- [ ] Find out why we're creating default styles in `ssrStyles.ts` for pattern nodes, cc @anwaar931


packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -143,22 +117,16 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
...
 
-        const defaultPatternDivStyles: Record<string, string> = Object.fromEntries(
-          Object.entries(buildCfStyles({}))
-            .filter(([, value]) => value !== undefined)
-            .map(([key, value]) => [toCSSAttribute(key), value]),
-        );
-
-        // I create a hash of the object above because that would ensure hash stability
-        const styleHash = md5(JSON.stringify(defaultPatternDivStyles));
+        // TODO: Why are we creating default styles here and where are they rendered in the end?
```
anwaar931, line 120: Hey Thomas, I think this was added back then for fixing a bug related to pattern not rendering correctly in SSR. So adding default styles to the div

```
<div data-test-id="assembly">
```
fixed the issue. 
However, I tested it by removing this and its rendering just fine in SSR. So IMO we can safely remove this. It will also reduce the complexity.

packages/core/src/utils/styleUtils/ssrStylesDesignTokenResolution.spec.tsx
```diff
@@ -180,18 +180,16 @@ describe('custom component with builtInStyles which are supporting design tokens
     expect(
       (customComponent.variables.cfSsrClassName as DesignValue).valuesByBreakpoint
         .desktop as string,
-    ).toEqual(
-      'cf-ae92b9eb2c20a262573a8085db3dbd67 cf-1fe22311d2b832dd529a9ce5e8d77281 cf-63009ce9ca644a23e001721d6358e491',
-    );
+    ).toEqual(expect.stringMatching(/cf-\w{32} cf-\w{32} cf-\w{32}/));
```
Spring3, line 183: is the point of this test - to validate that the hash matches the regexp or is it to make sure that after a change in the PR the hash stays the same compared to the main branch of the project?

This change makes the test check the hash format, which I think is not the intention

packages/core/src/utils/styleUtils/ssrStylesDesignTokenResolution.spec.tsx
```diff
@@ -180,18 +180,16 @@ describe('custom component with builtInStyles which are supporting design tokens
...
+    expect(styles).toMatchInlineSnapshot(
+      `".cf-7b1d28fe46d9488fa595e4ff21270f2e{box-sizing:border-box;margin:2rem;padding:2rem;background-color:white;width:50%;height:25%;border:1px solid black;gap:2rem 2rem;font-size:1rem;color:black;}@media(max-width:992px){.cf-af21918d8656639646a9eb86a7ca06e2{box-sizing:border-box;margin:1.5rem;padding:1.5rem;background-color:black;width:75%;height:50%;border:3px solid black;gap:1.5rem 1.5rem;font-size:0.75rem;color:green;}}@media(max-width:576px){.cf-5a0d66d5b7e5c60dac49f3a86e4c1c8d{box-sizing:border-box;margin:1rem;padding:1rem;background-color:red;width:100%;height:100%;border:0px solid transparent;gap:1rem 1rem;font-size:1.5rem;color:orange;}}"`,
```
Spring3, line 192: the fact that the hashes have changed here make me anxious. If this PR was a pure refactor, the hash would be the same

packages/core/src/utils/styleUtils/ssrStylesPattern.spec.ts
```diff
@@ -1425,8 +1425,8 @@ describe('pattern component', () => {
...
...55, 255, 0);gap:10px 10px;}.cf-1031d744e26de7b4e9422de0e74ae5cd{box-sizing:border-box;padding:0 0 0 0;background-color:rgba(255, 255, 255, 0);grid-column:span 6;border:0px outside rgba(255, 255, 255, 0);gap:0px 0px;align-items:start;justify-content:safe center;flex-direction:column;flex-wrap:nowrap;}.cf-4c0b9d7e44ec5898b6f5480ac5b1a634{box-sizing:border-box;padding:0 0 0 0;background-color:rgba(255, 255, 255, 0);grid-column:span 6;border:0px outside rgba(255, 255, 255, 0);gap:0px 0px;align-items:center;justify-content:safe center;flex-direction:column;flex-wrap:nowrap;}.cf-2745e04133568665b6d779d76342d59f{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;width:100%;max-width:none;}.cf-fe878a9e705110f2fee7b5698943918a{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;width:100%;max-width:none;font-size:16px;font-weight:400;line-height:20px;letter-spacing:0px;color:rgba(0, 0, 0, 1);text-align:left;text-transform:none;}.cf-0326e88429ff35544d5acf70391db2b2{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;}@media(max-width:992px){.cf-92b718a315702272accd18eb39a4a427{box-sizing:border-box;}}@media(max-width:576px){.cf-8187cd93307f9e0ceffff3cad17ce4d3{box-sizing:border-box;}}"`,
```
Spring3, line 1429: same question here - what made the hashes change?

packages/core/src/utils/styleUtils/stylesUtils.spec.ts
```diff
@@ -1,66 +1,94 @@
...
-      value: 'auto',
+  describe('createJoinedCSSRules', () => {
+    it('should join a given object of css rules into a string', () => {
+      const cssRules = {
+        background: 'green',
+        color: 'white',
+        'font-size': '1rem',
+      };
+
+      const res = createJoinedCSSRules(cssRules);
```
Spring3, line 77: to me this function name sounds very crypric. What is the Joined CSS rule and what is the joined vs not joined css rule.

Maybe we call it `toMinifiedCSS`?

packages/core/src/utils/styleUtils/stylesUtils.ts
```diff
@@ -20,19 +26,22 @@ export const toCSSAttribute = (key: string) => {
...
-    .reduce(
-      (acc, [key, value]) =>
-        `${acc}
-        ${toCSSAttribute(key)}: ${value};`,
-      '',
+    .map(([key, value]) =>
+      useWhitespaces ? `${toCSSAttribute(key)}: ${value};` : `${toCSSAttribute(key)}:${value};`,
     );
+  return rules.join(useWhitespaces ? '\n' : '');
+};
```
Spring3, line 39: do we have a use case where we need white spaces?

packages/core/src/utils/styleUtils/stylesUtils.ts
```diff
@@ -98,7 +107,26 @@ export const buildCfStyles = ({
     objectFit: cfImageOptions?.objectFit,
     objectPosition: cfImageOptions?.objectPosition,
   };
+  const filteredRules = Object.fromEntries(
```
Spring3, line 110: I think before it was called `cssWithoutUndefined`. I am still a fan of that name vs `filteredRules`. It tells me exactly what that variable holds.

packages/core/src/utils/styleUtils/stylesUtils.ts
```diff
@@ -68,7 +77,7 @@ export const buildCfStyles = ({
   cfColumnSpan,
   cfVisibility,
 }: Partial<StyleProps>): CSSProperties => {
-  return {
+  const translatedRules = {
```
Spring3, line 80: Can we agree to call `css` instead of `rules`? We are dealing with css here so in my opinion there is no need to alias it with a different term

packages/core/src/utils/styleUtils/stylesUtils.ts
```diff
@@ -98,7 +107,26 @@ export const buildCfStyles = ({
     objectFit: cfImageOptions?.objectFit,
     objectPosition: cfImageOptions?.objectPosition,
   };
+  const filteredRules = Object.fromEntries(
+    Object.entries(translatedRules).filter(([, value]) => value !== undefined),
+  );
+  return filteredRules;
 };
+
+export const buildCfStylesWithSpecialCasing = (
```
Spring3, line 116: this name is too cryptic. If I had to choose between `buildCfStyles` and `buildCfStylesWithSpecialCasing` I wouldn't know which one to call.

Can we refactor it to be called `adjustStructureComponent`, which takes teh `cfStyles` as input (so `buildCfStyles` is called outside) and then returns adjusted styles as output?

packages/core/src/utils/styleUtils/stylesUtils.ts
```diff
@@ -98,7 +107,26 @@ export const buildCfStyles = ({
...
+  const cfStyles = buildCfStyles(designProperties);
+  if (!node.children.length && isStructureWithRelativeHeight(node.definitionId, cfStyles.height)) {
+    return {
+      ...cfStyles,
+      minHeight: EMPTY_CONTAINER_HEIGHT,
+    };
+  }
+  return cfStyles;
+};
+
```
Spring3, line 129:
```suggestion
export const buildCfStylesWithSpecialCasing = (
  cfStyles: CfStylesType,
  node: ComponentTreeNode,
) => {
  if (!node.children.length && isStructureWithRelativeHeight(node.definitionId, cfStyles.height)) {
    return {
      ...cfStyles,
      minHeight: EMPTY_CONTAINER_HEIGHT,
    };
  }
  return cfStyles;
};

```

Spring3 (changes requested):
I left many comments, but I generally have these points as a reason why I don't think that this PR needs to be merged in this current state:
1. The comments were left there on purpose, so that in a couple of months when we read that code, we can see a visual example of how the data gets transformed and understand what is happening without adding any extra logging. I am puzzled to understand why they are being deleted.
2. I personally find certain namings too cryptic
* for example `buildCfStyles` vs `buildCfStylesWIthSpecialCasing` - I feel like we need to have only 1 function - `buildCfStyles`. The rest are additional transformers that we can call as a chain, but now we have 2 functions that do the same, just 1 is smarter than the other and now I, as a developer, need to know when to call one and when the other (and that makes me question - why do we have 2 functions?)
3. Hash changes in the tests. The goal was to make sure that hashes stayed unchanged. The fact that they have changed in this PR, makes me alert, because that means that output styles are going to change to everyone who is using this SDK in prod with SSR. What made them change?
4. The fact that instead of having hardcoded hashes, we are now testing their format. That check will never fail and that's not the purpose of those tests

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -143,36 +104,6 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
...
-        // pattern nodes again for every instance. Make sure to not overwrite the values from previous instances.
-        if (!currentNode.variables.cfSsrClassName) {
-          currentNode.variables.cfSsrClassName = {
-            type: 'DesignValue',
-            valuesByBreakpoint: {
-              [breakpointIds[0]]: className,
-            },
-          };
-        }
-
```
Spring3, line 175: This chunk of code was the one adding the `cfSsrClassName` prop to the returned tree node.

Have I missed another place where the same thing was done? Without it, I think that style extraction will stop working and will always be done client side

packages/core/src/utils/styleUtils/ssrStylesDesignTokenResolution.spec.tsx
Chaoste: Judging by the test name and previous code lines, I think this test was not about testing a specific hash string. To reduce the dependency on the outcome of `md5`, it only checks now that three class names are created but doesn't anymore fail when changing the hash generation function itself.

If we really want to test the result of `md5()`, I would do this in a separate test instead of all tests checking this.

packages/core/src/utils/styleUtils/ssrStyles.ts
Chaoste: Yes, I have tested the PR and also talked to @anwaar931  who added this part of the code and who had more background knowledge. I tested, checked the code if it actually had an impact, and then asked him to confirm that this code had no effect on the function result. This logic created a class name independent of the current node and which all consists of default styles. We never actually use it which is why we can remove it.

packages/core/src/utils/styleUtils/ssrStylesDesignTokenResolution.spec.tsx
Chaoste: Mmh, not sure what would be the correct prefix instead of `refactor` to be honest.

The hash generation changed as before we used the object of styles and stringified it before calling `md5()`. I changed it to use the resulting CSS string right away, so we have one less stringifying. The string itself is different (JS object vs. CSS rules) which is why the hash is different.

packages/core/src/utils/styleUtils/ssrStylesStructureComponent.spec.ts
```diff
@@ -157,18 +157,16 @@ describe('section component', () => {
     ).toBeDefined();
     expect(
       (section.variables.cfSsrClassName as DesignValue).valuesByBreakpoint.desktop as string,
-    ).toEqual(
-      'cf-c3b88e3d9a2c0a14abb4aae89756f03a cf-2b1553d9ba4c1c47dcc5e2fa6d922afe cf-f716508ce7bdaa5b5506b78e104f9816',
-    );
+    ).toEqual(expect.stringMatching(/cf-\w{32} cf-\w{32} cf-\w{32}/));
```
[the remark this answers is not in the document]
Chaoste: My intention was to remove the dependency on the cryptic hashes as it doesn't tell us what the resulting CSS looks like. I can bring it back for now, but I would honestly prefer a new test that checks the resulting CSS rules instead of comparing two hashes with each other. It makes it quite hard to reason about this test and so far, we only updated those hashes blindly when making changes to the style generation logic.

packages/core/src/utils/styleUtils/stylesUtils.ts
Chaoste: I didn't want to change the existing logic that would use whitespace and newlines for the editor mode. I guess this was done for easier inspecting the CSS code.

packages/core/src/utils/styleUtils/stylesUtils.ts
Chaoste: At this point, it's not yet CSS as keys like `boxSizing` are not valid key names which is why I didn't want to cause confusion by giving it a name that doesn't reflect the variable. How can we call this intermediate state between design value and actual CSS rule that is somehow a mix of both?

Chaoste: Ah I just see that React calls it `CSSProperties`! Will rename it accordingly.

packages/core/src/utils/styleUtils/stylesUtils.ts
Chaoste: I don't know why we're running this check only in some cases which is why I'm not merging it into `buildCfStyles` for now. I like your first recommendation - will use that name and remove `buildCfStyles`

packages/core/src/utils/styleUtils/stylesUtils.spec.ts
Chaoste: I was thinking of the function `Array.join` which turns an array into a joined string. Happy to rename it though if that makes it easier to understand. As it does creat both normal CSS and minified CSS, I would like to go with `stringifyCSSProperties` (thinking of JSON.stringify) or something similar to that. Or is `toCSS` maybe enough?

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -247,28 +147,6 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
           );
         },
       });
-      /**
-     * propsByBreakpoint {
```
SofiaMargariti, line 251: When I was reviewing the original PR, I found these comments quite helpful. Could we not remove them?
If we don't want them to bloat the code here, they could also be moved to the docstring of the respective functions that produce each data structure.
So this `propsByBreakpoint` example can be moved to the `indexByBreakpoint` function.
And the `stylesForBreakpoint` example later on can go into the doc string of builtCfStyles this way they are easy to see every time you use each function.

Chaoste: I was having a hard time to visually parse the whole function. So moving it to the docstring of each function would be great. Will do that!

packages/core/src/utils/styleUtils/ssrStylesDesignTokenResolution.spec.tsx
Chaoste: [Postponing the discussion and reverting my change]

packages/core/src/utils/styleUtils/ssrStylesStructureComponent.spec.ts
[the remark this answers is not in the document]
Chaoste: [Postponing the discussion and reverting my change]

packages/core/src/utils/styleUtils/stylesUtils.spec.ts
```diff
@@ -1,66 +1,94 @@
...
 
-    expect(result).toBe('100%');
+    it('adds whitespaces when boolean is true', () => {
+      const cssRules = {
+        background: 'green',
+        color: 'white',
+        'font-size': '1rem',
+      };
+
+      const res = createJoinedCSSRules(cssRules, true);
```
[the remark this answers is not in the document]
Chaoste: My bad! I left out the actual usage of this function at `buildStyleTag` in `stylesUtils.ts`. That one created the CSS with whitespaces as well perviously.

packages/core/src/utils/styleUtils/stylesUtils.ts
Chaoste: I think I finally understood that this is only meant to be used for editor mode where empty containers should have a min height. I'm not exactly sure if we're already doing this accidentally for preview with the currently published SDK version.

Chaoste:
Hey @Spring3 @SofiaMargariti
Thanks for your reviews and sorry for these big changes that might require some time to go through. I tried to incorporate all your feedback:
- Rename cryptic function names and follow existing terminology (example "rules")
- Keep example data format comments. I shrinked them by a bit but added more of them.
- I also added docstrings to bigger functions and added example inputs & outputs as requested.
- I reverted the test changes that replace the actual hash strings by regular expressions. We're checking the hashes again in every place but we're using `toMatchInlineSnapshots` as this allows us to manually update them in one go without having to go manually replace each string.

packages/core/src/utils/styleUtils/ssrStylesPattern.spec.ts
```diff
@@ -1554,8 +1554,8 @@ describe('pattern component', () => {
...
...ckground-repeat:no-repeat;}.cf-1031d744e26de7b4e9422de0e74ae5cd{box-sizing:border-box;padding:0 0 0 0;background-color:rgba(255, 255, 255, 0);grid-column:span 6;border:0px outside rgba(255, 255, 255, 0);gap:0px 0px;align-items:start;justify-content:safe center;flex-direction:column;flex-wrap:nowrap;}.cf-4c0b9d7e44ec5898b6f5480ac5b1a634{box-sizing:border-box;padding:0 0 0 0;background-color:rgba(255, 255, 255, 0);grid-column:span 6;border:0px outside rgba(255, 255, 255, 0);gap:0px 0px;align-items:center;justify-content:safe center;flex-direction:column;flex-wrap:nowrap;}.cf-2745e04133568665b6d779d76342d59f{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;width:100%;max-width:none;}.cf-fe878a9e705110f2fee7b5698943918a{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;width:100%;max-width:none;font-size:16px;font-weight:400;line-height:20px;letter-spacing:0px;color:rgba(0, 0, 0, 1);text-align:left;text-transform:none;}.cf-0326e88429ff35544d5acf70391db2b2{box-sizing:border-box;margin:0 0 0 0;padding:0 0 0 0;}@media(max-width:992px){.cf-92b718a315702272accd18eb39a4a427{box-sizing:border-box;}}@media(max-width:576px){.cf-8187cd93307f9e0ceffff3cad17ce4d3{box-sizing:border-box;}}"`,
```
Spring3, line 1558: is this an expected diff? Sorry, it's hard to understand just from the looks if there is an impact or not

Screenshot 2025-03-04 at 18 08 49

packages/experience-builder-sdk/src/utils/parseComponentProps.ts
```diff
@@ -122,36 +122,21 @@ export const parseComponentProps = ({
...
         contentProps[propName] = propDefinition.defaultValue;
         break;
       default:
         break;
     }
   }
+  /* [Data Format] After resolving all properties, `styleProps` contains solely the plain design values
+   * styleProps = {
+   *   cfMargin: { desktop: '42px', tablet: '13px' },
+   *   cfBackgroundColor: 'rgba(246, 246, 246, 1)',
```
Spring3, line 135: is it expected to not be prefixed by the breakpoint>

Spring3 (commented):
Thanks for bringing back the code samples.

The code looks good to me, but before approving I wanted to make sure that the difference in styles is planned, because we don't want to break something on accident.

Please take a look at the 2 comments I left above.

Spring3 (approved):

packages/core/src/utils/styleUtils/ssrStylesPattern.spec.ts
Chaoste: Discussed per DM that this was caused by removing the default class names which only defined `box-sizing: border-box`. To be safe, we define it now globally as well in case a node has no properties and thus no specific class with this rule.

PR #1008, opened 2025-02-20
feat: only provide isInExpEditorMode if activated in options [SPA-2607]

Chaoste:
## Purpose
We added the prop `isInExpEditorMode` as some customers want to render different content in editor mode than in delivery.
However, this will cause a React warning for every (non-structural) component.

Feb 20 2025 Sprint 2 Screenshot

## Approach
To still allow this level of customization without disrupting simple use cases, we only provide it if activated through the registration options.

I noticed that we still have two options that have no effect anymore on the SDK. I removed those (`wrapContainerTag` and `wrapContainer`)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -121,10 +121,12 @@ export type ComponentRegistration = {
   component: React.ElementType;
   definition: ComponentDefinition;
   options?: {
+    /**
+     * If true, the component receives the optional boolean property `isInExpEditorMode` to
+     * render different content between editor and delivery mode.
+     */
+    enableCustomEditorView?: boolean;
     wrapComponent?: boolean;
-    /** @deprecated use wrapContainer instead */
-    wrapContainerTag?: keyof JSX.IntrinsicElements;
-    wrapContainer?: keyof JSX.IntrinsicElements | React.ReactElement;
     wrapContainerWidth?: React.CSSProperties['width'];
   };
 };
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -121,10 +121,12 @@ export type ComponentRegistration = {
...
   options?: {
+    /**
+     * If true, the component receives the optional boolean property `isInExpEditorMode` to
+     * render different content between editor and delivery mode.
+     */
+    enableCustomEditorView?: boolean;
     wrapComponent?: boolean;
-    /** @deprecated use wrapContainer instead */
-    wrapContainerTag?: keyof JSX.IntrinsicElements;
-    wrapContainer?: keyof JSX.IntrinsicElements | React.ReactElement;
```
SofiaMargariti, line 127: Why are we removing the `wrapContainer`? This is not deprecated from what I understand. It is also mentioned in the docs: https://www.contentful.com/developers/docs/experiences/component-wrapper/#disable-the-wrapping-container

Chaoste: I guess we should then update our docs as this is not supported by our code. If you search for `wrapContainer` you see that it's only defined in types but never consumed

SofiaMargariti: Hmm, maybe we should double check with altitude in that case, just to understand the reasons for not being implemented.

Chaoste: Alright, will remove it again from this PR and ask them 👍🏻

SofiaMargariti (approved):

PR #1012, opened 2025-02-20
fix: restore wrapContainer option to work again

elylucas:
## Purpose

Adds back in the functionality of specifying a custom wrapping container tag that was accidentally removed in https://github.com/contentful/experience-builder/commit/22ef6775df4a551eb0c7f541f3f8ab14069b2e67#diff-032b7a2eac51d40c7a9354cf29f7e1e654073acfd5af7a0b0dc50267b1d309c2

### DIFF
--- a/packages/experience-builder-sdk/src/utils/withComponentWrapper.tsx
+++ b/packages/experience-builder-sdk/src/utils/withComponentWrapper.tsx
@@ -24,8 +24,6 @@ export function withComponentWrapper<T>(
   Component: React.ElementType,
   options: ComponentRegistration['options'] = {
     wrapComponent: true,
-    wrapContainerTag: 'div',
-    wrapContainer: 'div',
   },
 ) {
   const Wrapped: React.FC<CFProps & T> = ({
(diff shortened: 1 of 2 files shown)


primeinteger (approved):
Good catch! Looks good to me 🚀

packages/experience-builder-sdk/src/utils/withComponentWrapper.tsx
```diff
@@ -40,8 +38,11 @@ export function withComponentWrapper<T>(
       ToolTipAndPlaceholder,
       ...restOfDragProps
     } = dragProps;
+
+    const Tag = (options.wrapContainer || options.wrapContainerTag || 'div') as string;
```
SofiaMargariti, line 42: Maybe we could just cleanup the `wrapContainerTag` instead of supporting it? As far as I can see it's marked as deprecated and wasn't working until now, so no one is relying on it.

Chaoste:
Closing this PR as we removed the unused `wrapContainerTag` argument [here](https://github.com/contentful/experience-builder/pull/1111/files#diff-4c2458380e50c967c70d7afe819dc7412180f9f817043f27441de3498e4d0479L132).

PR #1022, opened 2025-02-26
fix: types export from core package [SPA-2612]

SofiaMargariti:
## Purpose

Ensure that types are properly exported so that they are provide hints/autosuggestion when writing component registrations.

Additionally export `ComponentRegistration` type as it seems to be more useful than `ComponentDefinition` when defining components.

## Approach

I noticed that when hovering over an import of `'@contentful/experiences-core/constants'` or `'@contentful/experiences-core'` in the `experiences-sdk-react` built version, visual studio was able to resolve and follow the path of those files unlike the `'@contentful/experiences-core/types'`

https://github.com/user-attachments/assets/b1b152ae-1bbb-4456-9b28-702071b49e78

This seemed to be the reason it couldn't resolve types for `ComponentDefinition` and `ComponentRegistration`.

After adding `./types` in the `exports` of the core package the problem was resolved. Now we get autocomplete suggestions and proper type-checking when defining components.

https://github.com/user-attachments/assets/a64801fd-a736-4b65-87bc-321f6741fbbf

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,7 +21,8 @@
   ],
   "exports": {
     ".": "./dist/index.js",
-    "./constants": "./dist/constants.js"
+    "./constants": "./dist/constants.js",
+    "./types": "./dist/types.d.ts"
   },
   "publishConfig": {
     "registry": "https://npm.pkg.github.com/"
--- a/packages/experience-builder-sdk/src/index.ts
+++ b/packages/experience-builder-sdk/src/index.ts
@@ -26,6 +26,10 @@ if (typeof window !== 'undefined') {
   window.__EB__.sdkVersion = SDK_VERSION;
 }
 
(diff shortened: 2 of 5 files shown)


Spring3 (approved):
👏

dimitrycf:
I verified it by:
 - creating new react project
 - adding current (not fixed) version of the SDK
 - monkey patching the `@contentful/studio-sdk-react/package.json` section `export` as Sofia has suggested (and exports too)
 - tested that imports started working


However, I wondered if there were an easier way? (eg. aren't PR branches deployed as dev-branches ? (i couldn't find anything on NPM with sha of this commit https://www.npmjs.com/package/@contentful/experiences-sdk-react?activeTab=versions) ? 

Eventually I tried installing dependencies via file-system (then transitive deps should be explicitly installed). and it worked too

image


image

PR #1035, opened 2025-03-05
fix: remove the key update via the use effect []

Spring3:
## Purpose

Removing the key overwrite after a use effect as we have fixed the breakpoint update.

This is where the useEffect was introduced: https://github.com/contentful/experience-builder/pull/979


dimitrycf (approved):
Looks good.

Tested manually 
https://www.loom.com/share/df27b95616334492a453f67723d58079

PR #1036, opened 2025-03-05
chore(prebinding): modify schema to support entity link directly on defaultValue [LUMOS-594]

chasepoirier:
## Purpose
Modifies the defaultValue of patternPropertyDefinitions to remove the need for using `dataSource` on a pattern entry and directly link the entry in the defaultValue.

Ticket: https://contentful.atlassian.net/browse/LUMOS-594


jsdalton (approved):
👍

PR #1047, opened 2025-03-20
feat(prebinding): render nested patterns with prebinding [LUMOS-525]

chasepoirier:
## Purpose
Modify rendering of assemblies to work with nested patterns.

### DIFF
--- a/packages/experience-builder-sdk/src/core/preview/assemblyUtils.spec.ts
+++ b/packages/experience-builder-sdk/src/core/preview/assemblyUtils.spec.ts
@@ -5,7 +5,10 @@ import {
   createAssemblyEntry,
 } from '../../../test/__fixtures__/assembly';
 import { assets, entries } from '../../../test/__fixtures__/entities';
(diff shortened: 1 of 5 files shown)


packages/experience-builder-sdk/src/core/preview/assemblyUtils.spec.ts
```diff
@@ -23,7 +23,11 @@ describe('resolveAssembly', () => {
       locale: 'en-US',
     });
 
-    const result = resolveAssembly({ node: containerNode, entityStore });
+    const result = resolveAssembly({
+      node: containerNode,
+      entityStore,
+      parentPatternProperties: {},
```
jsdalton, line 29: A test or two that specify pattern property inputs might be nice but since it's just a straightforward merge you can just use your best judgement there.

jsdalton (approved):
Looks good 💯 . A comment or two might be helpful to contextualize some of the behaviors here. Like it's just a bit hard to connect the dots that we're accumulating parent pattern properties as we traverse the tree. Or even a bit more info in the PR description for posterity.

PR #1063, opened 2025-03-26
chore: use --git-tag-command to override lerna version tag for prereleases

SofiaMargariti:
## Purpose

This is an attempt to force lerna to generate changelog from the last stable version. 

## Approach

Right now the problem is that, on the last step of releasing the package when merging to `main` we instruct lerna to generate a changelog using the `--conventional-commits` argument. This generates the changelog but updates it with placeholder text like:

Screenshot 2025-03-26 at 12 07 46

This is because lerna will use commits between the latest version and the last beta version, e.g. `v1.34.1-beta.0..v1.34.1`. But this diff doesn't include any meaningful commits, e.g.

Screenshot 2025-03-26 at 11 59 13

In reality what we need is the diff `v1.34.0..v1.34.1`

It seems that lerna is using the git tags for this comparison, so if we don't generate tags for the prerelease versions then the changelog will compare with the last stable version. This is a solution suggested in the relevant [thread](https://github.com/lerna/lerna/issues/2248). It's a bit obscure but there doesn't seem to be any better solution.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -85,11 +85,11 @@ jobs:
           git config user.email "${{ github.actor}}@users.noreply.github.com"
 
           if [ ${{ inputs.publish-prerelease }} ]; then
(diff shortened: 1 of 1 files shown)


SofiaMargariti:
> Nice discovery! The fix looks good to me :)
> 
> One thing to call out: We can't quickly checkout prereleases with something like `git checkout 1.34.0-beta` but instead we would have to find the according commit hash.

Good callout! Maybe we can look at tagging with a custom tag, unrelated to lerna, but I wouldn't add it if it's not necessary.

PR #1068, opened 2025-03-26
chore: remove reduntant private NPM registry authentication

SofiaMargariti:
## Purpose

We no longer use a private npm.js registry (we're using the GH one). Also build and check steps don't require access to private registries.


Chaoste (approved):
Good catch! :)

PR #1071, opened 2025-03-28
fix: hiding nodes bug [SPA-2656]

dimitrycf:
## Purpose

adds check for `cfVisibility` when generating wrapper styles. 
This bug (and this logic from `useComponentProps`) only applies to packages/visual-editor, when breakpoints always have current one (eg. defined via `useBreakpoints`  and `resolveDesignValue()` )

Within tests, it is important that component definition contains `cfVisibility` property. Otherwise iteration won't work.

### DIFF
--- a/packages/visual-editor/src/hooks/useComponentProps.spec.ts
+++ b/packages/visual-editor/src/hooks/useComponentProps.spec.ts
@@ -295,7 +295,7 @@ describe('useComponentProps', () => {
       type: 'block',
     };
 
-    it.only('should NOT set the component size in wrapperStyles when drag wrapper is enabled', () => {
+    it('should NOT set the component size in wrapperStyles when drag wrapper is enabled', () => {
       const newNode: ExperienceTreeNode = structuredClone(node);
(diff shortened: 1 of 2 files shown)


packages/visual-editor/src/hooks/useComponentProps.spec.ts
```diff
@@ -258,10 +295,50 @@ describe('useComponentProps', () => {
       type: 'block',
     };
 
-    it('should set the component size in wrapperStyles when drag wrapper is enabled', () => {
+    it.only('should NOT set the component size in wrapperStyles when drag wrapper is enabled', () => {
```
Chaoste, line 298:
```suggestion
    it('should NOT set the component size in wrapperStyles when drag wrapper is enabled', () => {
```

Chaoste (approved):
Nice work! Is there a reason to not add `display: none` directly to the wrapper? Is it maybe causing warning/errors with pangea DND when the wrapper is missing in the DOM?

PR #1078, opened 2025-04-03
feat: allow custom fetching for caching purposes [SPA-2706] [ZEND-6205]

Chaoste:
## Purpose

Customers want to reduce CDA requests and use their own caching layer. In one example, a user wants to fetch the content for all locales at once and then select a specific locale when rendering a page.

## Approach

To allow such custom fetching logic, we need to expose more of our core fetching logic with `fetchReferencedEntities` and `fetchExperienceEntry`. To call them manually but still ensure the right format for `ExperienceRoot`, there is a new shallow wrapper `useCustomFetch` that accepts any fetching function but ensures that it won't run in editor mode and provide loading state and error messages.

To support caching multi locales data, the related fetcher functions can handle the case of the locale being undefined.

I added an MD document to provide an example usage of this new interface.


docs/guides/caching.md
```diff
@@ -0,0 +1,117 @@
+# Caching
+
+When using the recommended approach for initializing experiences with this SDK, the logic will take fetch the experience entry, other entries, assets, and patterns under the hood and return the final experience object.
```
anwaar931, line 3: I think there is a typo. Maybe you meant, `the login will fetch`
```
the logic will take fetch the experience entry, other entries
```

anwaar931, line 3: [Optional] Also would be nice if we can be explicit about `other entries, assets`. That what these other entries, assets are. So maybe we can mention something about bound entries and assets.

docs/guides/caching.md
```diff
@@ -0,0 +1,117 @@
...
+    client,
+    experienceTypeId,
+    identifier: { slug },
+    locale: localeCode,
+  });
+  const references = await fetchReferencedEntities({
+    client,
+    experienceEntry,
+    locale: localeCode,
+  });
```
anwaar931, line 38: These references also include entries in `usedComponents`? I think so but just confirming.

Chaoste: Used Components are fetched together with the experience as we're using `include=3` in there. This one solely cares about dataSource.

packages/core/src/fetchers/index.ts
```diff
@@ -2,3 +2,6 @@ export * from './fetchBySlug';
 export * from './fetchById';
 export * from './fetchAllEntities';
 export * from './createExperience';
+// Exposing the underlying fetcher methods to allow caching of API responses
+export * from './fetchReferencedEntities';
+export * from './fetchExperienceEntry';
```
SofiaMargariti, line 7: You will need to re-export these from `packages/experience-builder-sdk/src/index.ts`, otherwise they are not really exposed to the end-user of the SDK.

packages/core/src/fetchers/fetchExperienceEntry.ts
```diff
@@ -3,7 +3,9 @@ import { ContentfulClientApi, Entry } from 'contentful';
 type fetchExperienceEntryArgs = {
```
SofiaMargariti, line 3: Now we're exposing these methods to the developers I think we should make sure each of the parameters and the function itself include a doc string so that they don't have to reference external docs to understand what each function does.

docs/guides/caching.md
```diff
@@ -0,0 +1,119 @@
...
+
+To cache the loaded entities and inject them manually, a custom fetcher function needs to be written that uses `fetchExperienceEntry`, `fetchReferencedEntities`, and `createExperience` which are exposed from the SDK as well.
+
+### Fetching Data
+
+```ts
+function customFetchAllEntities({ client, experienceTypeId, slug, localeCode }) {
+  // Fetch the experience entry including its used patterns (resolved links in fields.usedComponents)
+  const experienceEntry = await fetchExperienceEntry({
+    client,
```
SofiaMargariti, line 30: I've been trying this snippet with the `withAllLocales` modifier and I get a type error in this line:

Screenshot 2025-04-03 at 17 21 12

We probably need to adjust the types for the `client` but I am not sure if that's gonna have any other side-effects. If I understand correctly the response format with the modifier is different than the plain one, is that right?

packages/core/src/fetchers/fetchExperienceEntry.ts
Chaoste: Good point, I added those now directly on the function as the types per parameter are not readable for SDK users.

docs/guides/caching.md
Chaoste: Thanks for calling out. I resolved this by using a UNION type on the params. I'm also using this type now in the types to ensure that this is supported by the types.

docs/guides/caching.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Caching & Injecting Data
+
+To cache the loaded entities and inject them manually, a custom fetcher function needs to be written that uses `fetchExperienceEntry`, `fetchReferencedEntities`, and `createExperience` which are exposed from the SDK as well.
+
+### Fetching Data
+
+```ts
+import {
+  fetchExperienceEntry,
```
DanweDE, line 30: It doesn't have to be in this PR, but generally, all the functions we export from the SDK and refer to in guides/documentation should have top-notch Doxygen-style comments to guide developers without leaving the IDE. I think this isn't the case for some of these imports right now.

docs/guides/caching.md
```diff
@@ -0,0 +1,137 @@
...
+    experienceTypeId,
+    identifier: { slug },
+  });
+  const referencesLocales = await fetchReferencedEntities({
+    client: clientWithAllLocales,
+    experienceEntry: experienceEntryLocales,
+  });
+
+  // Localize cached entities
+  const experienceEntry = localizeEntry(experienceEntryLocales, locale);
```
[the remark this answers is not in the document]
Chaoste: So far, we don't provide this but the mentioned customer seems to have a custom logic for this already. I added a note to the document that points out that this is not provided by CF (yet) and what that function is supposed to do.

docs/guides/caching.md
Chaoste: Yes, that's a good call-out. Would be great to prepare one or two perfect examples that we aligned on, so it's clear what requirements those should match.
I added docstrings for all exported functions that are used in here, though I'm not sure what level of detail they should have.

packages/core/src/fetchers/fetchExperienceEntry.ts
```diff
@@ -1,48 +1,58 @@
 import { ContentfulClientApi, Entry } from 'contentful';
 
-type fetchExperienceEntryArgs = {
-  client: ContentfulClientApi<undefined>;
+type ClientAndLocaleParams =
+  | {
+      client: ContentfulClientApi<undefined>;
+      locale?: string;
```
SofiaMargariti, line 6:
```suggestion
      locale: string;
```
Does this need to be required in this case?

packages/core/src/utils/localizeEntity.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { describe, it, expect } from 'vitest';
+import { localizeEntity } from './localizeEntity';
+import { Entry } from 'contentful';
+
+// filepath: /Users/thomas.kellermeier/work/sparks/experience-builder/packages/core/src/utils/localizeEntity.test.ts
```
SofiaMargariti, line 5: Is this leftover comment or do we need it?

SofiaMargariti (approved):
Looks good, thank you making the adjustments and updating the .md file.

I only had a type related question

packages/core/src/fetchers/fetchReferencedEntities.ts
```diff
@@ -5,12 +5,29 @@ import { DeepReference, gatherDeepReferencesFromExperienceEntry } from '@/deep-b
 import { gatherAutoFetchedReferentsFromIncludes } from './gatherAutoFetchedReferentsFromIncludes';
 import { fetchAllEntries, fetchAllAssets } from './fetchAllEntities';
 
+type ClientAndLocaleParams =
+  | {
+      client: ContentfulClientApi<undefined>;
+      locale?: string;
```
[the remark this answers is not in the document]
Chaoste: When it's a regular client without the modifier, you could still leave it out and not specify it on the API request. This way, it will automatically fetch the default locale code. I looked that up in our documentation. Personally, I think it's nice if you need less to define to get to a functioning flow.

packages/core/src/utils/localizeEntity.spec.ts
Chaoste: Good catch! Don't know how that ended up in here. I'll delete it.

dimitrycf:
image

PR #1080, opened 2025-04-04
fix: pattern wrapper height to 100% only when the main component has 100% height [SPA-2580]

Spring3:
## Purpose

I recorded the video with the explanation, but as it's a public repository, I won't attach it here.

Find it [here](https://contentful.slack.com/archives/C01R9P3CBH7/p1743760799219389?thread_ts=1740556729.913469&cid=C01R9P3CBH7) instead

## Approach

Basically if the pattern's wrapper component doesn't have the `100%` height, then the dropzone component in the editor mode also shouldn't have it.


Chaoste (approved):
Nice fix, thanks for sharing the video with explanation :)

PR #1081, opened 2025-04-04
feat(test-apps/react-vite): add new component with Link field for testing [SPA-2692]

Chaoste:
## Purpose

So far, we had no example where a component would render content from a content that is linked through a reference field of an entry.
When using the type `Link` for a variable definition, it allows binding to the whole reference field so the rendering code can pick up any fields from that referenced entity.

We adjusted the fetching logic in the user interface recently for that use case and now add an example use case.

https://github.com/user-attachments/assets/85fad642-8de4-43fe-a1ce-b2f3e62364c7


SofiaMargariti (approved):
Looks good!

PR #1093, opened 2025-04-17
fix: resolve background image optimization in SSR [SPA-2698]

anwaar931:
## Purpose
We are adding background image optimization support in SSR, it was only supported on the CSR side. Background image optimization will add `image-set` in the CSS. 

This PR is focused on adding this support and fixing and adding new tests.

### DIFF
--- a/packages/core/src/utils/styleUtils/ssrStyles.ts
+++ b/packages/core/src/utils/styleUtils/ssrStyles.ts
@@ -26,7 +26,6 @@ import {
   BackgroundImageOptions,
 } from '@/types';
 import { CF_STYLE_ATTRIBUTES, SUPPORTED_IMAGE_FORMATS } from '@/constants';
-import { generateRandomId } from '@/utils';
 import { stringifyCssProperties } from './stylesUtils';
 import { getOptimizedBackgroundImageAsset } from '../transformers/media/getOptimizedBackgroundImageAsset';
 import { AssetDetails, AssetFile } from 'contentful';
--- a/packages/core/src/utils/styleUtils/ssrStyles.ts
+++ b/packages/core/src/utils/styleUtils/ssrStyles.ts
@@ -36,6 +36,11 @@ type FlattenedDesignTokens = Record<
   string | { width?: string; style?: string; color?: string }
 >;
 
+type QueueItem = {
+  node: ComponentTreeNode;
+  parentChain: string[];
+};
+
 export const detachExperienceStyles = (experience: Experience): string | undefined => {
   const experienceTreeRoot = experience.entityStore?.experienceEntryFields
     ?.componentTree as ExperienceComponentTree;
@@ -72,7 +77,7 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
     componentVariablesOverwrites,
     patternWrapper,
     wrappingPatternIds,
-    patternNodeIdsChain = '',
+    parentChainArr = [],
   }: {
     componentTree: ExperienceComponentTree;
     dataSource: ExperienceDataSource;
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -427,13 +460,42 @@ export const maybePopulateDesignTokenValue = (
...
+      asset.fields.file as AssetFile,
+      width as string,
+      options.quality,
+      options.format as (typeof SUPPORTED_IMAGE_FORMATS)[number],
+    );
+    return value;
+  } catch (error) {
+    console.error('Error transforming image asset', error);
+  }
+  return boundAsset.fields.file?.url as string;
```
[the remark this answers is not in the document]
anwaar931: We don't want to throw an error here, if the image Asset cannot be optimized, we will simply return boundAssed url to not break the production website.

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -67,23 +85,33 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
...
     patternWrapper?: ComponentTreeNode;
     wrappingPatternIds: Set<string>;
-    patternNodeIdsChain?: string;
+    parentChainArr?: string[];
   }) => {
     // traversing the tree
-    const queue: ComponentTreeNode[] = [];
-
-    queue.push(...componentTree.children);
+    const queue: QueueItem[] = [];
```
anwaar931, line 91: Changed the queue structure because we want to keep the parent chain. This is added to resolve correct nodeIds chain. Currently Pattern and Nested Pattern SSR styles are not applying correctly to the node while rendering in `CompositionBlock`.

e.g Create a container in an Experience and add a pattern inside that container, You may notice that SSR styles are not applied to pattern nodes.
In the image below, inside <`assembly`> div, the styles are `cfstyles` and not `cf-hash` which means that styles for pattern when wrapped in a container(s) are not applied on SSR.

image

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -596,13 +692,36 @@ export const indexByBreakpoint = ({
...
+        defaultBreakpoint,
+        componentSettings,
+        componentVariablesOverwrites,
+      }) as BackgroundImageOptions;
+
+      if (!options) {
+        console.error(
+          `Error transforming image asset: Required variable [cfBackgroundImageOptions] missing from component definition`,
+        );
+        continue;
```
maxcheremisin, line 713: this looks suspicious tbh, don't we want to return a plain url if there is no options to optimize the image based on?

packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -596,13 +692,36 @@ export const indexByBreakpoint = ({
...
+        componentSettings,
+        componentVariablesOverwrites,
+      }) as string;
+
+      const options = resolveImageOptimizationOptions({
+        variableData: variables['cfBackgroundImageOptions'],
+        defaultBreakpoint,
+        componentSettings,
+        componentVariablesOverwrites,
+      }) as BackgroundImageOptions;
```
maxcheremisin, line 707: these typecasts are looking error prone, especially given that's the same function

maxcheremisin: maybe separate functions, where each one should return a static type?

packages/core/src/utils/styleUtils/ssrStyles.ts
maxcheremisin: Sorry, not quite following. It looks like we used to resolve a plain url in ssr, but now if there is no options we return nothing. Why not keeping a plain url for this case?

maxcheremisin: So you saying that this shouldn't be the case, and the options should always be provided? that's fine with me 👍 thanks for providing the context

anwaar931: Yes, e.g Currently on cavas currently if there is not `BackgroundImageOptions`, the image will not be rendered in the canvas, but then on SSR side ( on preview ) if we return a plain image url they can get confused and could think it as a bug on Canvas.

packages/core/src/utils/styleUtils/ssrStyles.ts
anwaar931: Yeah, most of the logic in the function is same. So wanted it to be a single function.

maxcheremisin: ok then maybe at least rename it to something like `resolveVariable` since it's a generic function

anwaar931:
@Spring3 tested the changes with isolated pattern and it works.

PR #1094, opened 2025-04-17
fix: support for resolving L3 helper [SPA-2697]

dimitrycf:
USE THIS https://github.com/contentful/experience-builder/pull/1100

## Purpose

This PR implements approach which shows how shallow references (eg. of `Array`) can be resolved by the component.


dimitrycf:
Closed in favor of https://github.com/contentful/experience-builder/pull/1100

PR #1100, opened 2025-04-22
feat: manually resolving links [SPA-2697]

Spring3:


dimitrycf:
Seems that typings do not allow to have union type of `'Entry'|'Asset'` despite it being declared explicitly in the source

image


image

dimitrycf:
### Also if customer developers are going to be resolving L3 manually,  we need to document also how to resolve links of Arrays

image

packages/core/src/utils/transformers/getArrayValue.ts
```diff
@@ -28,19 +28,8 @@ export function getArrayValue(
     } else if (value?.sys?.type === 'Link') {
       const resolvedEntity = entityStore.getEntityFromLink(value);
       if (!resolvedEntity) {
-        return;
+        return value;
```
dimitrycf, line 31: i guess this was the line which was responsible for why there was an array of undefined's? 
I wonder, now, if we now return links, should the customers be aware that sometimes they will get links.. but those links are actually "unresolvable" links... 😕  (and they may want to try to resolve them again).
Maybe returning `undefined` wasn't such a bad choice?

dimitrycf: i wonder if this also can happen in case a link is not publish or smth like that? 

(in this case customer attempt to manually resolve also won't suceed)

dimitrycf:
### When using this `inMemory.maybeResolve` when resolving `Array` type variable, it's important to map over, and not mutate in place those entities..


When within custom component we use:

```
// Keep in mind that `potions` array will have referential integrity and also will be shallow
// however, if instead `potions` is mutated in place, then links will be resolved to entities and repeated resolution may not work as expected if it only expects links
const potions = shallowPotions.map((p) => resolveLinksUpToLevel3(p));
```

packages/core/src/utils/transformers/getArrayValue.ts
dimitrycf: Because react can rerender few times before all the entities are synced up, it shows them as links 

image

PR #1106, opened 2025-04-24
fix: remove unused variable from useSingleColumn hook [SPA-2738]

Chaoste:
## Purpose

When testing columns wrap count, I first assumed this was handled in the SDK and debugged the code in `useSingleColumn` (the only place referring this variable). Only after some log debugging, I figured this is solely handled in the web app by setting the `columnsSpan` accordingly so there was no work necessary in the SDK.

However, I found that we run some logic in here for every node that is never used which is why I removed that part form the hook.

### DIFF
--- a/packages/core/src/utils/styleUtils/styleTransformers.ts
+++ b/packages/core/src/utils/styleUtils/styleTransformers.ts
@@ -16,6 +16,7 @@ export const transformVisibility = (value?: boolean): CSSProperties => {
   return {};
 };
 
+// TODO: Remove in next major version v2 since the change is 17 months old
 // Keep this for backwards compatibility - deleting this would be a breaking change
 // because existing components on a users experience will have the width value as fill
 // rather than 100%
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/styleUtils/styleTransformers.ts
```diff
@@ -16,6 +16,7 @@ export const transformVisibility = (value?: boolean): CSSProperties => {
   return {};
 };
 
+// TODO: Remove in next major version v2 since the change is 17 months old
```
SofiaMargariti, line 19: I think the change is 12 months old, not 17.

Screenshot 2025-04-25 at 10 28 57

This is a tricky scenario, it would only impact any component dropped before Apr 2024 that is still using the default value for those styles :/ Probably a low probability of that being the case, but we should explain that in the release notes of the major version.

SofiaMargariti (approved):

packages/core/src/utils/styleUtils/styleTransformers.ts
Chaoste: Good call. I will mention that in the release notes.

PR #1111, opened 2025-04-25
chore: remove deprecated functions for upcoming major SDK release [SPA-2741]

Chaoste:
- [x] Merge this only when everything is ready and merge it into the other release branch.


SofiaMargariti:
I think all of these commits should be `chore!:` to trigger a major version and to be picked up as breaking changes in the release changelog. Also we shouldn't squash when we merge them into development.

packages/core/src/types.ts
```diff
@@ -111,6 +111,9 @@ export type ComponentDefinition<
   hyperlinkPattern?: string;
   variables: Partial<DesignVariableMap> & Record<string, ComponentDefinitionVariable<T>>;
   slots?: Record<string, { displayName: string }>;
+  // FIXME: While it's technically possible, we don't want to allow using built-in content props via
```
SofiaMargariti, line 114: Are you referring to `cfImageAsset` and `cfBackgroundImageUrl` here? Why we don't want to allow this?

SofiaMargariti (approved):

packages/core/src/types.ts
Chaoste: Mostly for the same reason as why we omit `cfHyperlink` and `cfOpenInNewTab`. This configuration is called `builtInStyles` while it's actually a mix of all built-in properties and not only about design/ styles.

Also note that, while the types omit two of the content properties already, those are still not excluded logically IIRC, i.e. if you just skip the types, you can still use `cfHyperlink` and `cfOpenInNewTab` through this configuration option.

SofiaMargariti: Yes, the naming is misleading here.
My question was mainly about whether there is any benefit of allowing users to use those attributes in which case we should probably adjust the naming (or expose them from a separate attribute) **OR** if there is no reason a user to ever use these in which case we need to make the validations stricter.
Nothing we need to change right now.. just thinking what is the right handling of those attributes in the long term.

PR #1115, opened 2025-05-02
feat: early prefetching L4 and deeper

dimitrycf:
## Purpose

Provides utilities to "early prefetch" arbitrary number of links


dimitrycf:
this was brainstorming PoC. Closing.

PR #1117, opened 2025-05-06
fix: merge breakpoint defaults visual editor [SPA-2602]

anwaar931:
## Purpose
Merge default and current design values for pattern block type nodes.


packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -330,6 +331,12 @@ const injectDefaultValuesForComponentValues = ({
   const propertyDefinitions = wrapperComponentSettings?.variableDefinitions;
   const resolvedProperties = Object.entries(patternNode.variables).reduce(
     (resolvedProperties, [propertyName, propertyValue]) => {
+      if (propertyValue.type === 'DesignValue') {
+        console.log('TK', {
+          propertyName,
+          propertyValue,
+        });
+      }
```
Chaoste, line 339:
```suggestion
```
My bad, that was a left over from my initial branch :)

packages/visual-editor/src/hooks/useComponentProps.ts
```diff
@@ -357,3 +375,51 @@ const addExtraDropzonePadding = (padding: string) =>
...
+}: {
+  findNodeById: (nodeId?: string) => ExperienceTreeNode | null;
+  variableName: string;
+  variableMapping: DesignValue;
+  node: ExperienceTreeNode;
+}) => {
+  if (node.type === ASSEMBLY_BLOCK_NODE_TYPE) {
+    const patternId = node.data.pattern?.id;
+
+    const exposedProperyName = node['exposedPropertyNameToKeyMap'][variableName];
```
Chaoste, line 393:
```suggestion
    const exposedPropertyName = node['exposedPropertyNameToKeyMap'][variableName];
```

packages/visual-editor/src/hooks/useComponentProps.ts
```diff
@@ -99,18 +103,31 @@ export const useComponentProps = ({
...
 
+        /* variableMapping {
+          type: 'DesignValue',
+          valuesByBreakpoint: {
+            desktop: '300px',
+            // tablet: '300px',
+            // mobile: '300px',
+          },
+        }
+        */
```
Chaoste, line 122: Was this added accidentally for debugging purposes or meant for documenting the shape of `variableMapping`? If it's the latter, would it help to have proper types instead?

packages/visual-editor/src/hooks/useComponentProps.ts
```diff
@@ -357,3 +375,51 @@ const addExtraDropzonePadding = (padding: string) =>
...
+      const parentPatternId = parentPatternNode.data.pattern?.id;
+      const nextKey = parentPatternNode['exposedPropertyNameToKeyMap'][exposedProperyName];
+
+      if (!parentPatternId || !nextKey) {
+        break;
+      }
+      const parentPatternVariableDefinition =
+        componentRegistry.get(parentPatternId)?.definition.variables[nextKey];
+
+      exposedDefaultValue = parentPatternVariableDefinition?.defaultValue;
```
Chaoste, line 414: Might be worth noting down (in a comment) that with this logic, we're not merging defaults from different nesting layers for cases like "upper pattern specifies mobile and nested pattern specifies tablet".

Chaoste: Should we handle the cases differently if there was no `defaultValue` for the property or are we populating the defaultValue on every level anyway in the web app?
If we didn't and the upper pattern had no default, we would not fall back to the nested definition.

Chaoste (changes requested):

packages/visual-editor/src/hooks/useComponentProps.ts
```diff
@@ -357,3 +375,51 @@ const addExtraDropzonePadding = (padding: string) =>
 
 const isPercentValue = (value?: string | number) =>
   typeof value === 'string' && value.endsWith('%');
+
+const calculateDesignVariableValue = ({
```
Chaoste, line 379: With no code comment, it might not be easy for other devs to judge what this function is actually doing. Can you think of a name that somehow represents its logic of looking up the default value from its wrapping pattern definition? Example `maybeMergePatternDefaultDesign`? (I know, very clumsy name ^^)

SofiaMargariti (commented):
I left some small comments.
One thing I wanted to point out is that I would consider this a breaking change if it affects the preview/delivery mode because if some experiences currently have breakpoint specific overwrites the styles will change when the current fix is released.
(The ticket mentions that delivery mode is already fixed but when I tested it I could see the original issue there as well - something to clarify when @Chaoste is back)

packages/visual-editor/src/hooks/useComponentProps.ts
anwaar931: This was for testing and deubugging purpose.

packages/visual-editor/src/utils/patternUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+    const exposedPropertyName = node['exposedPropertyNameToKeyMap'][variableName];
+    if (!exposedPropertyName || !patternId) {
+      return variableMapping.valuesByBreakpoint;
+    }
+
+    const exposedVariableDefinition =
+      componentRegistry.get(patternId)?.definition.variables[exposedPropertyName];
+
+    let exposedDefaultValue = exposedVariableDefinition?.defaultValue as DesignValue;
+    let parentPatternNode = findNodeById(node.data.pattern?.nodeId);
```
Chaoste, line 35: I think it's not directly blocking but theoretically, this can cause quite a high computational load. For every variable of every node in a pattern, it will run this search algorithm and scan the whole tree to find the pattern node.

While testing & debugging the preview mode, I discovered that over there we're replacing `ComponentValue` as we discover them on the spot while "saturating" pattern instances. This way, the design values are merged step by step for nested patterns saving me from implementing this recursive lookup. So to say, preview mode uses a "top-to-bottom" approach that collects the required information while walking through the tree for rendering.
Here, it's kind of a "bottom-to-top" approach to look up similar information all at once when resolving it finally. Not sure if it's feasible to do a similar efficient logic for the editor mode as we have it in preview mode. Wanted to share my discovery anyway in case this sparks any ideas on your side :)

packages/visual-editor/src/utils/patternUtils.ts
Chaoste, line 1: [nit] I suggest naming this file `maybeMergePatternDefaultDesignValues.ts` to save this in the future from becoming a collection of many functions that are not related to each other.

packages/visual-editor/src/utils/patternUtils.ts
Chaoste: A quicker solution to reduce at least the computational load, could be to create a look up table in the tree store that stores the pattern nodes once and then just looks them up by ID.

Chaoste (approved):
Tested it and seems to work nicely. I shared another bigger suggestion in the comments but it could require a larger refactoring, so it's fine for me to leave it out of the scope of this PR. Before merging, let's maybe wait for @SofiaMargariti to confirm as well in case I missed something during testing

packages/experience-builder-sdk/src/core/sdkFeatures.ts
```diff
@@ -4,4 +4,5 @@ export const sdkFeatures: Record<string, unknown> = {
   hasVersionHistory: true,
   cfVisibility: true,
   patternResolution: true,
+  patternBreakpointDesignValues: true,
```
Chaoste, line 7: Just by the name, it might be harder to understand for other devs what this is for.
```suggestion
  // [SPA-2602] allow merging the default and overwriting design values for patterns by breakpoint
  patternBreakpointDesignValues: true,
```

Chaoste:
⚠️ For visibility:
We reverted the merge into `development` through a **force push** overwriting the commit history (didn't want to pollute the history and the changelog with this).

As it's introducing breaking change in the way we render designs, it's a breaking change that will be merged into `v2-prerelease`:
https://github.com/contentful/experience-builder/pull/1117

PR #1120, opened 2025-05-13
fix: merge two pattern node id chains for CSS classes [SPA-2767]

Chaoste:
## Purpose

We implemented [this fix](https://github.com/contentful/experience-builder/pull/982) at the same time as when the prebinding solution was added with [this PR](https://github.com/contentful/experience-builder/pull/1060). Both add a similar logic of chaining pattern node IDs to ensure no conflicts between two similar nested patterns with different design/ content values (incl. different CSS as a result).

## Approach

Along the way I noticed that
- `patternNodeIdsChain` is currently a chain of all node Ids (not only pattern nodes) - it's only added for the pattern use case, thus the name
- `patternRootNodeIdsChain` is a chain of all wrapping pattern node Ids

We don't need these two concepts to compose unique node IDs in cases of nested patterns. In both cases, we can use the same approach, following the second shorter chain.


packages/core/src/utils/styleUtils/ssrStyles.ts
```diff
@@ -85,33 +80,20 @@ export const detachExperienceStyles = (experience: Experience): string | undefin
...
-        break;
-      }
-      const { node: currentNode, parentChain } = queueItem;
-
+      const currentNode = queue.shift();
       if (!currentNode) {
         break;
       }
 
-      const currentNodeParentChain = [...parentChain, currentNode.id || ''];
```
Chaoste, line 112: Previously, this was a long chain of every single node ID in the hierarchy. Instead of adding the ID on every level we only add those required for the uniqueness check, i.e. prefix the storage key with pattern node IDs only.

Before:
`section1-container1-pattern1-section2-pattern2-section3-container3-heading1`
After:
`pattern1-pattern2-heading1`

Spring3: Has it been tested with the pre-binding scenario?

Currently it relies on the path of ids as breadcrumbs as a way to reach the component from what I understood

We want to change that, but for now let's make sure it works with the current approach without any issues

Spring3: @anwaar931 this won't break anything related to the style propagation for nested patterns, will it?

anwaar931: @Chaoste 
image

It looks like there is some issue. We are not generating ssr styles for pattern nodes.

anwaar931: image
This is how my experience looks like

anwaar931 (changes requested):
I added these comments here: https://github.com/contentful/experience-builder/pull/1120#discussion_r2091039735

Let me know if you need more context on how to reproduce this.

anwaar931 (approved):
Tested it with latest changes and its looking good. Thanks @Chaoste

PR #1122, opened 2025-05-13
fix: remove style nodes from the DOM in the cleanup function of the injection hook [SPA-2765]

Chaoste:
## Purpose
When rendering your app with the SDK on the client side, there are many `<styles>` nodes added to the DOM with the exact same CSS code.

## Approach
The hook for creating the `<styles>` nodes did not clean them up. So every time the insertion effect is fired, it would add another `<styles>` node to the DOM. The clean up is added by this PR.

I noticed that `useClassName` is an old left over that is not used anymore -> removed the hook & its tests, and renamed the file.


SofiaMargariti (approved):
Works as expected!

PR #1126, opened 2025-05-15
style: simplify resolveDesignValue by dropping the unused name arg [SPA-2610]

Chaoste:
## Purpose
I noticed that `resolveDesignValue` required the `variableName` parameter while that was only used in very special cases, i.e. when it's a built-in property that supports design token. 90% of the cases passed explicitly a string that didn't match that condition. In those cases, we can also leave out the ignored parameter.

## Approach
Make the `propertyName` argument optional and adjust all the places that passed a name that had no effect.

### DIFF
--- a/packages/core/src/utils/breakpoints.ts
+++ b/packages/core/src/utils/breakpoints.ts
@@ -102,13 +102,17 @@ export const getValueForBreakpoint = (
   breakpoints: Breakpoint[],
   activeBreakpointIndex: number,
   fallbackBreakpointIndex: number,
-  variableName: string,
-  resolveDesignTokens = true,
+  /** Provide the name for built-in styles to replace design tokens. Supported properties are:
+   * cfMargin, cfPadding, cfGap, cfWidth, cfHeight, cfBackgroundColor,
+   * cfBorder, cfBorderRadius, cfFontSize, cfLineHeight, cfLetterSpacing,
+   * cfTextColor, cfMaxWidth
+   */
+  propertyName?: string,
 ) => {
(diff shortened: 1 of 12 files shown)


packages/core/src/utils/breakpoints.ts
```diff
@@ -102,13 +102,22 @@ export const getValueForBreakpoint = (
   breakpoints: Breakpoint[],
   activeBreakpointIndex: number,
   fallbackBreakpointIndex: number,
-  variableName: string,
+  /** Provide the name for built-in styles to replace design tokens. Supported properties are:
+   * cfMargin, cfPadding, cfGap, cfWidth, cfHeight, cfBackgroundColor,
+   * cfBorder, cfBorderRadius, cfFontSize, cfLineHeight, cfLetterSpacing,
+   * cfTextColor, cfMaxWidth
```
anwaar931, line 108: Not sure, if we should list all of these supported properties. Maybe we can point to `core/src/types.ts/StyleProps`

Chaoste: That would be not correct as it's a type of all style props. This function is limited to a subset of it which is defined 20 lines above. Unfortunately, this list is not typed but just a `string[]` (probably because the type would not be used anywhere).

As it helped me to spot all the places using it the wrong way, I foudn it actually very helpful to show the list directly in the JSDoc as you will also see this when hovering over the function from another file when you want to use it.

Since the actual list and this comment are close to each other, I would prefer it this way if that's fine with you.

anwaar931 (approved):

PR #1132, opened 2025-05-20
SPA-2753: split schema between experience and patterns

kizer-cf:
## Purpose
Split the validation schema for experience and pattern such that each can be validated independently.

## Approach
Effort has been made to avoid any breaking changes. This includes re-exporting things from their original location and making the experiences validator call the pattern validator if it detects a pattern. Any potential breaking change found during review (including typing) should be flagged.

A common file has been created to reuse common schemas and reduce duplication.

## Ticket
https://contentful.atlassian.net/browse/SPA-2753

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -357,6 +357,8 @@ export type ExperienceEntry = {
   metadata: Entry['metadata'];
 };
 
+export type { Entry } from 'contentful';
+
 export interface RawCoordinates {
   left: number;
   top: number;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,6 +1,6 @@
 import type { Asset, AssetFile, ChainModifiers, Entry } from 'contentful';
 import { SCROLL_STATES, OUTGOING_EVENTS, INCOMING_EVENTS, INTERNAL_EVENTS } from '@/constants';
-import { EntityStore } from './entity/EntityStore';
+import { EntityStore } from '@/entity';
 import { Document as RichTextDocument } from '@contentful/rich-text-types';
 
 // Types for experience entry fields (as fetched in the API) are inferred by Zod schema in `@contentful/experiences-validators`
@@ -356,8 +356,6 @@ export type ExperienceEntry = {
   fields: ExperienceFields;
   metadata: Entry['metadata'];
 };
-
-export type { Entry } from 'contentful';
 
 export interface RawCoordinates {
   left: number;
--- a/packages/validators/src/schemas/index.ts
+++ b/packages/validators/src/schemas/index.ts
@@ -1,10 +1,10 @@
 export {
   BindingSourceTypeEnum,
-  VariableMapping,
   ExperienceFieldsCMAShapeSchema as ExperienceSchema_2023_09_28,
 } from './v2023_09_28/experience';
 export {
   PatternFieldsCMAShapeSchema as PatternSchema_2023_09_28,
   type PatternPropertyDefinition,
+  VariableMapping,
 } from './v2023_09_28/pattern';
 export { ComponentDefinitionSchema } from './componentDefinition';
--- a/packages/validators/src/schemas/index.ts
+++ b/packages/validators/src/schemas/index.ts
@@ -1,8 +1,10 @@
 export {
   BindingSourceTypeEnum,
-  PatternProperty,
-  PatternPropertyDefinition,
   VariableMapping,
-  ExperienceFieldsCMAShapeSchema as Schema_2023_09_28,
+  ExperienceFieldsCMAShapeSchema as ExperienceSchema_2023_09_28,
 } from './v2023_09_28/experience';
+export {
+  PatternFieldsCMAShapeSchema as PatternSchema_2023_09_28,
+  type PatternPropertyDefinition,
+} from './v2023_09_28/pattern';
 export { ComponentDefinitionSchema } from './componentDefinition';
--- a/packages/validators/src/schemas/v2023_09_28/experience.ts
+++ b/packages/validators/src/schemas/v2023_09_28/experience.ts
@@ -43,10 +43,3 @@ export type ComponentValue = z.infer<typeof ComponentValueSchema>;
 export type BindingSourceTypeEnum = z.infer<typeof BindingSourceTypeEnumSchema>;
 export type PatternProperty = z.infer<typeof PatternPropertySchema>;
 export { breakpointsRefinement };
-
-// Re-exports for compatibility. Remove when no longer needed
-export { PATTERN_THUMBNAIL_IDS } from './pattern'; // re-export for compatibility
-export type {
-  VariableMapping,
-  PatternComponentSettings as ExperienceComponentSettings,
-} from './pattern';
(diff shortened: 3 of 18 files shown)


SofiaMargariti:
Hey @kizer-cf, thank you for opening this PR. I think you'd need to rebase on development branch before we can properly review as now I see some of Anwar's changes in the diff. (Not sure why this happened, probably because development branch was force pushed at some point)
Let me know if you need any help with that :)

kizer-cf:
thanks @SofiaMargariti for pointing this out, was likely a botched rebase. I have now updated the branch to remove the problem.

Chaoste (commented):
It's a lot to review, so I'll share my first comments for today already.

packages/validators/src/schemas/index.ts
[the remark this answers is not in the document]
kizer-cf: We're not renaming any exports, only changing the locations, and in all cases re-exporting these from the original location. If anything is renamed, it is an error and should be flagged.

Once concern I have is how these are imported where they're used. For example, are imports always from the index file or do users also sometimes deep import (which will make a location change a potentially breaking change)? I am also not sure if this is used only in user_interface or also by customers (and to what extent), which will determine what we consider, and how we treat, potentially breaking changes.

packages/core/src/types.ts
```diff
@@ -357,6 +357,8 @@ export type ExperienceEntry = {
   metadata: Entry['metadata'];
 };
 
+export type { Entry } from 'contentful';
```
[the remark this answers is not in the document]
Chaoste: You're adding `contentful` to validators anyway, but this way it's implicit and more hidden. I don't see a big con in doing it, though I think we should favor explicit dependency over implicit ones if they don't cost us anything besides an additional line in the `package.json` file.

I see that both ways sound legit, so it's really your call here.

packages/validators/src/validators/validateExperienceFields.ts
```diff
@@ -1,28 +1,44 @@
...
  *
  * @param experience The experience entry to validate
- * @param schemaVersionOverride Optional override for the schema version to validate against. By default the schema version is read from the experience entry
+ * @param schemaVersionOverride Optional override for the schema version to validate against.
+ * By default, the schema version is read from the experience entry
  * @returns object with success property and optional errors array
  */
 export const validateExperienceFields = (
+  // TODO: type this as Entry when the type is exposed
   experience: any,
```
Chaoste, line 25:
```suggestion
  experience: Entry,
```
Is there a reason not using `Entry` from the core package right away in here?

Chaoste: Mmmh, do you know how we can type this actually correct for the CMA shape with the locales inside?
IIRC `Entry` itself types the CDA shape (already resolved for one locale) but maybe we can still get the CMA shape by use the `'WITH_ALL_LOCALES'` modifier which is a generic type of `Entry`.

Chaoste: Judging by its usage in `user_interface` ([search results](https://github.com/search?type=code&auto_enroll=true&q=org%3Acontentful%20validateExperienceFields)), the correct type for the CMA use case would be `EntryProps` from `contentful-management/types` but I guess we don't want to add this dependency to the mono repo just for this single type.

packages/validators/src/validators/validatePatternFields.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ *
+ * @param pattern The pattern entry to validate
+ * @param schemaVersionOverride Optional override for the schema version to validate against.
+ * By default, the schema version is read from the pattern entry
+ * @returns object with success property and optional errors array
+ */
+export const validatePatternFields = (
+  // TODO: type this as Entry when the type is exposed
+  pattern: any,
```
Chaoste, line 19: Same question here, why not type this already now? Or at least use `unknown`

Chaoste (changes requested):
Sorry for keeping you waiting until now. Just finished it and overall like the clear separation that you did.

packages/validators/src/schemas/v2023_09_28/experience.ts
```diff
@@ -346,6 +41,12 @@ export type UnboundValue = z.infer<typeof UnboundValueSchema>;
 export type HyperlinkValue = z.infer<typeof HyperlinkValueSchema>;
 export type ComponentValue = z.infer<typeof ComponentValueSchema>;
 export type BindingSourceTypeEnum = z.infer<typeof BindingSourceTypeEnumSchema>;
-export type PatternPropertyDefinition = z.infer<typeof PatternPropertyDefinitionSchema>;
 export type PatternProperty = z.infer<typeof PatternPropertySchema>;
-export type VariableMapping = z.infer<typeof VariableMappingSchema>;
+export { breakpointsRefinement };
+
+// Re-exports for compatibility. Remove when no longer needed
+export { PATTERN_THUMBNAIL_IDS } from './pattern'; // re-export for compatibility
```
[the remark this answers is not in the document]
kizer-cf: Removed the re-export

packages/validators/src/schemas/v2023_09_28/pattern.ts
```diff
@@ -0,0 +1,106 @@
...
+  ComponentTreeSchema,
+  ComponentVariableSchema,
+  DataSourceSchema,
+  localeWrapper,
+  propertyKeySchema,
+  UnboundValuesSchema,
+  UsedComponentsSchema,
+} from '../v2023_09_28/common';
+
+const THUMBNAIL_IDS = [
```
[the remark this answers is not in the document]
kizer-cf: There was no need to rename it in the first place. I have fixed it.

packages/validators/src/validators/validateExperienceFields.ts
kizer-cf: `contentful-management` was already indirectly installed. This also makes the `Entry` re-export redundant

packages/validators/src/validators/validatePatternFields.ts
kizer-cf: Using `unknown` would have made it really difficult to work with this in any capacity without casting it to the right type. I thought the correct type was `Entry`, but this was not being picked up by TS (it was looking in the already-published version, not the local version of `core`).

However, this is now mute as I have changed the type to `EntryProps`

packages/validators/src/validators/validateExperienceFields.ts
```diff
@@ -1,28 +1,44 @@
 import { type SchemaVersions } from '../types';
 import { ValidatorReturnValue } from './ValidatorReturnValue';
-import { Schema_2023_09_28 } from '../schemas';
-import { zodToContentfulError, CodeNames } from '../utils/zodToContentfulError';
+import { ExperienceSchema_2023_09_28 } from '../schemas';
+import { zodToContentfulError, CodeNames } from '@/utils/zodToContentfulError';
+import { validatePatternFields } from '@/validators/validatePatternFields';
+import type { EntryProps } from 'contentful-management';
```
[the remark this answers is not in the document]
kizer-cf: I've added it explicitly

Chaoste (approved):
Tahnks a lot for integrating all the feedback. It's look great and I can also share that I was wrong about the bundle size. It's really not affected by the imported type, so we're good to go :)

PR #1134, opened 2025-05-22
feat: introduce SDK capability dndMigration [SPA-2745]

Chaoste:
## Purpose
We will remove all dnd related code from the SDK. To allow a smooth transition, there is a new branch `v3-prerelease`. All dnd related code will be removed from the codebase in this branch.

## Approach
To indicate to the web app that this SDK version supports the new native dnd handling, a new SDK capability is added called `nativeDnd`. The name might not be self-explanatory, so I added a comment as well.


maxcheremisin:
let's discuss the naming, as 

1. not sure it is entirely native
2. the sdk doesn't have to know how dnd is implemented in the web

maxcheremisin (approved):

PR #1138, opened 2025-05-26
feat: remove dnd [SPA-2803]

Chaoste:
Precessor: https://github.com/contentful/experience-builder/pull/812/files

## Code inside visual-editor (excluding config & tests)
Before: 4917 LOC
After: 1884 LOC


packages/core/src/types.ts
```diff
@@ -704,3 +667,16 @@ export type IncomingMessage = {
...
+      editorMode: true;
+      node: ExperienceTreeNode;
+    }
+  | {
+      editorMode?: false;
+    };
+
+export type StructureComponentProps<OtherProps> = React.PropsWithChildren<
+  StructuralEditorModeProps & OtherProps
+>;
```
maxcheremisin, line 682: nit: not a big fan of these implicit generics, it could be just `StructuralEditorModeProps` and a specific component props could be defined as 

```tsx
type MyProps = StructuralEditorModeProps & {children: ReactNode, otherProp: unknown}
```

packages/visual-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -0,0 +1,121 @@
...
+type NodeBlockProps = {
+  node: ExperienceTreeNode;
+  resolveDesignValue: ResolveDesignValueType;
+  wrappingPatternIds?: Set<string>;
+};
+
+export function EditorBlock({
+  node,
+  resolveDesignValue,
+  wrappingPatternIds: parentWrappingPatternIds = new Set(),
```
maxcheremisin, line 26: on each rerender this default prop will be a `new Set()`, which will trigger recalculation of the `useMemo`, but maybe it's fine in this case

maxcheremisin (approved):
Nice job removing all of this 🔥

Just a couple of optional comments

packages/core/src/constants.ts
```diff
@@ -8,20 +9,19 @@ export const OUTGOING_EVENTS = {
   Connected: 'connected',
   DesignTokens: 'registerDesignTokens',
   RegisteredBreakpoints: 'registeredBreakpoints',
+  /** @deprecated will be removed when dropping backward compatibility for old DND */
```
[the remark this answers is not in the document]
Chaoste: I think we have to as they are used in the web app. So if we bump the version there to get new code from this branch, we will still have to expose the previous stuff.

PR #1144, opened 2025-05-30
feat(visual-editor): Send CanvasGeometryUpdated Event [SPA-2804]

Chaoste:
## Purpose

The editor needs to receive the bounding client rectangle of each node on the canvas, especially when anything changes.

## Approach

Create a hook that lives next to the component, triggering it (`RootRenderer`). We can change this again if we find a better spot. Happy for any suggestions but also fine with leaving it there for the moment.

Use MutationObserver, a resize event listener, and a manual invocation to cover use cases like the following:
- Node was added, edited, or deleted
- An image was loaded & rendered delayed
- The windows was resized or a breakpoint selected
- Browser zoom in/ out
- Patterns/ entries changed through a pubsub event ("live")


packages/visual-editor/src/hooks/useEditorSubscriber.ts
```diff
@@ -220,6 +222,7 @@ export function useEditorSubscriber() {
           }
           // Update the tree when all necessary data is fetched and ready for rendering.
           updateTree(tree);
+          manuallyFireCanvasGeometryUpdate(tree);
```
maxcheremisin, line 225: do we need to trigger it _manually_ since it's already subscribed to all dom changes?

packages/core/src/types.ts
```diff
@@ -561,8 +561,15 @@ type OUTGOING_EVENT_PAYLOADS = {
...
   sdkFeatures: SDKFeaturesPayload;
   REQUEST_ENTITIES: RequestEntitiesPayload;
+  canvasGeometryUpdated: {
+    size: { width: number; height: number };
+    nodes: Record<string, { coordinates: Pick<DOMRect, 'x' | 'y' | 'width' | 'height'> }>;
+    sourceEvent: CanvasGeometryUpdateSourceEvent;
+  };
 };
 
+export type CanvasGeometryUpdateSourceEvent = 'resize' | 'mutation' | 'manual';
```
maxcheremisin, line 571: tbh `manual` sounds a bit misleading, since we don't actually trigger it by hand. maybe something like `treeUpdate`, but i'm not even sure if we need to trigger it there, since it will still fire on dom changes

packages/visual-editor/src/components/RootRenderer/sendCanvasGeometryUpdatedMessage.ts
```diff
@@ -0,0 +1,80 @@
...
+  const nodeRecords: Record<
+    string,
+    { coordinates: Pick<DOMRect, 'x' | 'y' | 'width' | 'height'> }
+  > = {};
+  const scrollX = window.scrollX;
+  const scrollY = window.scrollY;
+
+  await waitForAllImagesToBeLoaded();
+
+  const getNodeCoordinates = (node: ExperienceTreeNode) => {
```
maxcheremisin, line 26: nit: maybe for readability it can be declared outside `sendCanvasGeometryUpdatedMessage` function

packages/visual-editor/src/hooks/useEditorSubscriber.ts
Chaoste: When I unbound the asset of an image node, it doesn't trigger a DOM update. I don't know why that is, but I thought it might be a good safety mechanism to enforce an update when the tree changed just to be safe. Since we debounce the updates, I don't expect a major performance issue by doing so.

We could spend more time later on why the described example doesn't trigger the update but at least, we now have a stable system to work with :)

packages/core/src/types.ts
Chaoste: You're right, the name is misleading. As this is more like a general "others" scenario, I'm using instead the `undefined` value to not give it an explicit label.

packages/visual-editor/src/hooks/useEditorSubscriber.ts
maxcheremisin: Interesting, I expected it to work differently. Is it possible that the dom doesn't change and we update the image using css only?

maxcheremisin: maybe call this function `sendCanvasGeometryUpdate` or something more generic?

packages/visual-editor/src/components/RootRenderer/useCanvasGeometryUpdates.ts
```diff
@@ -0,0 +1,71 @@
...
+      ),
+    [],
+  );
+
+  useEffect(() => {
+    const resizeEventListener = () => debouncedUpdateGeometry(tree, 'resize');
+    window.addEventListener('resize', resizeEventListener);
+
+    // In some cases like duplication of a node, this fires two mutation events which both are
+    // required. The final coordinates might be available at the first or the second invocation.
```
maxcheremisin, line 49: what this comment is related to?

packages/core/src/types.ts
Chaoste: We don't need it necessarily. I thought it might be helpful during debugging to surface the reason for SDK messages being fired. Would you prefer leaving it out?

packages/visual-editor/src/components/RootRenderer/useCanvasGeometryUpdates.ts
Chaoste: It relates to the mutation observer one line below. As it causes unexpected behavior, I wanted to highlight that in a comment next to it.

packages/visual-editor/src/components/RootRenderer/useCanvasGeometryUpdates.ts
```diff
@@ -0,0 +1,71 @@
...
+        (tree: ExperienceTree, sourceEvent: CanvasGeometryUpdateSourceEvent) => {
+          // When the DOM changed, we still need to wait for the next frame to ensure that
+          // rendering is complete (e.g. this is required when deleting a node).
+          window.requestAnimationFrame(() => {
+            sendCanvasGeometryUpdatedMessage(tree, sourceEvent);
+          });
+        },
+        100,
+        {
+          leading: true,
```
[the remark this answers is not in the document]
Chaoste: I'll add it for now but I would love to revisit this as it can be quite expensive to do the same calculation twice if we don't see the need for it. The tree changes often trigger multiple mutation events at once, even though it's just one DOM change (e.g., duplicating a container with child nodes).
In an ideal world, I would keep the amount of geometry update messages at a bare minimum instead of firing too many equal events.
We could extend the logic in here to save the last message and compare them - shouldn't be too expensive. Wdyt?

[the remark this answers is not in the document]
maxcheremisin: Agree, we should keep an eye on performance in the future, test it with extensive experiences and optimize if needed

packages/visual-editor/src/components/RootRenderer/useCanvasGeometryUpdates.ts
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  useEffect(() => {
+    const resizeEventListener = () => debouncedUpdateGeometry(tree, 'resize');
+    window.addEventListener('resize', resizeEventListener);
+
+    // In some cases like duplication of a node, this fires two mutation events which both are
+    // required. The final coordinates might be available at the first or the second invocation.
+    const observer = new MutationObserver(() => debouncedUpdateGeometry(tree, 'mutation'));
+    observer.observe(window.document.documentElement, {
```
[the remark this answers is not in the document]
Chaoste: Good idea to use the ref (which wasn't used anymore). It's now being passed through to this hook and used as target.

maxcheremisin (approved):
Neat 🚀

PR #1139, opened 2025-05-27
feat!: manually resolving links [SPA-2754] PR to merge into v2-prerelease branch

dimitrycf:
## Purpose

This PR is related to the original [PR ](https://github.com/contentful/experience-builder/pull/1129)is intended to merge the ticket instead of `development` branch into `v2-prerelease` branch

PR #1156, opened 2025-06-04
fix(core): deep binding resolution with reference field already being a resolve entity [SPA-2843]

Chaoste:
## Purpose

A customer noticed that assets stopped rendering when another component on the experience used the recently added L2 resolution of reference fields.

## Approach

As soon a registered component uses the `'Link'` variable type, the SDK receives entities with the reference fields not being links but actual entities already.
When resolving deep bindings (for assets), the entity store is now able to handle this scenario as well. I added a test that also showcases this scenario.

### DIFF
--- a/packages/core/src/entity/EntityStoreBase.ts
+++ b/packages/core/src/entity/EntityStoreBase.ts
@@ -203,7 +203,7 @@ export abstract class EntityStoreBase {
           }
           resolvedFieldset.push([entityToResolveFieldsFrom, field, _localeQualifier]);
           entityToResolveFieldsFrom = entity; // we move up
-        } else if (typeof fieldValue === 'object' && this.isAsset(fieldValue)) {
+        } else if (this.isAsset(fieldValue)) {
           resolvedFieldset.push([entityToResolveFieldsFrom, field, _localeQualifier]);
           entityToResolveFieldsFrom = fieldValue; // we move up
         }
@@ -242,8 +242,13 @@ export abstract class EntityStoreBase {
     return leafEntity;
   }
 
-  private isAsset(entity: Entry | Asset): entity is Asset {
-    return entity.sys.type === 'Asset';
+  private isAsset(value: unknown): value is Asset {
+    return (
+      null !== value &&
+      typeof value === 'object' &&
+      'sys' in value &&
+      (value as Asset).sys?.type === 'Asset'
+    );
   }
 
   private getEntity(type: 'Asset' | 'Entry', id: string) {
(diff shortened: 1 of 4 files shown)


packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -203,6 +203,9 @@ export abstract class EntityStoreBase {
           }
           resolvedFieldset.push([entityToResolveFieldsFrom, field, _localeQualifier]);
           entityToResolveFieldsFrom = entity; // we move up
+        } else if (typeof fieldValue === 'object' && this.isAsset(fieldValue)) {
```
dimitrycf, line 206: maybe instead of guarding from `string` by this `typeof fieldValue == 'object'` which is techincally a hack, we can simply replace implemenation of the `EntityStore#isAsset()` with:



```
export const isAsset = (value: unknown): value is Asset => {
  return (
    null !== value &&
    typeof value === 'object' &&
    'sys' in value &&
    (value as Asset).sys?.type === 'Asset'
  );
};
```

this implementation:
 - uses unknown and thus no type issues
 - runtime-wise it's also safer, as it doesn't assume any specific shape. (As original implemenation will throw if object doesn't have sys.type)

ps. I have already added `isAsset` as a utility (will be available as of SDK v2), but here for simplicity of merging and putting it live, I think it's ok and we can simply copy/paste this function.

https://github.com/contentful/experience-builder/pull/1139/files#diff-13d28f6e2e67b70d5c4b248c138f6875209bd3c491aa0ac35b0fe68861c9e633R26

packages/core/src/entity/EntityStoreBase.ts
```diff
@@ -203,6 +203,9 @@ export abstract class EntityStoreBase {
           }
           resolvedFieldset.push([entityToResolveFieldsFrom, field, _localeQualifier]);
           entityToResolveFieldsFrom = entity; // we move up
+        } else if (this.isAsset(fieldValue) || this.isEntry(fieldValue)) {
+          resolvedFieldset.push([entityToResolveFieldsFrom, field, _localeQualifier]);
+          entityToResolveFieldsFrom = fieldValue; // we move up
         }
```
dimitrycf, line 209: I think we need to add 
```
else { 
  return {
              resolvedFieldset,
              isFullyResolved: false,
              reason: `FSOME ERROR CODEink (sys.id=${fieldValue.sys.id}) not found in the EntityStore, waiting...`,
}

dimitrycf (approved):
Thanks for the changes! 🙇

PR #1161, opened 2025-06-06
fix(core): move typeof localStorage check into try catch block to support aggressive cookie settings in Brave [SPA-2838]

Chaoste:
## Purpose

When using the Brave browser, activating the deprecated flag `#block-all-cookies-toggle` and toggling on the legacy option "Block all cookies", a check like `typeof localStorage` will throw an error. While logic like `typeof asdf` will gracefully return `undefined`, the browser seems to even punish the this check with an error.

Screenshot 2025-06-06 at 13 33 09

Screenshot 2025-06-06 at 13 33 14

Screenshot 2025-06-06 at 13 33 31

**Background to this flag:**
It "broke the web" for users as the aggressive behavior would cause several issues on web page with Studio just being yet another anecdote that proves it.
- https://github.com/brave/brave-browser/issues/42061
- https://github.com/brave/brave-browser/wiki/Block-all-cookies-global-Shields-setting

## Approach
Moving the `typeof localStorage` check inside the try-catch-block will also handle this case gracefully from now on.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2507,6 +2507,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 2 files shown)


maxcheremisin (approved):
good finding

PR #1170, opened 2025-06-11
fix(core): throw error for invalid store objects [SPA-2862]

Chaoste:
## Purpose

We're often seeing this error: `Cannot read properties of undefined (reading 'type')`
I discovered two causes for this:
- Either the parent in `useSingleColumn` is not a "columns" node
- The entry/ asset in the EntityStore is neither an entry nor an asset (e.g. see ZEND-6498, caused by wrong JSON serialization logic)

## Approach
In both cases, the page crashes, leaving little information to investigate. As it already crashes for those unexpected situations, the logic catches those now more explicit and throws an explicit error instead. The main reason seems to be having objects in the store that are neither entries nor assets which will also be addressed by [this change](https://github.com/contentful/experience-builder/commit/556fd40f7fe5c100a7b8e534df5c9b29ed9295b4).

As they are not necessary for the store, I moved the class methods `isAsset` and `isEntry` into a separate util file.


SofiaMargariti (approved):
Looks good to me!

PR #1171, opened 2025-06-11
fix: remove unbound preview component [SPA-2789]

Chaoste:
Reducing the special casing in `CompositionBlock` and dropping the second image wrapper.

As we don't need `useMediaQuery` anymore but still its util functions, I moved those into separate files outside of the `hooks` folder.


Chaoste:
> Could you add some more context for this change? Why was the special casing needed before, but not anymore? Also some testing instructions would be nice, what scenarios to validate, etc.

In `development` branch, you'll see that we're passing `editorMode={false}` for those cases which is not used anymore in v3, so those components receive the same props as every other registered component.

IMO the only thing that needs to be tested is the rendering of image components with no asset bound. Previously, we dealt with it in two different places, which makes no sense - so that's why I'm consolidating/cleaning up this now.

SofiaMargariti:
> In `development` branch, you'll see that we're passing `editorMode={false}` for those cases which is not used anymore in v3, so those components receive the same props as every other registered component.
> 
> IMO the only thing that needs to be tested is the rendering of image components with no asset bound. Previously, we dealt with it in two different places, which makes no sense - so that's why I'm consolidating/cleaning up this now. Sorry, not sure if that's the answer you were seeking for 😅

Thanks for the explanation, I somehow missed that this was only targeting `v3-prerelease`. So now in the v3 we don't need this special handling for Preview/Delivery, is that right?
Sorry for all the questions, I think I am missing some context, so I will leave it for Max to review this one :)

SofiaMargariti:
> > In `development` branch, you'll see that we're passing `editorMode={false}` for those cases which is not used anymore in v3, so those components receive the same props as every other registered component.
> > IMO the only thing that needs to be tested is the rendering of image components with no asset bound. Previously, we dealt with it in two different places, which makes no sense - so that's why I'm consolidating/cleaning up this now. Sorry, not sure if that's the answer you were seeking for 😅
> 
> Thanks for the explanation, I somehow missed that this was only targeting `v3-prerelease`. So now in the v3 we don't need this special handling for Preview/Delivery, is that right? Sorry for all the questions, I think I am missing some context, so I will leave it for Max to review this one :)

Ok, I found the context I was missing in the Jira ticket, so we're doing the same thing in the [Image component](https://github.com/contentful/experience-builder/blob/development/packages/components/src/components/Image/Image.tsx#L11) itself, that's why we're cleaning it up here.

Chaoste:
Thanks for calling that out. I didn't realise that the PR description didn't cover this aspect at all. Thanks for pointing that out!

> Ok, I found the context I was missing in the Jira ticket, so we're doing the same thing in the [Image component](https://github.com/contentful/experience-builder/blob/development/packages/components/src/components/Image/Image.tsx#L11) itself, that's why we're cleaning it up here.

PR #1182, opened 2025-06-16
fix: bundle CSS and add it directly to the HTML [SPA-2865]

Chaoste:
## Purpose

While testing some changes, I noticed that the static CSS from native components is loaded via JS. For SSR applications, this can cause some flash of layout changes. While it's not very realistic, the most drastic change happens when rendering an image with no bound asset as can be seen on the recording. We have styles for a few components, incl. basic ones like `Container` which is why it is important that those should not be loaded delayed.

https://github.com/user-attachments/assets/59345f69-519e-4da1-9a2e-4dc3e224a69e

## Approach

I adjusted the rollup config of the `components` package to extract a minimized CSS file. This file is imported in the main SDK package close to the function definition of `defineComponents`. This way, the CSS get's loaded both for server and client side, as both will have to call that function.

I also noticed that the main SDK package uses `postcss` while that's only necessary for the styles of the `ErrorBoundary` component. I turned those into inline styles to make those static as well and reduce the size of this package (though it's a dev dep, ofc).

Add-on: I noticed that `Flex.tsx` is not used and removed it as well.


SofiaMargariti:
The change makes sense but I am not sure if importing it in the componentRegistry will cause any issues. Testing in react-vite test app I see this warning, (although in next js I dont see something similar):
Screenshot 2025-06-18 at 14 52 36
Would now the developer need to manually import this in their app at the right place?

SofiaMargariti (approved):
Looks good, thanks for looking into the import issue!

PR #1191, opened 2025-06-23
fix: when PATTERN OPENED FOR EDITING, and it nests a pattern in Preview+Delivery mode default value for prebinding works [SPA-2583]

dimitrycf:


dimitrycf:
> Hey @dimitrycf could you add steps to test/reproduce this issue?
> 
> > when PATTERN OPENED FOR EDITING, and it nests a pattern in Preview+Delivery
> 
> Also I am not sure I understand the above line. How can it be "opened for editing" **and** "nesting in Preview/Delivery" at the same time.

Thanks for the comment @SofiaMargariti! I have fixed the title.  
What I should have written was somethign like:
 - open pattern `pA` for editing
 - this `pA` should have a nested pattern `pB` 
 - the `pB` should have prebinding setup and have default value
 - expected behaviour: when previewing in `pA`  in P+D mode  you can see that `pB` is nested there and it HAS default prebinding value displayed
 - actual behaviour: when previewing `pA` in P+D mode, you can se that `pB` is nested, but it HAS NO default prebinding values displayed

PR #1201, opened 2025-06-30
refactor: add prebindingDefinitions in pattern schema [SPA-2900]

SofiaMargariti:
## Purpose

Introduce wrapping object for prebinding definitions.

⚠️ This is breaking change for pattern with prebinding enabled, but it's ok since the feature is still in development.

## Approach

* Updated schema to reflect the new format and update all related code references and test fixtures to use the new format.

Before:
```
"componentSettings": {
  "parameterDefinitions": {
    ...
  },
  "variableMappings": {
     ...
  }
}
```

After:
```
"componentSettings": {
  "prebindingDefinitions": [
    {
      "id": "$randomId",
      "parameterDefinitions": {
        ...
      },
      "variableMappings": {
        ...
      }
    }
  ]
}
```

* Added also support for `passToNodes` and `allowedVariableOverrides` in Pattern Schema.
* Made `componentSettings`, `prebindingDefinitions` and `passToNode` objects strict.

### DIFF
--- a/packages/validators/src/schemas/v2023_09_28/pattern.ts
+++ b/packages/validators/src/schemas/v2023_09_28/pattern.ts
@@ -80,13 +80,20 @@ export const ComponentVariablesSchema = z.record(
   ComponentVariableSchema,
 );
 
-const ComponentSettingsSchema = z.object({
-  variableDefinitions: ComponentVariablesSchema,
-  thumbnailId: z.enum(THUMBNAIL_IDS).optional(),
-  category: z.string().max(50, 'Category must contain at most 50 characters').optional(),
+export const PrebindingDefinitionSchema = z.object({
+  id: propertyKeySchema,
   variableMappings: VariableMappingsSchema.optional(),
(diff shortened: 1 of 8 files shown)


packages/validators/src/schemas/v2023_09_28/pattern.ts
```diff
@@ -80,13 +80,31 @@ export const ComponentVariablesSchema = z.record(
...
+    passToNodes: z.array(PassToNodeSchema).optional(),
+  })
+  .strict();
+
+const ComponentSettingsSchema = z
+  .object({
+    variableDefinitions: ComponentVariablesSchema,
+    thumbnailId: z.enum(THUMBNAIL_IDS).optional(),
+    category: z.string().max(50, 'Category must contain at most 50 characters').optional(),
+    prebindingDefinitions: z.array(PrebindingDefinitionSchema).optional(),
```
Spring3, line 105: can we also have a validation rule for length === 1?

PR #1202, opened 2025-06-30
fix: makes in Preview+Delivery mode default value for nested pattern's  prebinding work (scope applies to when pattern nests pattern, not experience  embeds pattern) [SPA-2853]

dimitrycf:
## Purpose

So basically as part of working on SPA-2853, I did implement the feature and did a bit of refactoring:

instead of attachPrebindingDefaultValueAsDataSource() the best simplification i could come up with is to introduce a concept of "sideloading". eg. if apart from normal "experience loading" we're going to "inject" some more stuff into that - let's call it "sideloading"
in this example - we're sideloading default prebinding entries
to avoid confusion with the keys in the dataSource  map,  we can prefix all of the "sideloaded" keys with sideloaded_ prefix. This way it's clear that it wasn't originally on the original experienceEntry.dataSource field and was added later during sideloading.
also the wonderful method you added prebindingUtils.ts#resolveMaybePrebindingDefaultValuePath()  was calling resolvePrebindingPath() , but that isn't strictly necessary, as it's possible to make it self contained and we can avoid this awkward  fake patternProperties object which causes extra indirection and extra cognitive load. Now your method is self contained.
also as I refactored "prebinding default value" sideloading mechanism, now it sideloads ALL of the default value entries, not just first one it finds...  (here it was finding first prebinding with defaultValue.. I am not sure when there will be multiple ones, but if data structure allows it, we can just load all the default values we can find). 

### How to reproduce/manually test

Prerequisites:
 - create pB (this my [pB pattern](https://app.contentful.com/spaces/tofsejyzyo24/experiences/5BdJFKCqX5tK6DbJg7PuqE?focusedField=title))
  - with pB setup prebinding
  - with pB setup prebinding default value
  - checkpoint: when previewing pB (in P+D mode) you will see that your default value works
 - create pA
   - you can drop a Button or smth for reference  into it
   - add nested pB
   - save pA
   - in EDITOR mode, you will NOT see default values of pB (fixing this used to be part of the scope of the SPA-2583, but as per DanV the data structures were not available to make it work, so scope was reduced to fixing it in P+D mode s only).
   - bug: in P+D mode, when you preview pA you will NOT see default values rendered for the nested pB 
   - fix-of-the-bug: in P+D mode when you preview pA you WILL see default values rendered for the nested pB.

##### Bug original  symptoms
 - open pattern `pA` for editing
 - this `pA` should have a nested pattern `pB` 
 - the `pB` should have prebinding setup and have default value
 - expected behaviour: when previewing in `pA`  in P+D mode  you can see that `pB` is nested there and it HAS default prebinding value displayed
 - actual behaviour: when previewing `pA` in P+D mode, you can se that `pB` is nested, but it HAS NO default prebinding values displayed


dimitrycf:
> Hey @dimitrycf could you add steps to test/reproduce this issue?
> 
> > when PATTERN OPENED FOR EDITING, and it nests a pattern in Preview+Delivery
> 
> Also I am not sure I understand the above line. How can it be "opened for editing" **and** "nesting in Preview/Delivery" at the same time.

Thanks for the comment @SofiaMargariti! I have fixed the title.  
What I should have written was somethign like:
 - open pattern `pA` for editing
 - this `pA` should have a nested pattern `pB` 
 - the `pB` should have prebinding setup and have default value
 - expected behaviour: when previewing in `pA`  in P+D mode  you can see that `pB` is nested there and it HAS default prebinding value displayed
 - actual behaviour: when previewing `pA` in P+D mode, you can se that `pB` is nested, but it HAS NO default prebinding values displayed

Spring3:
Btw, feel free to park this ticket for now, because it looks like it hits on the same problem - lack of the nested pattern's part of the data structure. Sorry, I probably should have pointed it out earlier

Spring3:
From my tests, the latest version of the SDK already renders the simple pattern with pre-binding correctly on preview, as well as editor. Experience is rendered correctly as well. 

Nested patterns get their default content rendered when previewing a pattern (expected)
Parent patterns do not get rendered at all when previewing an experience (not expected)

dimitrycf:
> Btw, feel free to park this ticket for now, because it looks like it hits on the same problem - lack of the nested pattern's part of the data structure. Sorry, I probably should have pointed it out earlier

As i mentioned in our DMs when it comes to rendering in Preview+Delivery modes, all data structures are there.  The video should clarify the specific use cases which are part of the scope here.

Spring3 (approved):
Ok, I probably have built the bundle without clearing the cache or something. I have double-checked that it works with patterns.

Sorry for the confusion

PR #1205, opened 2025-07-01
fix: ship CSS-in-JS instead of importing CSS for Nextjs support [SPA-2865]

Chaoste:
## Purpose

Importing the CSS separately to improve SSR caused an error as nextjs enforced the user to import all CSS manually instead of having modules import on their own.

Previous PR: https://github.com/contentful/experience-builder/pull/1182

## Approach
I tried to support a hybrid approach, i.e. per default we still load via CSS-in-JS but also offer an option to turn that off and import the CSS on your own for SSR. Generating a CSS and creating in inside JS was not easy to achieve due to https://github.com/egoist/rollup-plugin-postcss/issues/401 . After making it work with two separate rollup configs, I realized that it's not feasible to generate a function `injectGlobalStyles` that can be invoked conditionally. `inject` by postcss only allows the global injection it seems.

To fix the SDK for now, we're rolling back the change and use CSS-in-JS everywhere. We will introduce a breaking change with one of the upcoming major releases to enforce manual CSS imports.

### DIFF
--- a/examples/astrojs-ssr/README.md
+++ b/examples/astrojs-ssr/README.md
@@ -35,6 +35,6 @@ Now that you have set up your environment variables, you can start the developme
 npm run dev
 ```
 
-The app is set up to run on `http://localhost:4321`. By default, the root URL will pull up an experience with the slug of 'home-page'. The locale will be determined by your browser settings.
+The app is set up to run on `http://localhost:3000`. By default, the root URL will pull up an experience with the slug of 'home-page'. The locale will be determined by your browser settings.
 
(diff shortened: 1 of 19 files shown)


examples/astrojs-ssr/README.md
```diff
@@ -35,6 +35,6 @@ Now that you have set up your environment variables, you can start the developme
 npm run dev
 ```
 
-The app is set up to run on `http://localhost:4321`. By default, the root URL will pull up an experience with the slug of 'home-page'. The locale will be determined by your browser settings.
+The app is set up to run on `http://localhost:3000`. By default, the root URL will pull up an experience with the slug of 'home-page'. The locale will be determined by your browser settings.
```
Chaoste, line 38: Streamlining ports in example apps

SofiaMargariti (approved):
Looks good! Thank you for updating the SDK version on the example apps as well!

PR #1209, opened 2025-07-02
feat: v2-prerelease with development merged in

dimitrycf:
## Purpose
The branch `tmp-v2-prerelease-try-merge-in-development` is synced with `development`  (up to versions of 1.42.1 or smth).


examples/binding-to-reference-fields/Sdk1vs2.tsx
Chaoste, line 1: The `examples` folder contains runnable examples applications for various frameworks. As this is not an app on its own (it shouldn't ofc), can we separate this somehow from the others? I suggest moving this from `examples/binding-to-reference-field` to `examples/snippets/binding-to-reference-field`

dimitrycf: Let's do it in some followup.

PR #1240, opened 2025-07-31
fix(visual-editor): render cfVisibility in CSS for CSR apps [SPA-2956]

Chaoste:
## Purpose

As part of #1158, we had to change the way we render `cfVisibility` to not override custom display values for visible nodes.
Due to a faulty merge conflict resolution, the visibility was not rendered at all in editor mode for v3.

## Approach

Turn this prop also in CSS - we have to do it explicitly but at least, it's only one function call.


maxcheremisin (approved):
🙏

PR #1271, opened 2025-08-19
fix: [ZEND-6782] calculate 100% page size

maxcheremisin:
## Purpose

If the page have `html, body { height: 100% }` set, the height is calculated based off the default web app container height, which is 200px. 

Fallback to the `body`'s children deepest node to calculate the content height.


Chaoste (approved):
👍🏻

PR #1275, opened 2025-08-21
fix: render slot children only once in preview [SPA-3147]

Chaoste:
## Purpose

The logic for preview rendering didn't correctly handle the case of using `children: true` together with slots. It would render every slot node as well as direct child node.

## Approach

Since the editor was already treating this correctly, I decided to streamline this logic, so it can't run out of sync anymore.
Further, I added a JSDoc warning devs from using both at the same time as everything could be achieved with using either one of these options.

## Tested scenario
Screenshot 2025-08-21 at 13 40 55

| Before | After |
|--------|--------|
| Screenshot 2025-08-21 at 13 37 05 |  Screenshot 2025-08-21 at 13 39 21 |


maxcheremisin (approved):
LGTM

PR #1285, opened 2025-08-26
fix: improve resolver types and streamline logging shape [SPA-3178]

Chaoste:
## Purpose

Users are not aware that they have to handle `undefined` for link/ array variables.

## Approach

Adjusted the type to cover this explicitly and add a JSDoc that explains what to do.

Also, I replaced all `console` calls with our `debug` utility. This will remove the noise for customers - we can then show specific logs explicitly while hiding the others.

### DIFF
--- a/packages/experience-builder-sdk/src/ExperienceRoot.tsx
+++ b/packages/experience-builder-sdk/src/ExperienceRoot.tsx
@@ -34,7 +34,7 @@ export const ExperienceRoot = ({
   const mode = useDetectCanvasMode();
 
   if (typeof debug === 'boolean') {
-    cfDebug.setEnabled(debug);
+    cfDebug.setActiveLevel(debug ? 'debug' : 'warn');
   }
 
   //If experience is passed in as a JSON string, recreate it to an experience object


packages/core/src/utils/debugLogger.spec.ts
```diff
@@ -1,36 +1,36 @@
 import { debug, DebugLogger } from './debugLogger';
 import { describe, it, expect, vi } from 'vitest';
 
-describe('debugLogger', () => {
+describe.only('debugLogger', () => {
```
SofiaMargariti, line 4:
```suggestion
describe('debugLogger', () => {
```

packages/core/src/utils/debugLogger.spec.ts
```diff
@@ -1,36 +1,36 @@
...
   it('should not log anything if debug is not enabled', () => {
     const consoleLogSpy = vi.spyOn(console, 'log');
     debug.log('test');
     expect(consoleLogSpy).not.toHaveBeenCalled();
   });
 
-  it('should log if debug is enabled', () => {
+  it.only('should log if debug is enabled', () => {
```
SofiaMargariti, line 11:
```suggestion
  it('should log if debug is enabled', () => {
```
we should probably have an eslint rule for this :)

SofiaMargariti (approved):
Looks good, just make sure to remove the `.only`.

Worth calling this change out in slack once we merge it, as now people will need to enable the logging to see the debug logs.

packages/core/src/utils/debugLogger.spec.ts
Chaoste: Oops xD

Chaoste: Made myself a memo for that

PR #1291, opened 2025-09-01
fix: introduce new registry options to remove special casing for stru…

Chaoste:
…ctures

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -142,8 +142,17 @@ export type ComponentRegistration = {
     /**
      * If true, the component receives the optional boolean property `isInExpEditorMode` to
      * render different content between editor and delivery mode.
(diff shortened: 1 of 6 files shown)


packages/core/src/types.ts
```diff
@@ -142,8 +142,17 @@ export type ComponentRegistration = {
...
      * render different content between editor and delivery mode.
+     *
+     * @deprecated this will be replaced by editorProperties in the next major version
      */
     enableCustomEditorView?: boolean;
+    /**
+     * If set, the specified properties are passed to the component when rendered in the editor.
+     */
+    editorProperties?: {
+      node?: boolean;
```
SofiaMargariti, line 153: Should we have a doc string per editor property to explain what each one does?

PR #1294, opened 2025-09-01
style: remove unused function useSingleColumn []

Chaoste:
## Purpose

This function `useSingleColumn` is a relict from pre-v3 times. It's not used anymore and can be removed ocne and for all.


maxcheremisin (approved):
🧹

PR #1305, opened 2025-09-03
fix: allow passing children property if there are no children [SPA-3210]

Chaoste:
## Purpose

If the user defined a custom property `children`, it should be passed through to their custom component.

Only if the component accepts true child nodes (`children: true`) and it's a non-empty list, they will take precedence over that custom property.

## Approach
After having a look at the type of `React.createElement`, I learned that `children` should be passed as [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) (`...children`). I worked probably this way because React can now render multiple nodes on the top-level of one render call ([added in React 16.0](https://legacy.reactjs.org/blog/2017/09/26/react-v16.0.html)).

This is a regression of the change I did [here](https://github.com/contentful/experience-builder/pull/1275/files#diff-e801c2e3de23af7313c1e3cd5d16aed2dd3a7e5389409fac9b59db20bb522136).

### DIFF
--- a/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
+++ b/packages/experience-builder-sdk/src/blocks/preview/CompositionBlock.tsx
@@ -308,7 +308,8 @@ export const CompositionBlock = ({
       ...sanitizeNodeProps(props),
       ...renderedSlotNodesMap,
     },
-    renderedChildren,
+    // If there are no children, a custom property called `children` can be passed through to the custom component
+    ...(renderedChildren ?? []),
   );
 };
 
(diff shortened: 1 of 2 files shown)


maxcheremisin (approved):

SofiaMargariti (approved):
[Non-blocking] shouldn't we have some unit tests for behaviours like these?

Chaoste:
> [Non-blocking] shouldn't we have some unit tests for behaviours like these?

Will follow-up on this likely tomorrow, including the validation.

PR #1311, opened 2025-09-04
fix: render nodes and canvas as empty if all nodes are hidden [SPA-3216]

Chaoste:
## Purpose
When all nodes are hidden (via `cfVisibility`), the canvas would render empty in the editor experience. This looks confusing.

## Approach
Check for the visibility of each node to determine whether to render the "empty canvas" message or not.

https://github.com/user-attachments/assets/3e74fb66-a245-41c6-95ef-022ffcff05e9


kizer-cf (approved):
Tested and confirmed to work

PR #1326, opened 2025-09-16
feat: render prebinding data in preview mode [SPA-2923]

Spring3:


packages/experience-builder-sdk/src/core/preview/PrebindingManager.ts
```diff
@@ -0,0 +1,547 @@
+import {
```
SofiaMargariti, line 1: Would it make sense to keep a minimal version of the prebindingManager that only includes:
* `storePrebindingDefinitions`
* `storeParameterDefinitions`
* `getHoistedParameterByNodeId`

Given that the one in the Web App has already diverged from this one, it just makes things more complicated to keep them in sync.

Spring3: yes, it makes sense 👍

Spring3: I will align it with the latest version from the web app I think

packages/experience-builder-sdk/src/core/preview/assemblyUtils.spec.ts
```diff
@@ -84,44 +80,104 @@ describe('resolvePattern', () => {
...
-      );
+      expect(result.children[0].children[0]).toEqual({
+        ...assemblyEntry.fields.componentTree.children[0].children[0],
+        displayName: undefined,
+        parameters: undefined,
+        pattern: {
+          nodeIdOnPattern: 'test-custom-component-id',
+          parentPatternNodeId: undefined,
+          prefixedNodeId: 'test-custom-component-id',
+        },
```
Spring3, line 99: wanted to point out that I decided to keep these properties just in case, but they aren't used anywhere. So we can clean them up. I'm happy to hear your opinions

packages/experience-builder-sdk/src/core/preview/PrebindingManager.ts
```diff
@@ -0,0 +1,323 @@
...
+    return Object.keys(PrebindingManager.getAllParameterDefinitionsForNodeId(nodeId));
+  }
+
+  /**
+   * Get the highest level overriding parameterId for a given parameterId + nodeId combination
+   * @param parameterId - The parameter to resolve
+   * @param indexedNodeId - The indexed node ID to resolve the parameterId for
+   * @returns string - The resolved parameterId - will be from "root" in patterns or the parent pattern node in experiences
+   */
+  public static getHoistedIdForParameterId(parameterId: string, indexedNodeId: string): string {
```
Ollie-H, line 256: Do we not need the latest version with the fix applied from this morning?

Spring3: Since we don't have the editor functionality at this point, we don't care of the component node ids in the rendering path

Ollie-H: Makes sense. Hopefully we don't need to touch it much but the fact they are mostly identical could get confusing why/when they should diverge.

packages/experience-builder-sdk/src/utils/prebindingUtils.ts
SofiaMargariti, line 139: It's not in your changes, but this will fail if there is no mapping for a specific CT. It's the scenario where I am mapping:
* `CT-A` to:
    * propA
    * propB
* `CT-B` to:
    * prop C
 And my selected source is of type `CT-A`. When resolving mapping for propC, `variableMapping.pathsByContentType[contentTypeId]` will be undefined and this will throw. Adding some optional chaining should cover this case.

SofiaMargariti (approved):
Sorry it took me a while to test this. The code looks good overall, just an edge-case that I noticed and I've comment above!

PR #1328, opened 2025-09-16
feat: add validations for pre-binding attributes in component tree nodes [SPA-3201]

SofiaMargariti:
## Purpose

* Add validation to make sure that `parameters` and `prebindingId` appear always as a pair in component tree nodes.
* Parameters can be an empty object in case a pattern instance has prebinding configured but not source has been selected yet.


Chaoste (approved):

Spring3 (approved):
👍

PR #1340, opened 2025-09-22
chore: consolidate rollup configs and exported builds [SPA-3248]

Chaoste:
## Purpose

Follow up from #1332

When using the latest SDK version `3.7.0`+, you might see an issue with doing `jest.spyOn` on an exported function of the core package. The current workaround for that would be to add `__esModule: true` manually through `jest.mock`.

However, it made me wonder why that is and pushed me too understand more deeply the mechanics of the rollup config and the build format. While revisiting the config, I added some improvements which I'll explain via PR comments.

Resulting build of the core package, notice the reduced number of files and structure of esm/cjs/types:
image

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -51,9 +51,6 @@ export default [
       alias({
         entries: [{ find: '@', replacement: path.resolve(projectRoot, 'src') }],
       }),
-      nodeResolve({
-        extensions: ['.mjs', '.js', '.json', '.ts', '.tsx'],
-      }),
       dts({ compilerOptions: { noEmitOnError: process.env.DEV ? false : true } }),
     ],
     external: [/.css/],


packages/experience-builder-sdk/rollup.config.mjs
```diff
@@ -37,5 +37,6 @@ export default [
     input: 'src/index.ts',
     output: [{ file: 'dist/index.d.ts', format: 'es' }],
     plugins: [dts({ compilerOptions: { noEmitOnError: process.env.DEV ? false : true } })],
+    external: [/@contentful\/experiences-core/],
```
Chaoste, line 40: Importing from `core/types` in the SDK raises the warning that they are not added to the final build. In the `components` package, I decided to refactor all cases but this would is the quick solution for now to solve it through the rollup config.

duruld (approved):

PR #1342, opened 2025-09-22
feat: make passToNode mandatory in parameter definition validator [SPA-3223]

kizer-cf:
Ticket: https://contentful.atlassian.net/browse/SPA-3223

Also added code to define design tokens for the test app, so we use more of what we support

### DIFF
--- a/packages/validators/src/schemas/v2023_09_28/pattern.ts
+++ b/packages/validators/src/schemas/v2023_09_28/pattern.ts
@@ -81,8 +81,8 @@ const ParameterDefinitionSchema = z.object({
   contentTypes: z.array(z.string()).min(1),
   passToNodes: z
     .array(PassToNodeSchema)
-    .max(1, 'At most one "passToNodes" element is allowed per parameter definition.')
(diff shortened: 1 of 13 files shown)


packages/validators/src/schemas/v2023_09_28/pattern.ts
```diff
@@ -81,8 +81,8 @@ const ParameterDefinitionSchema = z.object({
   contentTypes: z.array(z.string()).min(1),
   passToNodes: z
     .array(PassToNodeSchema)
-    .max(1, 'At most one "passToNodes" element is allowed per parameter definition.')
-    .optional(), // we might change this to be empty array for native parameter definitions, that's why we don't use .length(1)
+    .max(1, 'At most one "passToNodes" element is allowed per parameter definition.'),
+  // we might change this to be empty array for native parameter definitions, that's why we don't use .length(1)
```
SofiaMargariti, line 85: nit: this comment can now be removed - also it's wrong 😄, we shouldn't use `length(1)` in any case.
```suggestion
```

PR #1357, opened 2025-09-30
Release@main

kizer-cf:
- https://github.com/contentful/experience-builder/pull/1350
- https://github.com/contentful/experience-builder/pull/1326
- https://github.com/contentful/experience-builder/pull/1353
- https://github.com/contentful/experience-builder/pull/1354
- https://github.com/contentful/experience-builder/pull/1346
- https://github.com/contentful/experience-builder/pull/1347
- https://github.com/contentful/experience-builder/pull/1342
- https://github.com/contentful/experience-builder/pull/1348
- https://github.com/contentful/experience-builder/pull/1358

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "npmClient": "npm",
-  "version": "3.8.0-beta.2",
+  "version": "3.8.0-beta.3",
   "command": {
     "version": {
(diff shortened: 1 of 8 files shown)


Chaoste:
@kizer-cf 
Is there anything in this set of changes that would be of value right now for customers? Or is it already the final release for prebinding? If not, I would suggest waiting for a new official release until we have something, so it generates a value for them to upgrade. If it's just for testing prebinding, a beta release would be sufficient, wouldn't it?

PR #1365, opened 2025-10-07
fix: detect layout shift on media loading [ZEND-6964]

maxcheremisin:
## Purpose

Use `ResizeObserver` to detect video or image resizing, e.g. after lazy loading, or changing `src` attribute, instead of relying on load event.

Avoid an extra rerender that could cause infinite updates if a custom component mutates the dom.


Chaoste (approved):
Nice work, thank you!

PR #1425, opened 2026-03-27
fix(core): add md5 dep [FE-545]

denkristoffer:
## Purpose

The core package is importing `md5` but the dependency is missing.

Related to ticket https://contentful.atlassian.net/browse/FE-545 because it's causing https://github.com/contentful/user_interface/pull/30820 to fail.


Chaoste (approved):
🙏 thanks for catching that one

